PR #165, opened 2021-12-01
Add lots of error handling for allocation failures, and add an allocation failure injection framework to the tests

rtjohnso:
This patch starts the work of plumbing through error handling for disk-space allocation failures.

It replaces many asserts with code to clean up and return an error.  The error handling code has not been extensively tested, but it's a start.

This patch also adds a `MUST_CHECK_RETURN` annotation for functions, indicating that it is an error to ignore their return value.  Unfortunately, gcc and clang don't require that you actually do anything useful with the return value other than save it to a local variable, but it does help find call sites that are discarding a return value.

The patch also adds a new `allocator` that wraps the existing `rc_allocator` and throws random allocation failures.  The goal of this allocator is to test the error handling in splinter.

This patch also modifies the `functionality` test to retry insertions when there is an error.  Thus, in theory, `functionality` should run and get correct results even when the fault injection allocation scheme is being used.  It does not yet pass that test, so I've not yet added it to `test.sh`.

One major outstanding item is `splinter_split_leaf`, but I am sure there are many other places that also slipped through.

### DIFF
--- a/include/splinterdb/kvstore.h
+++ b/include/splinterdb/kvstore.h
@@ -56,7 +56,7 @@ kvstore_open(const kvstore_config *cfg, kvstore **kvs);
 // Close a kvstore
 //
 // This will flush everything to disk and release all resources
-void
+int
 kvstore_close(kvstore *kvs);
 
 // Register the current thread so that it can be used with the kvstore.
--- a/src/allocator.h
+++ b/src/allocator.h
@@ -40,7 +40,7 @@ typedef struct allocator allocator;
 
 typedef platform_status (*alloc_fn)(allocator *al,
                                     uint64 *   addr,
-                                    page_type  type);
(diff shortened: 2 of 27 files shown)


gapisback:
Update:

- Pieces of this work like a fault-injecting allocator is useful, and we should try to restore that.
- MUST_CHECK_RETURN attribute ... can try to salvage that. (Can apply this to platform_status return.)
- May need to re-do this whole machinery as /main has moved on quite a bit.
- Find ways to break this work into smaller items that can be checked-in separately.
- **Revisit this work after 4/15th deadline** ... and find ways how to carve this out into smaller pieces.

ajhconway:
Sadly this is never getting merged as is, and while we want to return to this effort, it will probably require redoing this work.

PR #190, opened 2021-12-20
Avoid calling memmove with an invalid source pointer

danielgustafsson:
While it's legal to pass zero as the `n` parameter to `memmove()`, it is not legal to pass a `NULL` pointer.  Per the C11 standard 7.24.1/2:

> "Where an argument declared as size_t n specifies the length of the array for a function, n can have the value zero on a call to that function.  Unless explicitly stated otherwise in the description of a particular function in this subclause, pointer arguments on such a call shall still have valid values, as described in 7.1.4."

Skip the call to `memmove()` when value isn't set, saving a useless syscall in the process.  This applies to encoding `MESSAGE_TYPE_DELETE` messages.


ajhconway:
We should consider doing an audit of the rest of the code base for similar things that might happen, but `encode_value` has been worked out, so I'm closing out this PR.

PR #302, opened 2022-03-02
Clean up platform_log and friends

ajhconway:
Modifies `platform_log()` to take a `platform_log_handle *` parameter. The
old functionality is now handled by `platform_default_log()`.

Changes `Platform_stdout_fh` to `Platform_default_log_handle`.
Changes `Platform_stderr_fh` to `Platform_error_log_handle`.
Removes `PLATFORM_DEFAULT_LOG_HANDLE` and `PLATFORM_ERR_LOG_HANDLE`.

Replaces many instances of `platform_log_stream` with `platform_log`.

Adds a `platform_log_handle *` parameter to `*_print_*` functions.

Adds a `platform_stream_handle` struct that holds a string, length and log handle.
Explicitly passes a stream parameter to stream printing functions.

### DIFF
--- a/include/splinterdb/platform_linux/public_platform.h
+++ b/include/splinterdb/platform_linux/public_platform.h
@@ -8,8 +8,8 @@
  *     for internal use.
  */
 
-#ifndef __PLATFORM_PUBLIC_H
-#define __PLATFORM_PUBLIC_H
+#ifndef __PUBLIC_PLATFORM_H
+#define __PUBLIC_PLATFORM_H
 
 #include <stdio.h>
 
@@ -68,7 +68,8 @@ typedef int32 bool;
 typedef uint8 bool8;
 
 // See platform.c
-extern FILE *Platform_stdout_fh; // File handle for stdout
-extern FILE *Platform_stderr_fh; // File handle for stderr
+typedef FILE                platform_log_handle;
+extern platform_log_handle *Platform_default_log_handle; // stdout FILE handle
+extern platform_log_handle *Platform_error_log_handle;   // stderr FILE handle
 
-#endif // __PLATFORM_PUBLIC_H
+#endif // __PUBLIC_PLATFORM_H
--- a/src/btree.c
+++ b/src/btree.c
@@ -2170,7 +2170,7 @@ btree_async_callback(cache_async_ctxt *cache_ctxt)
 
    platform_assert(SUCCESS(cache_ctxt->status));
    platform_assert(cache_ctxt->page);
-   //   platform_log("%s:%d tid %2lu: ctxt %p is callback with page %p
+   //   platform_default_log("%s:%d tid %2lu: ctxt %p is callback with page %p
    //   (%#lx)\n",
    //                __FILE__, __LINE__, platform_get_tid(), ctxt,
    //                cache_ctxt->page, ctxt->child_addr);
@@ -2235,7 +2235,7 @@ btree_lookup_async_with_ref(cache            *cc,        // IN
             switch (res) {
                case async_locked:
                case async_no_reqs:
-                  //            platform_log("%s:%d tid %2lu: ctxt %p is
+                  //            platform_default_log("%s:%d tid %2lu: ctxt %p is
                   //            retry\n",
                   //                         __FILE__, __LINE__,
                   //                         platform_get_tid(), ctxt);
@@ -2246,7 +2246,7 @@ btree_lookup_async_with_ref(cache            *cc,        // IN
                   done = TRUE;
                   break;
                case async_io_started:
-                  //            platform_log("%s:%d tid %2lu: ctxt %p is
+                  //            platform_default_log("%s:%d tid %2lu: ctxt %p is
                   //            io_started\n",
                   //                         __FILE__, __LINE__,
                   //                         platform_get_tid(), ctxt);
@@ -2603,19 +2603,19 @@ btree_iterator_print(iterator *itor)
    debug_assert(itor != NULL);
    btree_iterator *btree_itor = (btree_iterator *)itor;
 
-   platform_log("########################################\n");
-   platform_log("## btree_itor: %p\n", itor);
-   platform_log("## root: %lu\n", btree_itor->root_addr);
-   platform_log(
+   platform_default_log("########################################\n");
+   platform_default_log("## btree_itor: %p\n", itor);
+   platform_default_log("## root: %lu\n", btree_itor->root_addr);
+   platform_default_log(
       "## curr %lu end %lu\n", btree_itor->curr.addr, btree_itor->end_addr);
-   platform_log("## idx %lu end_idx %lu generation %lu\n",
-                btree_itor->idx,
-                btree_itor->end_idx,
-                btree_itor->end_generation);
-   btree_print_node(btree_itor->cc,
+   platform_default_log("## idx %lu end_idx %lu generation %lu\n",
+                        btree_itor->idx,
+                        btree_itor->end_idx,
+                        btree_itor->end_generation);
+   btree_print_node(Platform_default_log_handle,
+                    btree_itor->cc,
                     btree_itor->cfg,
-                    &btree_itor->curr,
-                    PLATFORM_DEFAULT_LOG_HANDLE);
+                    &btree_itor->curr);
 }
 
 const static iterator_ops btree_iterator_ops = {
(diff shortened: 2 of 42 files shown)


include/splinterdb/platform_linux/public_platform.h
```diff
@@ -68,7 +68,9 @@ typedef int32 bool;
 typedef uint8 bool8;
 
 // See platform.c
-extern FILE *Platform_stdout_fh; // File handle for stdout
-extern FILE *Platform_stderr_fh; // File handle for stderr
+typedef FILE platform_log_handle;
+extern platform_log_handle
+   *Platform_default_log_handle;                       // File handle for stdout
```
gapisback, line 73: Minor request: For readability, if you can comment these two lines as below, it will be better. And clang-format will not complain / re-indent then:

```
 71 typedef FILE                platform_log_handle;
 72 extern platform_log_handle *Platform_default_log_handle; // stdout FILE handle
 73 extern platform_log_handle *Platform_error_log_handle;   // stderr FILE handle

File: src/btree.c [hunk 2]
```diff
@@ -3145,100 +3145,104 @@ btree_count_in_range_by_iterator(cache        *cc,
...
  */
 void
-btree_print_locked_node(btree_config          *cfg,
-                        uint64                 addr,
-                        btree_hdr             *hdr,
-                        platform_stream_handle stream)
+btree_print_locked_node(btree_config        *cfg,
+                        uint64               addr,
+                        btree_hdr           *hdr,
+                        platform_log_handle *log_handle)
```
gapisback, line 3151: Minor request / suggestion: As long as you are overhauling these interfaces, can you consider making the output stream / log-handle arg; i.e. `platform_log_handle *log_handle` the 1st arg to the print method?

That will keep the interface somewhat consistent with the underlying `platform_log*()` APIs, too, which take the output stream as the 1st arg.

It will be good to consistently move to that convention as we develop / retrofit print methods.

src/btree.c
```diff
@@ -3145,100 +3145,104 @@ btree_count_in_range_by_iterator(cache        *cc,
...
 
 void
-btree_print_node(cache                 *cc,
-                 btree_config          *cfg,
-                 btree_node            *node,
-                 platform_stream_handle stream)
+btree_print_node(cache               *cc,
+                 btree_config        *cfg,
+                 btree_node          *node,
+                 platform_log_handle *log_handle)
```
gapisback, line 3223: Here is another example where the `platform_log_handle *log_handle` could be moved to become the 1st arg.

Btw, I don't have a very strong reason to justify this, except that when I see diffs like this (somewhere else in this file):

```
-   btree_print_node(cc, cfg, &node, PLATFORM_DEFAULT_LOG_HANDLE);
+   btree_print_node(cc, cfg, &node, Platform_default_log_handle);
```

I wondered ... if we could take the opportunity to make the log-handle / stream as the 1st arg, while we are anyways over-hauling such lines of code. 

(If we don't do it now, we won't come back to change it one more time in the future.)

src/btree.c
```diff
@@ -3248,29 +3252,33 @@ btree_print_subtree(cache                 *cc,
...
+                       platform_log_handle *log_handle)
 {
    btree_node node;
    node.addr = addr;
    btree_node_get(cc, cfg, &node, PAGE_TYPE_BRANCH);
 
-   platform_log("Tree stats: height %u\n", node.hdr->height);
-   cache_print_stats(cc);
+   platform_default_log("Tree stats: height %u\n", node.hdr->height);
+   cache_print_stats(cc, log_handle);
```
gapisback, line 3281: btw, I pondered on my own request to consider changing `log_handle` to be the 1st arg. That would cascade to cache-printing function, also, whose interface before this change was:

```
387 cache_print_stats(cache *cc)
```

And after this change becomes:

```
388 cache_print_stats(cache *cc, platform_log_handle *log_handle)
```

I reasoned that as these print methods take anywhere from 1, 2, 3 parameters, it will be [more] consistent to say that the 1st arg is **always**-always the output log-handle / stream.

I am saying: While we are roto-tilling these interfaces, again -- it seems to me to be the right time to move to this 1st-arg-convention I am suggesting here.

src/btree.c
```diff
@@ -3317,8 +3324,8 @@ btree_verify_node(cache        *cc,
                                   btree_get_tuple_key(cfg, node.hdr, idx + 1))
                 >= 0)
             {
-               platform_log_stream("out of order tuples\n");
-               platform_log_stream("addr: %lu idx %2u\n", node.addr, idx);
+               platform_error_log("out of order tuples\n");
```
gapisback, line 3327: This change is **very nice**! Along with L3310 which deletes `platform_open_log_stream()`, this change makes the code much clear and readable that the output messages are error indicators, going to an output error-log stream.

So, thanks!

src/memtable.h
```diff
@@ -282,15 +282,15 @@ memtable_verify(cache *cc, memtable *mt)
 }
 
 static inline void
-memtable_print(cache *cc, memtable *mt)
+memtable_print(cache *cc, memtable *mt, platform_log_handle *log_handle)
```
gapisback, line 285: To reiterate ... I am seeing more print methods' interfaces now changing to take the `platform_log_handle *log_handle` as a new arg. Here, it's the 3rd arg.

I would have gone the other way (as suggested previously) of making this output `log_handle` consistently the 1st-arg in all print-methods.

src/platform_linux/platform.c
```diff
@@ -212,27 +212,35 @@ platform_histo_destroy(platform_heap_id heap_id, platform_histo_handle histo)
 }
 
 void
-platform_histo_print(platform_histo_handle histo, const char *name)
+platform_histo_print(platform_histo_handle histo,
```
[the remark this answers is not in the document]
gapisback: @ajhconway - I agree that this looks fishy. Even more so, this `typedef struct * platform_histo_handle` is declared as a ptr to a struct in `platform_linux/platform_types.h`.

Why not open an issue to re-visit the location of these definitions and do some cleansing, once this goes in?

src/platform_linux/platform_inline.h
```diff
@@ -227,64 +227,71 @@ platform_status_to_string(const platform_status status)
...
-   do {                                                                        \
-      fprintf(stream, __VA_ARGS__);                                            \
-   } while (0)
-
-#define platform_log(...)                                                      \
+#define platform_log_stream(_stream, ...)                                      \
    do {                                                                        \
-      fprintf(Platform_stdout_fh, __VA_ARGS__);                                \
-      fflush(Platform_stdout_fh);                                              \
+      fprintf(_stream->stream, __VA_ARGS__);                                   \
```
gapisback, line 278: Suggestion: What you have also works, but ... now that you have tried to make all outputs go to log-handles, and given that you have also provided this `platform_log_stream_to_log_handle()` interfaces, seems like we can rework this line to do:

```
   platform_log(platform_log_stream_to_log_handle(_stream), __VA_ARGS__);
```

Basically, I am wondering why not have the `log_stream()` also invoke `platform_log_handle()`, so everything is funneled thru the latter?

I think from layering angle that seems right, but wonder if it will make code readability too too hard?

I didn't think it would make the calling sequence too hard to de-mangle, so am suggesting you consider bringing this in.

src/trunk.h
```diff
@@ -458,19 +464,21 @@ uint64
...
+                  uint64               fanout,
+                  uint64               max_branches_per_node,
+                  uint64               btree_rough_count_height,
+                  uint64               filter_remainder_size,
+                  uint64               filter_index_size,
+                  uint64               reclaim_threshold,
+                  bool                 use_log,
+                  bool                 use_stats,
+                  bool                 verbose_logging,
+                  platform_log_handle *log_handle);
```
gapisback, line 481: If you do accept the suggestions [made elsewhere] to keep the `log_handle` as 1st arg to all print-methods, but choose to leave this here, as the last arg, ... that would still be fine with me.

tests/config.c
```diff
@@ -230,6 +234,23 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
                cfg[cfg_idx].use_log = TRUE;
             }
          }
+         config_has_option("no-log")
```
gapisback, line 237: [Aside remark: In the spirit of what we discussed the other day to keep change-sets small, maybe ... in future, we should move-off these types of config options enhancements to a new issue. ??? Just seeking your thoughts, too , here .]

tests/config.c
```diff
@@ -230,6 +234,23 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
...
+         {
+            for (cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
+               cfg[cfg_idx].use_log = FALSE;
+            }
+         }
+         config_has_option("verbose-logging")
```
gapisback, line 243: This name is fine. Just fyi, that I have issue #353 open for something very similar. For that, then, I will name the option `--verbose-progress` as you have reserved this name.

tests/config.c
gapisback: Qs, though ... I couldn't figure this out ... `--no-log` seems to be an existing option?

is this code below a bug-fix, then, to set `use_log = FALSE` ?

Just asking for my own clarification.

tests/functional/cache_test.c
```diff
@@ -740,9 +741,8 @@ test_reader_thread(void *arg)
    platform_semaphore_init(&params->batch_sema, 0, params->hid);
    for (i = k = 0; i < num_pages; i += READER_BATCH_SIZE) {
       if (params->logger) {
-         platform_throttled_error_log(DEFAULT_THROTTLE_INTERVAL_SEC,
-                                      PLATFORM_CR "test %3lu%% complete",
-                                      i * 100 / num_pages);
+         platform_error_log(PLATFORM_CR "test %3lu%% complete",
```
gapisback, line 744: Any reason why this was changed from throttled to just `_error_log()`?

I have issue #336 opened to clean-up `throttled_error_log`. Once that happens, this code will function better.

Did you want to simply move away from throttled to normal error_log here?

tests/functional/splinter_test.c
```diff
@@ -223,11 +223,9 @@ test_trunk_lookup_thread(void *arg)
       for (uint8 spl_idx = 0; spl_idx < num_tables; spl_idx++) {
          if (test_is_done(done, spl_idx))
             continue;
-         platform_throttled_error_log(
-            DEFAULT_THROTTLE_INTERVAL_SEC,
-            PLATFORM_CR "lookups %3lu%% complete for table %u",
-            lookup_base[spl_idx] / (total_ops[spl_idx] / 100),
-            spl_idx);
+         platform_error_log(PLATFORM_CR "lookups %3lu%% complete for table %u",
```
gapisback, line 226: Again, why was throttled removed? (Ref issue #336, which suggests ways to fix this throttling mechanism)

gapisback:
Hi, @ajhconway - The change is fine, by itself, in that it does cleanup the interfaces, and makes things more consistent. So, thanks for that!

However, given that we have such a big churn in the code, I do want to take the opportunity to submit couple of re-think suggestions.

One, as remarked in individual files, I do see some advantage to consistently passing-in `platform_log_handle *log_handle` as the 1st arg to all print methods. I see there is an open discussion on this request.

Second is the nomenclature of `platform_log()` v/s `platform_default_log()`. B seeing the 2nd name, I can tell that it's doing some logging to some default output stream. But I have to then "peek into" the implementation to "know" that this output stream is _the_ `Platform_default_log_handle`.

One consequence of this choice is that there are lots of diffs like this one below, e.g. in `btree.c`:
```
-   platform_log("## idx %lu end_idx %lu generation %lu\n",
-                btree_itor->idx,
-                btree_itor->end_idx,
-                btree_itor->end_generation);
+   platform_default_log("## idx %lu end_idx %lu generation %lu\n",
+                        btree_itor->idx,
+                        btree_itor->end_idx,
+                        btree_itor->end_generation);
```

If I slightly changed my mind-set to "understand" that there are only two logging interfaces:

1. `platform_log()` - which always is the default logging interface, and prints always to `Platform_default_log_handle`. (Here, in my opinion, the term `_default_` is redundant in the method's name, and can be distilled out.
2. And, `platform_error_log` - which is the error logging print interface

... then, I would not see the diff above. Things will just patch-through.

The one 'issue' with the current round of diffs is that the line-length is messed-up and our friend clang-format further indents the code -- all of which affects overall code readability.

That leaves us with existing usages of `platform_log()`, which now need to be changed to receive a `log_handle`.
An example of such a diff from the `trunk.` is:

```
-   platform_log("Space used by level:\n");
+   platform_log(log_handle, "Space used by level:\n");
```

As long as this line of code is being changed, one other way would have been re-name `platform_log()` to `platform_log2` or `platform_log_to` or `platform_log_lh`. 

Each of which still conveys the sentiment of logging "to something" (a log-handle) and keeps the namespace still tight & identifiable.

The diff will still be seen, but it won't de-stabilize clang-formatting choices too much. And I think will not affect overall code readability adversely.

I will leave both suggestions for your consideration, and will move the review along.

Thanks for this hard-grunge-work cleanup!

gapisback (approved):
Left few file-specific comments, and two higher-level comments for consideration.

Thanks!

src/clockcache.c
```diff
@@ -3136,25 +3136,25 @@ clockcache_print_stats(clockcache *cc)
...
-   platform_log("Cache Statistics\n");
-   platform_log("-----------------------------------------------------------------------------------------------\n");
-   platform_log("page type       |      trunk |     branch |   memtable |     filter |        log |       misc |\n");
-   platform_log("----------------|------------|------------|------------|------------|------------|------------|\n");
-   platform_log("cache hits      | %10lu | %10lu | %10lu | %10lu | %10lu | %10lu |\n",
+   platform_log(log_handle, "Cache Statistics\n");
+   platform_log(log_handle, "-----------------------------------------------------------------------------------------------\n");
+   platform_log(log_handle, "page type       |      trunk |     branch |   memtable |     filter |        log |       misc |\n");
+   platform_log(log_handle, "----------------|------------|------------|------------|------------|------------|------------|\n");
+   platform_log(log_handle, "cache hits      | %10lu | %10lu | %10lu | %10lu | %10lu | %10lu |\n",
```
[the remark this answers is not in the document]
gapisback: Oh, I see now ... Seems like locally we must have done `clang-format off`. Ignore my remark ..

src/platform_linux/platform_inline.h
ajhconway: This is fine with me.

Really it may make sense to get rid of this function altogether now and just pass the `log_handle` obtained from the stream to `platform_log`. I'll create an issue for that I think.

src/trunk.h
ajhconway: Yeah, it's serving a different role here, so I'll leave it.

tests/config.c
ajhconway: I thought it was missing, but happy enough to remove it. It's the default option anyway.

The way the boolean options work with our config, they don't take a true/false argument, they just exist or don't.

tests/config.c
ajhconway: Ack.

tests/functional/cache_test.c
ajhconway: Since it wasn't implemented, I thought we might as well deprecate it. But that was prior to your issue, so I'm happy to revisit it and possibly restore this code.

I believe this was something only implemented on the kernel side.

tests/functional/splinter_test.c
```diff
@@ -122,10 +122,10 @@ test_trunk_insert_thread(void *arg)
          if (test_is_done(done, spl_idx)) {
             continue;
          }
-         platform_default_log(PLATFORM_CR
-                              "inserting %3lu%% complete for table %u ... ",
-                              insert_base[spl_idx] / (total_ops[spl_idx] / 100),
-                              spl_idx);
+         platform_error_log(PLATFORM_CR
```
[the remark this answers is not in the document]
ajhconway: I'm happy with either.

My thinking for this to go to stderr is that I usually would have these sorts of status messages on stderr, but if you have a preference for stdout, I can revert.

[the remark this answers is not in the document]
ajhconway: I went ahead and reverted this.

tests/functional/cache_test.c
ajhconway: After reading your later comment and thinking about it, I think it makes sense to leave these. I went ahead and reverted them.

ajhconway (commented):
Responding to comments

PR #362, opened 2022-03-14
Robj/message encoding

rtjohnso:
This PR eliminates the message encoding stuff in splinterdb.c.

Instead, message types are encoded in the metadata in the btree.

This eliminates a copy on inserts.

This also makes the interface for performing inserts much simpler:
`splinterdb_update(spl, key, msg);`

And it eliminates some limitations imposed by the splinterdb.c implementation (e.g. `MAX_ENCODED_MESSAGE_SIZE`) that, although not currently important, could become a hassle down the road.

It also introduces a new `message` type for passing around messages along with their types.  This will also be useful down the road for storing other metadata we will need for indirect messages.

Analogously, it introduces a `merge_accumulator` for performing merges.

These types are very analogous to `slice`s and `writable_buffer`s when they are used for messages, so the code changes are largely mechanical.

These types are used internally and in the `data.h` apis, but don't show up in query or insert apis.

This PR also adds an indirect bit for keys and values in the btree.  This bit is currently unused, but putting it in place will enable us to add indirect keys/values to the btree w/o changing the on-disk format.

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -105,9 +105,9 @@ index_entry_size(const slice key)
 }
 
 static inline uint64
-leaf_entry_size(const slice key, const slice message)
+leaf_entry_size(const slice key, const message msg)
 {
-   return sizeof(leaf_entry) + slice_length(key) + slice_length(message);
+   return sizeof(leaf_entry) + slice_length(key) + message_length(msg);
 }
 
 static inline uint64
@@ -176,6 +176,7 @@ btree_fill_index_entry(const btree_config *cfg,
                 <= btree_page_size(cfg));
    memcpy(entry->key, slice_data(new_pivot_key), slice_length(new_pivot_key));
    entry->key_size                            = slice_length(new_pivot_key);
+   entry->key_indirect                        = FALSE;
    entry->pivot_data.child_addr               = new_addr;
    entry->pivot_data.num_kvs_in_subtree       = kv_pairs;
    entry->pivot_data.key_bytes_in_subtree     = key_bytes;
@@ -289,16 +290,21 @@ btree_fill_leaf_entry(const btree_config *cfg,
                       btree_hdr          *hdr,
                       leaf_entry         *entry,
                       slice               key,
-                      slice               message)
+                      message             msg)
 {
-   debug_assert(pointer_byte_offset(entry, leaf_entry_size(key, message))
+   debug_assert(pointer_byte_offset(entry, leaf_entry_size(key, msg))
                 <= pointer_byte_offset(hdr, btree_page_size(cfg)));
    memcpy(entry->key_and_message, slice_data(key), slice_length(key));
    memcpy(entry->key_and_message + slice_length(key),
-          slice_data(message),
-          slice_length(message));
+          message_data(msg),
+          message_length(msg));
    entry->key_size     = slice_length(key);
-   entry->message_size = slice_length(message);
+   entry->key_indirect = FALSE;
+   entry->type         = message_class(msg);
+   /* This assertion ensures that entry->type is large enough to hold type. */
+   debug_assert(entry->type == message_class(msg));
+   entry->message_size = message_length(msg);
+   entry->message_indirect = FALSE;
 }
 
 static inline bool
@@ -306,7 +312,7 @@ btree_can_set_leaf_entry(const btree_config *cfg,
                          const btree_hdr    *hdr,
                          table_index         k,
                          slice               new_key,
-                         slice               new_message)
+                         message             new_message)
 {
    if (hdr->num_entries < k)
       return FALSE;
@@ -335,7 +341,7 @@ btree_set_leaf_entry(const btree_config *cfg,
                      btree_hdr          *hdr,
                      table_index         k,
                      slice               new_key,
-                     slice               new_message)
+                     message             new_message)
 {
    if (k < hdr->num_entries) {
       leaf_entry *old_entry = btree_get_leaf_entry(cfg, hdr, k);
@@ -379,7 +385,7 @@ btree_insert_leaf_entry(const btree_config *cfg,
                         btree_hdr          *hdr,
                         table_index         k,
                         slice               new_key,
-                        slice               new_message)
+                        message             new_message)
 {
    debug_assert(k <= hdr->num_entries);
    bool succeeded =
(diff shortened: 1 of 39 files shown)


src/trunk.c
```diff
@@ -4366,24 +4366,27 @@ trunk_btree_skiperator_init(trunk_handle           *spl,
...
-      iterator_at_end(&skip_itor->itor[skip_itor->curr].super, &at_end);
+      iterator_at_end((iterator *)&skip_itor->itor[skip_itor->curr].super,
+                      &at_end);
    else
       at_end = TRUE;
 
    while (skip_itor->curr != skip_itor->end && at_end) {
-      iterator_at_end(&skip_itor->itor[skip_itor->curr].super, &at_end);
+      iterator_at_end((iterator *)&skip_itor->itor[skip_itor->curr].super,
+                      &at_end);
```
ajhconway, line 4376: Not your fault, but {} in these if statements

src/trunk.c
```diff
@@ -4366,24 +4366,27 @@ trunk_btree_skiperator_init(trunk_handle           *spl,
...
 
 void
-trunk_btree_skiperator_get_curr(iterator *itor, slice *key, slice *data)
+trunk_btree_skiperator_get_curr(iterator *itor, slice *key, message *data)
 {
    debug_assert(itor != NULL);
    trunk_btree_skiperator *skip_itor = (trunk_btree_skiperator *)itor;
-   iterator_get_curr(&skip_itor->itor[skip_itor->curr].super, key, data);
+   iterator_get_curr(
+      (iterator *)&skip_itor->itor[skip_itor->curr].super, key, data);
```
ajhconway, line 4389: Again, I don't get this cast.

src/trunk.c
```diff
@@ -6111,9 +6116,8 @@ trunk_compacted_subbundle_lookup(trunk_handle    *spl,
             spl->stats[tid].branch_lookups[height]++;
          }
          if (local_found) {
-            slice message = writable_buffer_to_slice(data);
-            if (data_message_class(spl->cfg.data_cfg, message)
-                != MESSAGE_TYPE_UPDATE) {
+            message msg = merge_accumulator_to_message(data);
+            if (message_class(msg) != MESSAGE_TYPE_UPDATE) {
```
ajhconway, line 6120: We have a bunch of these and the complementary type check elsewhere.

I think it might be both more readable and future-proof if we had a `message_is_final`-type function that make what the purpose of this type of check is explicit.

src/trunk.c
```diff
@@ -6183,19 +6187,16 @@ trunk_pivot_lookup(trunk_handle     *spl,
 // If any change is made in here, please make similar change in
 // trunk_lookup_async
 platform_status
-trunk_lookup(trunk_handle *spl, char *key, writable_buffer *data)
+trunk_lookup(trunk_handle *spl, char *key, merge_accumulator *data)
```
ajhconway, line 6190: I find `data` to be a confusing name. What about `lookup_result` or just `result`?

src/trunk.c
```diff
@@ -6205,13 +6206,11 @@ trunk_lookup(trunk_handle *spl, char *key, writable_buffer *data)
...
       rc = trunk_memtable_lookup(spl, mt_gen, key, data);
       platform_assert_status_ok(rc);
-      if (!writable_buffer_is_null(data)) {
-         slice message = writable_buffer_to_slice(data);
-         type          = data_message_class(data_cfg, message);
-         if (type != MESSAGE_TYPE_UPDATE) {
-            found_in_memtable = TRUE;
-            goto found_final_answer_early;
-         }
+      if (!merge_accumulator_is_null(data)
```
ajhconway, line 6209: Maybe a wrapper function that conveys the higher-level meaning of this? Something like `trunk_lookup_found(result)`? (including the renaming I suggest above)

ajhconway: Then this whole statement reads as
```
if (trunk_lookup_found(result) && message_is_final(result))
```
or something similar.

tests/functional/test_functionality.c
```diff
@@ -54,17 +54,17 @@ search_for_key_via_iterator(trunk_handle *spl, slice target)
 static void
 verify_tuple(trunk_handle    *spl,
              const char      *keybuf,
-             slice            message,
+             message          mmsg,
```
ajhconway, line 57: `mmsg` -> `msg`?

tests/functional/test_functionality.c
```diff
@@ -160,9 +158,9 @@ verify_against_shadow(trunk_handle               *spl,
 
    platform_status rc, result = STATUS_OK;
 
-   uint64          i;
-   writable_buffer message;
-   writable_buffer_init(&message, spl->heap_id);
+   uint64            i;
+   merge_accumulator msgacc;
```
ajhconway, line 162: I think we should just name `merge_accumulators` as `msg` or perhaps `result`.

ajhconway (approved):
This is a great change. It's remarkable how much it cleans up some of the test code and so on.

Minor comments only.

src/btree.c
```diff
@@ -176,6 +176,7 @@ btree_fill_index_entry(const btree_config *cfg,
                 <= btree_page_size(cfg));
    memcpy(entry->key, slice_data(new_pivot_key), slice_length(new_pivot_key));
    entry->key_size                            = slice_length(new_pivot_key);
+   entry->key_indirect                        = FALSE;
```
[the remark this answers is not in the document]
rtjohnso: You mean to replace the `FALSE`?

Can we not just search for `key_indirect` or `message_indirect`?

src/btree.c
```diff
@@ -289,24 +290,29 @@ btree_fill_leaf_entry(const btree_config *cfg,
...
-          slice_data(message),
-          slice_length(message));
+          message_data(msg),
+          message_length(msg));
    entry->key_size     = slice_length(key);
-   entry->message_size = slice_length(message);
+   entry->key_indirect = FALSE;
+   entry->type         = message_class(msg);
+   /* This assertion ensures that entry->type is large enough to hold type. */
+   debug_assert(entry->type == message_class(msg));
```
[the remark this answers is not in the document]
rtjohnso: I've googled and futzed around and cannot find a way to write a static assert about bitfields directly.  :/

Instead, I've written an assertion about the `#define` of the number of bits in the field.

src/data_internal.c
```diff
@@ -1,2 +1,8 @@
 
 #include "data_internal.h"
+
+message NULL_MESSAGE = {.type = MESSAGE_TYPE_INVALID,
```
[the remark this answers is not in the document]
rtjohnso: I think we should think about this.

Note that data.h never deals with null messages.  We pass only non-null messages to the callback functions, and they are not allowed to set messages to the null state.

So null messages really are inited-but-invalid.

src/data_internal.h
```diff
@@ -13,41 +13,204 @@
...
+      return 1;
+   else
+      return slice_lex_cmp(a.data, b.data);
+}
+
+static inline const char *
+message_class_string(message msg)
+{
+   return message_type_string(msg.type);
+}
```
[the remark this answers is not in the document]
rtjohnso: Trying to encourage users to access message through accessor functions.

PR #332, opened 2022-03-08
build into directories based on flags

rtjohnso:
This PR attempts to make builds more reliable, even as we change things like `CFLAGS`, `LDFLAGS`, `CC`, etc.

All builds go into `build/<hash>`, where `<hash>` is a hash of all the relevant make variables, such as `CFLAGS`, etc.

It then creates symlinks `bin`, `obj`, and `lib` in the top-level directory.

Unfortunately, this doesn't work with target-specific variables, so I had to get rid of `make release`, `make debug`, and `make debug-log`.  Instead, you can do 
`make D=<anything>`
or
`make DL=<anything>`
for debug or debug-log mode, respectively.

For debug builds, the top-level directories are `bin-debug`, `lib-debug`, and `obj-debug`.

The makefile also saves all relevant make variables in `build/<hash>/flags`.

`make clean` blows away the whole `build` directory.

PR #369, opened 2022-03-18
Fixes applied on top of Rob's Makefile rework to get CI tests to run.

gapisback:
These are edits on top of Rob's` Makefile` work. ... to run thru CI-jobs, and see whether they run to completion.

With this one single commit, we are now able to run thru CI successfully.

Rob: Here is what I suggest you can do:

1. Review this single commit, and confirm that the diffs are acceptable to you.
2. Cherry-pick this commit to your dev-branch, and move the code to merge to /main. (I am fairly confident that they will go thru CI.)

Peeled-off following issues, which can be done in /main after your main PR #332 integrates.

1. (rob) https://github.com/vmware/splinterdb/issues/379 - make install fails
2. (gapisback) https://github.com/vmware/splinterdb/issues/378 - Code cleanup of indentation in `test.sh` script
3. (gapisback) https://github.com/vmware/splinterdb/issues/377 - Code cleanup of CI-scripts' use of `debug` flags

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,6 @@
 #
 # To build the image from a local source checkout:
 # $ docker build -t splinterdb .
-# To then run the splinterdb-cli within the build image:
-# $ docker run --rm splinterdb bin/splinterdb-cli --help
 # or to run tests:
 # $ docker run --rm splinterdb ./test.sh
 
@@ -42,10 +40,14 @@ COPY --from=build /splinterdb-install/lib/ /usr/local/lib/
 COPY --from=build /splinterdb-install/include/splinterdb/ /usr/local/include/splinterdb/
 
 # Copy over the test binaries under bin/ (recursively) and the test script
-COPY --from=build /splinterdb-src/bin/ /splinterdb/bin/
+# (Default BUILD_DIR is 'build'.)
+COPY --from=build /splinterdb-src/build/bin/ /splinterdb/build/bin/
 COPY --from=build /splinterdb-src/test.sh /splinterdb/test.sh
+
 # TODO: Currently driver_test dynamically links against the relative path lib/libsplinterdb.so
 # Instead we should link driver_test statically against libsplinterdb.a so that this hack isn't necessary
-RUN mkdir /splinterdb/lib && ln -s /usr/local/lib/libsplinterdb.so /splinterdb/lib/libsplinterdb.so
+# As all test binaries are linked against the libraries produced during the build,
+# adjust the lib-paths to drive off of default $BUILD_DIR dir; i.e. 'build'.
+RUN mkdir -p /splinterdb/build/lib && ln -s /usr/local/lib/libsplinterdb.so /splinterdb/build/lib/libsplinterdb.so
 
 WORKDIR "/splinterdb"
--- a/Makefile
+++ b/Makefile
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
 #
 
 help::
-	@echo Environment variables controlling the build
-	@echo '  BUILD_DIR: Base name for build output (Default: "build").'
+	@echo Environment variables controlling the build:
+	@echo '  BUILD_DIR: Base dir name for build outputs (Default: "build").'
 	@echo '    $$(BUILD_DIR)/obj: object files'
 	@echo '    $$(BUILD_DIR)/lib: libraries'
 	@echo '    $$(BUILD_DIR)/bin: executables'
-	@echo '  Note: setting BUILD_MODE and other flags may further modify BUILD_DIR'
-	@echo
 
 ifndef BUILD_DIR
-BUILD_DIR := build
+   BUILD_DIR := build
 endif
 
 #
 # Debug mode
 #
-ifndef DEBUGMODE
-DEBUGMODE=release
+ifndef BUILD_MODE
+   BUILD_MODE=release
 endif
 
-ifeq "$(DEBUGMODE)" "debug"
-CFLAGS  += -DSPLINTER_DEBUG
+ifeq "$(BUILD_MODE)" "debug"
+   CFLAGS  += -DSPLINTER_DEBUG
 #LDFLAGS +=
-BUILD_SUFFIX:=$(BUILD_SUFFIX)-debug
-else ifeq "$(DEBUGMODE)" "release"
-CFLAGS   += -Ofast -flto
-LDFLAGS  += -Ofast -flto
+   BUILD_DIR:=$(BUILD_DIR)-debug
+else ifeq "$(BUILD_MODE)" "release"
+   CFLAGS   += -Ofast -flto
+   LDFLAGS  += -Ofast -flto
+else ifeq "$(BUILD_MODE)" "optimized-debug"
+   CFLAGS  += -DSPLINTER_DEBUG
+   CFLAGS  += -Ofast -flto
+   LDFLAGS += -Ofast -flto
+   BUILD_DIR := $(BUILD_DIR)-optimized-debug
 else
-$(error Unknown DEBUGMODE "$(DEBUGMODE)".  Valid options are "debug" and "release".  Default is "release")
+   $(error Unknown BUILD_MODE "$(BUILD_MODE)".  Valid options are "debug", "optimized-debug", and "release".  Default is "release")
 endif
 
 help::
-	@echo '  DEBUGMODE: "release" or "debug" (Default: "release")'
+	@echo '  BUILD_MODE: "release" or "debug" (Default: "release")'
 
-#
-# address sanitizer
-#
+# ************************************************************************
+# Address sanitizer
+#   - Ctests will be silently skipped with clang builds. (Known issue.)
+#   - Use gcc to build in Asan mode to run unit-tests.
+#   - Tests will run slow in address sanitizer builds.
 ifndef BUILD_ASAN
-BUILD_ASAN=0
+   BUILD_ASAN=0
 endif
 
 ifeq "$(BUILD_ASAN)" "1"
-CFLAGS  += -fsanitize=address
-LDFLAGS += -fsanitize=address
-BUILD_DIR:=$(BUILD_DIR)-asan
-else ifeq "$(BUILD_ASAN)" "0"
-else
-$(error Unknown BUILD_ASAN mode "$(BUILD_ASAN)".  Valid values are "0" or "1". Default is "0")
+   CFLAGS  += -fsanitize=address
+   LDFLAGS += -fsanitize=address
+   BUILD_DIR:=$(BUILD_DIR)-asan
+else ifneq "$(BUILD_ASAN)" "0"
+   $(error Unknown BUILD_ASAN mode "$(BUILD_ASAN)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
 	@echo '  BUILD_ASAN={0,1}: Disable/enable address-sanitizer (Default: disabled)'
+	@echo '                    Use gcc to run unit-tests with ASAN-builds.'
 
-#
-# memory sanitizer
-#
+# ************************************************************************
+# Memory sanitizer
+#   - Builds will fail with gcc due to compiler error. Use clang instead.
+#   - Tests will run even slower in memory sanitizer builds.
 ifndef BUILD_MSAN
-BUILD_MSAN=0
+   BUILD_MSAN=0
 endif
 
 ifeq "$(BUILD_MSAN)" "1"
-CFLAGS  += -fsanitize=memory
-LDFLAGS += -fsanitize=memory
-BUILD_DIR:=$(BUILD_DIR)-msan
-else ifeq "$(BUILD_MSAN)" "0"
-else
-$(error Unknown BUILD_MSAN mode "$(BUILD_MSAN)".  Valid values are "0" or "1". Default is "0")
+   CFLAGS  += -fsanitize=memory
+   LDFLAGS += -fsanitize=memory
+   BUILD_DIR:=$(BUILD_DIR)-msan
+else ifneq "$(BUILD_MSAN)" "0"
+   $(error Unknown BUILD_MSAN mode "$(BUILD_MSAN)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
 	@echo '  BUILD_MSAN={0,1}: Disable/enable memory-sanitizer (Default: disabled)'
+	@echo '                    Use clang for MSAN-builds.'
 
 #
 # Verbosity
 #
 ifndef BUILD_VERBOSE
-BUILD_VERBOSE=0
+   BUILD_VERBOSE=0
 endif
 
 ifeq "$(BUILD_VERBOSE)" "1"
-COMMAND=
-PROLIX=@echo
-BRIEF=@ >/dev/null echo
-BRIEF_FORMATTED=@ >/dev/null echo
-BRIEF_PARTIAL=@ >/dev/null echo
+   COMMAND=
+   PROLIX=@echo
+   BRIEF=@ >/dev/null echo
+   BRIEF_FORMATTED=@ >/dev/null echo
+   BRIEF_PARTIAL=@ >/dev/null echo
 else ifeq "$(BUILD_VERBOSE)" "0"
-COMMAND=@
-PROLIX=@ >/dev/null echo
-BRIEF=@echo
-BRIEF_FORMATTED=@printf
-BRIEF_PARTIAL=@echo -n
+   COMMAND=@
+   PROLIX=@ >/dev/null echo
+   BRIEF=@echo
+   BRIEF_FORMATTED=@printf
+   BRIEF_PARTIAL=@echo -n
 else
-$(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
+   $(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
+	@echo '  Note: setting BUILD_MODE and other flags may further modify BUILD_DIR'
+	@echo
 	@echo '  BUILD_VERBOSE={0,1}: Disable/enable verbose output (Default: disabled)'
+
+ifeq "$(BUILD_VERBOSE)" "1"
+
+	@echo '  '
+	@echo 'Examples:'
+	@echo '  - Default release build: make'
+	@echo '  - Debug build: BUILD_DEBUG=1 make'
+	@echo '  - Build release binary and run all tests: make all run-tests'
+endif
 
 
 ###################################################################
 # BUILD DIRECTORIES AND FILES
 #
-
-OBJDIR    = obj$(BUILD_SUFFIX)
-BINDIR    = bin$(BUILD_SUFFIX)
-LIBDIR    = lib$(BUILD_SUFFIX)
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -227,7 +225,7 @@ CONFIG_FILE = $(CONFIG_FILE_PREFIX)$(CONFIG_HASH)
 mismatched_config_file_check: | $(BUILD_DIR)/.
 	$(BRIEF_PARTIAL) Checking for mismatched config...
 	$(COMMAND) ls $(CONFIG_FILE_PREFIX)* 2>/dev/null | grep -v $(CONFIG_FILE) | xargs -ri sh -c 'echo "Mismatched config file \"{}\" detected.  You need to \"make clean\"."; false'
-	$(BRIEF) No mismatched config found
+	$(BRIEF) No mismatched config found. Nothing to build.
```
ajhconway: This target still runs as part of other targets which may still build stuff, right? In that case, I wouldn't add this message.

test.sh
```diff
@@ -6,8 +6,37 @@
...
-# location of binaries
-BINDIR="${BINDIR:-bin}"
+# location of binaries, which live under $BUILD_DIR, if set.
+build_dir="${BUILD_DIR:-build}"
+BINDIR="${BINDIR:-${build_dir}/bin}"
+
+echo "$Me: build_dir='${build_dir}', BINDIR='${BINDIR}'"
+
+# Location of binaries, which live under $BUILD_DIR, if BINDIR is not set.
+# This means: If BINDIR was -not- set ...
```
ajhconway, line 14: I don't know what this means, didn't this assignment just happen on 11?

test.sh
```diff
@@ -130,7 +159,7 @@ function nightly_functionality_stress_tests() {
     local dbname="splinter_test.functionality.db"
     echo "$Me: Run ${test_name} with ${n_mills} million rows, on ${ntables} tables, with ${cache_size} GiB cache"
     run_with_timing "Functionality Stress test ${test_descr}" \
-            $BINDIR/driver_test splinter_test --functionality  ${num_rows} 1000 \
+            "$BINDIR"/driver_test splinter_test --functionality  ${num_rows} 1000 \
```
ajhconway, line 159: These changes from `$BINDIR` to `"$BINDIR"` are cosmetic?

Makefile
```diff
@@ -103,69 +101,84 @@ endif
...
+   BRIEF_FORMATTED=@printf
+   BRIEF_PARTIAL=@echo -n
 else
-$(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
+   $(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
+	@echo '  Note: setting BUILD_MODE and other flags may further modify BUILD_DIR'
+	@echo
```
ajhconway, line 170: Why is this help comment here? It feels like it belongs in the `BUILD_DIR` block.

Dockerfile
```diff
@@ -13,8 +13,6 @@
 #
 # To build the image from a local source checkout:
 # $ docker build -t splinterdb .
-# To then run the splinterdb-cli within the build image:
-# $ docker run --rm splinterdb bin/splinterdb-cli --help
```
ajhconway, line 17: Why is this going away?

ajhconway (approved):
Minor stuff only. Ship it.

Dockerfile
gapisback: @ajhconway - Gabriel pulled-out support for Rust-based CLI tooling under PR #307 @ SHA 916625a.

This comment in the `Dockerfile` is a straggling reference.

Makefile
gapisback: It's here because (seems like) `make help` emits messages as it encounters them in the `Makefile`.

The `help` output will now look like this with the message as you see it on L 172;

```
Environment variables controlling the build:
  BUILD_DIR: Base dir name for build outputs (Default: "build").
    $(BUILD_DIR)/obj: object files
    $(BUILD_DIR)/lib: libraries
    $(BUILD_DIR)/bin: executables
  BUILD_MODE: "release" or "debug" (Default: "release")
  BUILD_ASAN={0,1}: Disable/enable address-sanitizer (Default: disabled)
                    Use gcc to run unit-tests with ASAN-builds.
  BUILD_MSAN={0,1}: Disable/enable memory-sanitizer (Default: disabled)
                    Use clang for MSAN-builds.
  Note: setting BUILD_MODE and other flags may further modify BUILD_DIR

  BUILD_VERBOSE={0,1}: Disable/enable verbose output (Default: disabled)
```

I wanted this "Note: " line to appear -after- all the help messages for other `BUILD_*` env-vars are emitted.

test.sh
gapisback: I'll fix the comment to read like so: (Would that be more clear?)

```
 14 # If BINDIR -was- set in the user's env, we need to drive off of that.
 15 # To avoid raising a script error by referencing this variable if it was
 16 # -not-set- we just initialized it above to this value. Thus, if the
 17 # following is true, it means the env-var BINDIR was -not- set already.
```

I concede that this script's logic to reliably check if env-var `BINDIR` was set, and not fail, and this comment. ... is cumbersome.

Suggestions to improve this (w/o failing if BINDIR is not set in the env) are welcome!

test.sh
gapisback: Not exactly. Otherwise we get errors from `shellcheck` and that step will fail.

test.sh
```diff
@@ -229,23 +254,10 @@ function nightly_sync_perf_tests() {
...
-                                          --max-async-inflight 0 \
-                                          --num-insert-threads ${nins_t} \
-                                          --num-lookup-threads ${nlookup_t} \
-                                          --num-range-lookup-threads ${nrange_lookup_t} \
-                                          --lookup-positive-percent 10 \
-                                          --db-capacity-gib 60 \
-                                          --db-location ${dbname}
-    rm ${dbname}
-
->>>>>>> respond to all comments and feature requests
```
gapisback: Merge issue ... Deleted messed-up code.

(NOTE: I am still wading thru diffs, and merges, trying to stabilize CI-runs. So this commit is not quite ready for a review.

Ideally, once I have got all CI runs coming out clean, @rtjohnso would want to scrutinize this commit's diffs, and cherry-pick it into his branch, and then push it forward. Until then, this work is really under dev.)

Makefile
gapisback: You are right ... I discovered that for some usages, this extra message is -actually- wrong and misleading.

I seem to have taken it out in my branch. This diff-block is possibly stale.

test.sh
ajhconway: Ack

Dockerfile
```diff
@@ -42,10 +40,14 @@ COPY --from=build /splinterdb-install/lib/ /usr/local/lib/
 COPY --from=build /splinterdb-install/include/splinterdb/ /usr/local/include/splinterdb/
 
 # Copy over the test binaries under bin/ (recursively) and the test script
-COPY --from=build /splinterdb-src/bin/ /splinterdb/bin/
+# (Default BUILD_DIR is 'build'.)
+COPY --from=build /splinterdb-src/build/bin/ /splinterdb/build/bin/
```
gapisback, line 44: This is needed because, by default, the build outputs come to `build/bin`. W/o this fix, one of the CI-jobs that runs thru Docker image building fails to run any tests.

Dockerfile
```diff
@@ -42,10 +40,14 @@ COPY --from=build /splinterdb-install/lib/ /usr/local/lib/
...
 COPY --from=build /splinterdb-src/test.sh /splinterdb/test.sh
+
 # TODO: Currently driver_test dynamically links against the relative path lib/libsplinterdb.so
 # Instead we should link driver_test statically against libsplinterdb.a so that this hack isn't necessary
-RUN mkdir /splinterdb/lib && ln -s /usr/local/lib/libsplinterdb.so /splinterdb/lib/libsplinterdb.so
+# As all test binaries are linked against the libraries produced during the build,
+# adjust the lib-paths to drive off of default $BUILD_DIR dir; i.e. 'build'.
+RUN mkdir -p /splinterdb/build/lib && ln -s /usr/local/lib/libsplinterdb.so /splinterdb/build/lib/libsplinterdb.so
```
gapisback, line 51: Same reason as above ..

Makefile
```diff
@@ -1,7 +1,7 @@
 # Copyright 2018-2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-.DEFAULT_GOAL := release
+.DEFAULT_GOAL := all
```
gapisback, line 4: @rtjohnso - **NOTE**: All these diffs in `Makefile` show up as I rebased v/s /main. Seems like this reapplied the diffs you had introduced, too. (Not sure how hard it will make your effort to cherry-pick this to your branch.)

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
+	@echo '  BUILD_DIR: Base dir name for build outputs (Default: "build").'
 	@echo '    $$(BUILD_DIR)/obj: object files'
 	@echo '    $$(BUILD_DIR)/lib: libraries'
 	@echo '    $$(BUILD_DIR)/bin: executables'
-	@echo '  Note: setting BUILD_MODE and other flags may further modify BUILD_DIR'
-	@echo
 
 ifndef BUILD_DIR
-BUILD_DIR := build
+   BUILD_DIR := build
```
gapisback, line 78: I consistently changed `if` blocks to indent their body, for readability.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
-CFLAGS  += -fsanitize=address
-LDFLAGS += -fsanitize=address
-BUILD_DIR:=$(BUILD_DIR)-asan
-else ifeq "$(BUILD_ASAN)" "0"
-else
-$(error Unknown BUILD_ASAN mode "$(BUILD_ASAN)".  Valid values are "0" or "1". Default is "0")
+   CFLAGS  += -fsanitize=address
+   LDFLAGS += -fsanitize=address
+   BUILD_DIR:=$(BUILD_DIR)-asan
+else ifneq "$(BUILD_ASAN)" "0"
```
gapisback, line 120: Changed to `ifneq` to remove one conditional block.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
+	@echo '  BUILD_MODE: "release" or "debug" (Default: "release")'
 
-#
-# address sanitizer
-#
+# ************************************************************************
+# Address sanitizer
+#   - Ctests will be silently skipped with clang builds. (Known issue.)
+#   - Use gcc to build in Asan mode to run unit-tests.
+#   - Tests will run slow in address sanitizer builds.
```
gapisback, line 111: Brought back these comment blocks from old code.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
+   CFLAGS  += -fsanitize=address
+   LDFLAGS += -fsanitize=address
+   BUILD_DIR:=$(BUILD_DIR)-asan
+else ifneq "$(BUILD_ASAN)" "0"
+   $(error Unknown BUILD_ASAN mode "$(BUILD_ASAN)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
 	@echo '  BUILD_ASAN={0,1}: Disable/enable address-sanitizer (Default: disabled)'
+	@echo '                    Use gcc to run unit-tests with ASAN-builds.'
```
gapisback, line 126: Added this line, so users know of this limitation.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
 
-#
-# memory sanitizer
-#
+# ************************************************************************
+# Memory sanitizer
+#   - Builds will fail with gcc due to compiler error. Use clang instead.
+#   - Tests will run even slower in memory sanitizer builds.
```
gapisback, line 131: Brought back this comment block.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
-CFLAGS  += -fsanitize=memory
-LDFLAGS += -fsanitize=memory
-BUILD_DIR:=$(BUILD_DIR)-msan
-else ifeq "$(BUILD_MSAN)" "0"
-else
-$(error Unknown BUILD_MSAN mode "$(BUILD_MSAN)".  Valid values are "0" or "1". Default is "0")
+   CFLAGS  += -fsanitize=memory
+   LDFLAGS += -fsanitize=memory
+   BUILD_DIR:=$(BUILD_DIR)-msan
+else ifneq "$(BUILD_MSAN)" "0"
```
gapisback, line 140: Changed to use `ifneq` check.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
+   $(error Unknown BUILD_MSAN mode "$(BUILD_MSAN)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
 	@echo '  BUILD_MSAN={0,1}: Disable/enable memory-sanitizer (Default: disabled)'
+	@echo '                    Use clang for MSAN-builds.'
```
gapisback, line 146: Added this line.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
+   BRIEF=@echo
+   BRIEF_FORMATTED=@printf
+   BRIEF_PARTIAL=@echo -n
 else
-$(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
+   $(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
 endif
 
 help::
+	@echo '  Note: setting BUILD_MODE and other flags may further modify BUILD_DIR'
```
gapisback, line 172: Moved this down here, for the reasons discussed in the comment w/ Alex.

Makefile
```diff
@@ -68,111 +68,127 @@ DEPFLAGS  = -MMD -MP
...
+	@echo
 	@echo '  BUILD_VERBOSE={0,1}: Disable/enable verbose output (Default: disabled)'
 
+ifeq "$(BUILD_VERBOSE)" "1"
```
gapisback, line 176: Report a few usage examples under `verbose help` output.

Makefile
```diff
@@ -200,9 +216,9 @@ UNIT_TESTBINS=$(UNIT_TESTBIN_SRC:$(TESTS_DIR)/%_test.c=$(BINDIR)/%_test)
 # The main targets
 #
 
-all: libs tests $(EXTRA_TARGETS)
+all: libs all-tests $(EXTRA_TARGETS)
 libs: $(LIBDIR)/libsplinterdb.so $(LIBDIR)/libsplinterdb.a
-tests: $(BINDIR)/driver_test $(BINDIR)/unit_test $(UNIT_TESTBINS)
+all-tests: $(BINDIR)/driver_test $(BINDIR)/unit_test $(UNIT_TESTBINS)
```
gapisback, line 221: This and L219 ... seems like merge errors on my part when rebasing v/s /main.

Makefile
```diff
@@ -222,14 +238,17 @@ CONFIG_FILE_PREFIX=.Makefile.config$(BUILD_SUFFIX).
...
-	$(BRIEF) No mismatched config found
-
+	$(BRIEF) No mismatched config found in $(CONFIG_FILE)
 
-$(CONFIG_FILE): | mismatched_config_file_check
+# Search for CC also in 'env' command below, while storing config to a file.
+# Otherwise, 'make' may fail if user's environment has no BUILD_ env-vars set.
+$(CONFIG_FILE): | $(BUILD_DIR)/. mismatched_config_file_check
 	$(BRIEF) Saving config to $@
+	$(COMMAND) env | grep -E "BUILD_|CC"         >> $@
```
gapisback, line 251: Am dumping any `BUILD_` env-vars to the config file. Need to pull-in `CC` with this `grep`, otherwise if no `BUILD_` env-vars are send, this `grep` will fail. That causes whole `make` to fail.

Only question is: Do we still need L 252 below? I think it's ok to leave both ... doesn't hurt anything.

Makefile
```diff
@@ -265,25 +284,25 @@ $(BINDIR)/%/.: $(CONFIG_FILE)
 
 COMPILE.c = $(CC) $(DEPFLAGS) -MT $@ -MF $(OBJDIR)/$*.d $(CFLAGS) $(INCLUDE) $(TARGET_ARCH) -c
 
-$(OBJDIR)/%.o: %.c | $$(@D)/.
-	$(BRIEF_FORMATTED) "%-20s %-40s [%s]\n" COMPILING $< $@
+$(OBJDIR)/%.o: %.c | $$(@D)/. $(CONFIG_FILE)
+	$(BRIEF_FORMATTED) "%-20s %-50s [%s]\n" Compiling $< $@
```
gapisback, line 288: Changes these terms to `Compiling` ... etc.

test.sh
```diff
@@ -6,8 +6,37 @@
 Me=$(basename "$0")
 set -euo pipefail
 
-# location of binaries
-BINDIR="${BINDIR:-bin}"
+# location of binaries, which live under $BUILD_DIR, if set.
```
gapisback, line 9: This whole block was needed just to allow someone to run `test.sh` by itself.

One use-case is to export these `BUILD_` env-vars in my session, do the `make` and then do the `test.sh` to run tests.

I don't recall fully now, but it _could be the case_ that CI-jobs may run even w/o this fix. Certainly, manual invocation of `test.sh` w/o specifying `BINDIR` on the env will certainly fail.

test.sh
```diff
@@ -6,8 +6,37 @@
...
+build_dir="${BUILD_DIR:-build}"
+BINDIR="${BINDIR:-${build_dir}/bin}"
+
+# Location of binaries, which live under $BUILD_DIR, if BINDIR is not set.
+# If BINDIR -was- set in the user's env, we need to drive off of that.
+# To avoid raising a script error by referencing this variable if it was
+# -not-set- we just initialized it above to this value. Thus, if the
+# following is true, it means the env-var BINDIR was -not- set already.
+if [ "${BINDIR}" == "${build_dir}/bin" ]; then
```
gapisback, line 18: This check "Is BINDIR set in the env-var? Otherwise, synthesize it here ..." and the preceding comment block is a bit tacky. Suggestions welcome to rework that, but I think it's good enough the way it is now.

test.sh
gapisback: Btw, I recall now ... In all CI-scripts, the invocation is `make all run-tests`, with an earlier change (See PR #373).

Prior to that commit of 0edbba6, the CI-scripts used to execute the tests as: `make && ./test.sh`.

With that usage, this chunk of code in `test.sh` to establish `BINDIR` was necessary.

With the rework to run tests using `make all run-tests`, we really don't need this chunk as part of this change set, but to not regress other dev's usage, we should include this chunk in this change-set.

rtjohnso:
I think we can close this PR.  I replicated its changes in my PR #332, which is now merged.

gapisback:
Changes have been integrated with PR #332 and merged to /main under this commit:

```
a5f2a1b Rob Johnson 23 minutes ago Sat, 19-Mar-2022, 06:09:12 PM (Authored: Sat, 19-Mar-2022, 06:09:12 PM)
 build into directories based on flags (#332)

```

PR #372, opened 2022-03-18
(#370) Add 'help' to Makefile, report minimal usage msg.

gapisback:
Pre-stage support for 'help' target to Makefile, so that associated
changes to CI-build scripts can be pre-staged (to address issue #371).
All this is needed to smooth the integration of larger Makefile
rework coming up under PR #332 .

Issue #371, opened 2022-03-18
Pre-stage CI-jobs' script changes that are developed under PR#332 to merge to /main after Makefile changes to-be-added by issue #370.


gapisback:
### Is your feature request related to a problem? Please describe.

PR #332 is adding a large collection of changes to `Makefile` and it changes the way the `make` commands have to be issued to build the product. These invocation changes then cascade into CI-jobs' scripts, which currently issue commands like:

` make debug && ./test.sh`

This will then change to: `make $MAKE_HELP all run-tests`.

If we can integrate the CI-jobs' script changes up-front, it will make the merge of changes from PR #332 lot smoother.

### Describe the solution you'd like

Port the changes to CI-scripts out of PR #332 to this issue, as an independent early commit. 
Changes like these diffs will come into /main:

```
+       BUILD_ASAN: "1"
+       BUILD_MODE: debug
+       BUILD_VERBOSE: "1"
```

and depending on the build-job, for debug builds, we will be adding:

```
      params:
+       BUILD_MODE: debug
+       BUILD_VERBOSE: "1"
        CC: clang
        INCLUDE_SLOW_TESTS: "false"
        LD: clang
+       MAKE_HELP: help
        RUN_NIGHTLY_TESTS: "false"
```

This will smoothen out the path for merging changes from the said PR, by pre-staging build-job commands with the upcoming syntax.

### Describe alternatives you've considered

Alternative would be to wait for the entire collection of `Makefile`, `test.sh` and CI-jobs scripts changes to all merge in one go, which has a greater chance of regressing CI-jobs. 

By pre-staging the commands issued by CI-jobs scripts we can then pre-test upcoming changes to ensure that they will not regress CI-jobs.

PR #373, opened 2022-03-19
(#371) Pre-stage CI-jobs' script changes to work with Makefile changes.

gapisback:
Upcoming PR #332 changes Makefile targets to build the library. Those
changes will require changes to the commands invoked by CI-jobs to build
and test the library. This commit pre-stages those CI-build script
changes, so that the upcoming Makefile changes can be tested independently
before they are merged to /main. (This change depends on PR #372, which
adds minimal support for 'help' target to Makefile.)

### DIFF
--- a/ci/steps.lib.yml
+++ b/ci/steps.lib.yml
@@ -104,24 +104,41 @@ config:
     LD: #@ compiler
     INCLUDE_SLOW_TESTS: #@ str(not quick).lower()
     RUN_NIGHTLY_TESTS: #@ str(is_nightly).lower()
+
+    #! Exercise 'make help' in quick tests mode, to ensure 'help' still works.
+    #@ if quick:
+    MAKE_HELP: help
+    #@ else:
+    MAKE_HELP: ""
+    #@ end
+
+    #@ if is_debug:
+    BUILD_MODE: "debug"
+    #! Add this, just to ensure that verbose build outputs feature doesn't regress
+    BUILD_VERBOSE: "1"
+    #@ end
+
     #@ if sanitize == "asan":
     DEFAULT_CFLAGS: "-fsanitize=address"
     DEFAULT_LDFLAGS: "-fsanitize=address"
+    BUILD_ASAN: "1"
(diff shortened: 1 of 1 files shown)


gapisback:
Hi, @rtjohnso - I assigned this CI-scripts review to you, which goes on top of (after) the fix for PR #372 . That one is also assigned to you.

This is a song-and-dance we have to play to smooth out the pavement for integrating your `Makefile` changes upcoming under PR #332. 

If you can do a quick review & sign-off of the previous PR; i.e. PR #372 ), I have to do some manual testing of this PR's changes to apply them to the pipeline, after merging that to /main.

It's confusing ... and late on a Friday. Let's see what you can do ... or I'll wait till Monday.

ci/steps.lib.yml
```diff
@@ -104,24 +104,41 @@ config:
...
     #@ end
+
   run:
     path: sh
     dir: #@ input_name
     args:
     - "-c"
     #@ if is_debug:
-    - "make debug && ./test.sh"
+    - "make $MAKE_HELP debug all run-tests"
```
ajhconway, line 139: This target will still need to change when @rtjohnso's PR lands, right?

ajhconway (approved):
My understanding of this changeset is that it tries to incorporate in a `/dev/null` sort of way as many of the changes from #332 as possible. It looks good to me.

ci/steps.lib.yml
gapisback: Well, @ajhconway -- Actually the hope was that for the initial merge of @rtjohnso 's `Makefile` rework, this target need not change.

Rob has added support for `make debug` as an acceptable target, so the thinking is that that PR will merge cleanly.

We will then have to come back and rip out these older targets, to smooth out the entire CI-usage. It's a multi-pass cleanup.

gapisback:
Thanks for reviewing this at short-notice, @ajhconway and also for looking at https://github.com/vmware/splinterdb/pull/372.

I am going to merge this to /main, so I can go back to re-stabilizing Rob's `Makefile` rework on top of these CI-infra changes.

PR #375, opened 2022-03-19
Add ownership to `platform_mutex`, fix/cleanup bundle accounting

ajhconway:
Adds an `owner` field to `platform_mutex`, which is used for asserting that threads are correctly locking and unlocking the mutex.

Fixes a bug where `trunk_bundle_live` and `trunk_bundle_live_for_pivot` would answer incorrectly. This is because they used the bundle start branch to determine liveness, but accessing the start branch of a dead bundle results in undefined behavior.

Partially fixes the convoluted logic in `trunk_flush_fullest`, where the return value would frequently be FALSE, despite a successful flush.

### DIFF
--- a/src/platform_linux/platform.c
+++ b/src/platform_linux/platform.c
@@ -137,24 +137,21 @@ platform_thread_join(platform_thread thread)
 }
 
 platform_status
-platform_mutex_init(platform_mutex    *mu,
+platform_mutex_init(platform_mutex    *lock,
                     platform_module_id UNUSED_PARAM(module_id),
                     platform_heap_id   UNUSED_PARAM(heap_id))
 {
-   int ret;
-
-   ret = pthread_mutex_init(mu, NULL);
-
+   int ret     = pthread_mutex_init(&lock->mutex, NULL);
+   lock->owner = INVALID_TID;
    return CONST_STATUS(ret);
 }
 
 platform_status
-platform_mutex_destroy(platform_mutex *mu)
+platform_mutex_destroy(platform_mutex *lock)
 {
-   int ret;
-
-   ret = pthread_mutex_destroy(mu);
-
+   // Cannot call destroy on a locked lock
+   platform_assert(lock->owner == INVALID_TID);
+   int ret = pthread_mutex_destroy(&lock->mutex);
    return CONST_STATUS(ret);
 }
 
@@ -257,14 +254,14 @@ platform_condvar_init(platform_condvar *cv, platform_heap_id heap_id)
 {
    platform_status status;
 
-   status = platform_mutex_init(&cv->lock, platform_get_module_id(), heap_id);
+   status.r = pthread_mutex_init(&cv->lock, NULL);
    if (!SUCCESS(status)) {
       return status;
    }
 
    status.r = pthread_cond_init(&cv->cond, NULL);
    if (!SUCCESS(status)) {
-      platform_mutex_destroy(&cv->lock);
+      status.r = pthread_mutex_destroy(&cv->lock);
    }
 
    return status;
--- a/src/platform_linux/platform.h
+++ b/src/platform_linux/platform.h
@@ -64,7 +64,15 @@
 /* Evaluates to sizeof() a field, f, in a C-structure, s */
 #define FSIZEOF(s, f) sizeof(((typeof(s) *)NULL)->f)
 
+/*
+ * MAX_THREADS is used primarily for convenience, where allocations made on a
+ * per-thread basis create an array with MAX_THREADS items, e.g. the
+ * trunk_stats field in trunk_handle. The task subsystem also uses a 64-bit
+ * bit-array to track thread IDs in use. This could be changed relatively
+ * easily if needed.
+ */
 #define MAX_THREADS (64)
+#define INVALID_TID (MAX_THREADS)
 
 #define HASH_SEED         (42)
 #define UNUSED_FUNCTION() __attribute__((__unused__))
--- a/src/platform_linux/platform_inline.h
+++ b/src/platform_linux/platform_inline.h
@@ -150,22 +150,30 @@ platform_semaphore_try_wait(platform_semaphore *sema)
 }
 
 static inline platform_status
-platform_mutex_lock(platform_mutex *mu)
+platform_mutex_lock(platform_mutex *lock)
 {
-   int ret;
-
-   ret = pthread_mutex_lock(mu);
-
+   int ret = pthread_mutex_lock(&lock->mutex);
+   platform_assert(lock->owner == INVALID_TID,
+                   "Found an unlocked a mutex with an existing owner:\n"
+                   "lock: %p, tid: %lu, owner: %lu\n",
+                   lock,
+                   platform_get_tid(),
+                   lock->owner);
+   lock->owner = platform_get_tid();
    return CONST_STATUS(ret);
 }
 
 static inline platform_status
-platform_mutex_unlock(platform_mutex *mu)
+platform_mutex_unlock(platform_mutex *lock)
 {
-   int ret;
-
-   ret = pthread_mutex_unlock(mu);
-
+   platform_assert(lock->owner == platform_get_tid(),
+                   "Attempt to unlock a mutex without ownership:\n"
+                   "lock: %p, tid: %lu, owner: %lu\n",
+                   lock,
+                   platform_get_tid(),
+                   lock->owner);
+   lock->owner = INVALID_TID;
+   int ret     = pthread_mutex_unlock(&lock->mutex);
    return CONST_STATUS(ret);
 }
 
--- a/src/platform_linux/platform_types.h
+++ b/src/platform_linux/platform_types.h
@@ -104,8 +104,11 @@ platform_assert_msg(platform_stream_handle stream,
 
 typedef pthread_t platform_thread;
 
-// Mutex
-typedef pthread_mutex_t platform_mutex;
+// Thread-specific mutex, with ownership tracking.
+typedef struct {
+   pthread_mutex_t mutex;
+   threadid        owner;
+} platform_mutex;
 
 // Spin lock
 typedef pthread_spinlock_t platform_spinlock;
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -664,8 +664,8 @@ void                               trunk_bundle_build_filters      (void *arg, v
 static inline void                 trunk_inc_filter                (trunk_handle *spl, routing_filter *filter);
 static inline void                 trunk_dec_filter                (trunk_handle *spl, routing_filter *filter);
 void                               trunk_compact_bundle            (void *arg, void *scratch);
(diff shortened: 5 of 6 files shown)


rtjohnso (approved):
Looks like a reasonable cleanup.

src/platform_linux/platform.c
```diff
@@ -137,24 +137,19 @@ platform_thread_join(platform_thread thread)
...
 
 platform_status
-platform_mutex_destroy(platform_mutex *mu)
+platform_mutex_destroy(platform_mutex *lock)
 {
-   int ret;
-
-   ret = pthread_mutex_destroy(mu);
-
+   int ret = pthread_mutex_destroy(&lock->mutex);
```
gapisback, line 152: For hygiene, better to also do `lock->owner = INVALID_TID;` before returning. No ? Will keep callers safe.

src/platform_linux/platform.h
```diff
@@ -65,6 +65,7 @@
 #define FSIZEOF(s, f) sizeof(((typeof(s) *)NULL)->f)
 
 #define MAX_THREADS (64)
+#define INVALID_TID (MAX_THREADS)
```
gapisback, line 68: While you are here ... Like we did in a previous round of code-cleanup ... Any chance we can comment the semantic of `MAX_THREADS` ? I guess it's a static hard-limit ... A comment that sheds light on the origins of this limit will be helpful to the new readers.

src/platform_linux/platform_inline.h
```diff
@@ -150,22 +150,32 @@ platform_semaphore_try_wait(platform_semaphore *sema)
...
-{
-   int ret;
-
-   ret = pthread_mutex_lock(mu);
+platform_mutex_lock(platform_mutex *lock)
+{
+   int ret = pthread_mutex_lock(&lock->mutex);
+   platform_assert(
+      lock->owner == INVALID_TID,
+      "Locked a mutex with an existing owner, lock: %p, tid: %lu, owner: %lu\n",
```
gapisback, line 158: I think this message wants to say: "Found a locked mutex ..."

And I found that this formatting keeps clang-format happy, while improving readability:

```
156    platform_assert(lock->owner == INVALID_TID,
157                    "Found a locked a mutex with an existing owner"
158                    ", lock: %p, tid: %lu, owner: %lu\n",
```

src/platform_linux/platform_inline.h
```diff
@@ -150,22 +150,32 @@ platform_semaphore_try_wait(platform_semaphore *sema)
...
+      lock,
+      platform_get_tid(),
+      lock->owner);
+   lock->owner = platform_get_tid();
+   platform_assert(lock->owner == platform_get_tid());
```
gapisback, line 163: Isn't this assert() redundant, especially considering the immediately preceding L162?

Are we suspecting some race here, which I don't see where it could occur, given that we have L155 where `pthread_mutex_lock() ` was acquired.

Seems like this line could be deleted? If you do wish to retain it, please do add a comment why it lives here.

src/platform_linux/platform_inline.h
```diff
@@ -150,22 +150,32 @@ platform_semaphore_try_wait(platform_semaphore *sema)
...
-
+platform_mutex_unlock(platform_mutex *lock)
+{
+   platform_assert(lock->owner == platform_get_tid(),
+                   "Attempt to unlock a mutex without ownership, lock: %p, "
+                   "tid: %lu, owner: %lu\n",
+                   lock,
+                   platform_get_tid(),
+                   lock->owner);
+   lock->owner = INVALID_TID;
```
gapisback, line 177: I've seen, in past implementations, that calls to these kinds of lock / unlock APIs proliferate, and then we run into lock-ownership sanity checks / doubts.

I see the `platform_assert()` on L171 as a way of our own book-keeping to make sure that lock / unlock sequences are kosher.

Here's another way to cross-check this more deeply. In the `pthread_mutex_lock()` [ref-man](https://pubs.opengroup.org/onlinepubs/007908799/xsh/pthread_mutex_lock.html), I see this mentioned:

```
The pthread_mutex_lock() function may fail if: 

[EDEADLK]
    The current thread already owns the mutex.
```

What if we add a `debug_assert(pthread_mutex_lock(&lock->mutex) == EDEADLK)` ?

Would n't that be a further validation that, on top of our own internal book-keeping, the pthread-lock structures themselves re-confirm that we **are truly** owning the lock?

src/platform_linux/platform_types.h
```diff
@@ -105,7 +105,10 @@ platform_assert_msg(platform_stream_handle stream,
 typedef pthread_t platform_thread;
 
 // Mutex
-typedef pthread_mutex_t platform_mutex;
+typedef struct {
```
gapisback, line 108: Maybe [?] expand on the lonely `Mutex` comment to something like:

`Thread-specific mutex, with ownership tracking.` ?

src/trunk.c
```diff
@@ -1230,6 +1230,13 @@ trunk_generation(trunk_handle *spl, page_handle *node)
    return hdr->generation;
 }
 
+static inline uint16
+trunk_pivot_start_bundle(trunk_handle *spl, page_handle *node, uint16 pivot_no)
```
gapisback, line 1234: About this collection of changes around trunk-bundle mgmt, some comments (food for thought, for follow-up):

1. What tests ran into bugs that you were fixing in this small set of changes? Are those tests already included in CI- or nightly-tests? If not, what is our plan / approach to augment test run to 'discover' those bugs, and to ensure that these changes do fix them. 

(Point being, can we plan better to integrate test-changes along with code-changes, especially for such bug-fix type code-changes,  so that we are continually improving code coverage thru tests.)

2. I noted the addition of small interface like this `trunk_pivot_start_bundle(). And found few more existing minimalist ones like`trunk_start_bundle`, `trunk_end_bundle`, `trunk_bundle_in_range` and so on. All these types of core bundle-mgmt APIs are quite amenable to CTests, where build small isolated test cases to just exercise and validate that these functions return the 'right thing' for a given set of inputs. 

That model of CTest-dev ... we may have to taken on in future dev-items, so that we start building a corpus of test-code to validate these lowermost-level APIs. (We can talk off-line ... w/ @rtjohnso )

src/trunk.c
```diff
@@ -2300,9 +2314,10 @@ trunk_bundle_live_for_pivot(trunk_handle *spl,
                             uint16        pivot_no)
 {
    debug_assert(pivot_no < trunk_num_children(spl, node));
-   trunk_bundle *bundle    = trunk_get_bundle(spl, node, bundle_no);
-   uint16        branch_no = trunk_bundle_start_branch(spl, node, bundle);
-   return trunk_branch_live_for_pivot(spl, node, branch_no, pivot_no);
+   return trunk_bundle_in_range(spl,
+                                bundle_no,
+                                trunk_pivot_start_bundle(spl, node, pivot_no),
+                                trunk_end_bundle(spl, node));
```
gapisback, line 2320: I don't have enough data-structural context on these bundle handles, but finding the use of `trunk_end_bundle` alongside `trunk_pivot_start_bundle` raised a doubt as to whether we are mixing-up interfaces here?

Would it be too wrong, or too misleading, to add a wrapper `trunk_pivot_end_bundle`, aliased to `trunk_end_bundle`, adding a prolog-comment to the new fn explaining why pivot_end is the same as end_bundle?

This way, at least, to the casual reader of this code, the calls will appear 'symmetric'.

src/trunk.c
```diff
@@ -4198,12 +4211,16 @@ trunk_flush_fullest(trunk_handle *spl, page_handle *node)
       tid = platform_get_tid();
    }
    if (trunk_pivot_needs_flush(spl, node, fullest_pivot_data)) {
-      trunk_flush(spl, node, fullest_pivot_data, FALSE);
+      bool flushed = trunk_flush(spl, node, fullest_pivot_data, FALSE);
+      if (!flushed) {
```
gapisback, line 4215: Again, the question arises as to what tests ran into this bug? Presumably a stress test?

What can we do to integrate that test / stress-test in a reduced form to cause this bug and, thus, verify that this fix does fix that issue? Overall, test stability will improve if we can  bring online some test case with this fix.

src/trunk.c
```diff
@@ -4235,7 +4252,7 @@ trunk_flush_fullest(trunk_handle *spl, page_handle *node)
       }
       return trunk_flush(spl, node, fullest_pivot_data, FALSE);
    }
-   return FALSE;
+   return TRUE;
```
gapisback, line 4255: Wow! Big BUG! I am worried that our test-/code-coverage is not up to snuff. 
We should really try to enhance the tests that are run in CI to trigger this kind of bug, to ensure that we get improved stability / coverage.

gapisback (approved):
changes are fine. Left some suggestions for code tightening, and for us to try to find ways to improve test coverage.

src/trunk.c
ajhconway: This came up on the `data-recovery` feature branch, when I was running a relatively straightforward performance test:
```
./bin/driver_test splinter_test --perf --max-async-inflight 0 --num-insert-threads 8 --num-lookup-threads 0 --num-range-lookup-threads 0 --tree-size-gib 80 --db-location /dev/nvme0n1 --set-O_DIRECT --unset-O_CREAT --cache-capacity-gib 20 --set-hugetlb --set-mlock --db-capacity-gib 120
```

I haven't recreated the issue on `main`, but the bugs definitely exist there.

All of these functions are artifacts of the current node format, which is rigid, cumbersome and bug-prone. @rtjohnso and I have tentatively discussed a new node format which will hopefully get rid of a lot of these accounting functions. I think we should punt on these tests until then. Let's have a meeting to discuss this in the not-too-distant future.

src/trunk.c
ajhconway: I understand your point and concern here, but wouldn't that interface imply that pivots have their own end bundles?

This is another place where the medium-term fix will be that the node format will change.

src/trunk.c
ajhconway: I should note that this didn't trigger an outwardly visible bug, but rather I observed a bunch of looping at L3353 in the debugger. In some sense that's worse. This was the same perf test as above.

Without tests that have access to trunk node structure, which is a big undertaking, I'm really not sure how we can directly test this sort of thing. I'll think about it.

src/trunk.c
ajhconway: This is a smaller bug than it appears for two reasons.

The first is that this function is only called when the node is full, so it always tries to flush. The only way to get to this return case is if the node flushes because pivots have too many branches and the node is no longer full. This doesn't really happen very much in practice.

The second is that returning false generally just causes the flush to be retried.

ajhconway (commented):
Responding to comments.

src/platform_linux/platform_inline.h
ajhconway: So in contravention to the man page, this just deadlocks on my system. I'm going to get rid of it.

ajhconway:
> ( I took a very quick look ... about percolating this `rc` up the code-flow. Am hoping you've done the due diligence here. )

Done.

PR #376, opened 2022-03-19
Fix filter replacement bug

ajhconway:
Fixes a bug where the filter replacement algorithm would use the same
start and end keys as the filter building logic, causing replacements to
be attempted on nodes which did not have filters built.

### DIFF
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -869,7 +869,8 @@ trunk_for_each_subtree(trunk_handle *spl, uint64 addr, node_fn func, void *arg)
       uint16 num_children = trunk_num_children(spl, node);
       for (uint16 pivot_no = 0; pivot_no < num_children; pivot_no++) {
          trunk_pivot_data *pdata = trunk_get_pivot_data(spl, node, pivot_no);
-         bool succeeded_on_subtree = trunk_for_each_subtree(spl, pdata->addr, func, arg);
+         bool              succeeded_on_subtree =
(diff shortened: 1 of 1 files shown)


rtjohnso (approved):
So far I don't see any logic changes.  This seems to just extend the filter_scratch structure to contain enough info to find the node again, and updates logic to use that structure instead of the compact_bundle structure.

Which is all find, but just asking in case I missed something.

ajhconway:
This is in fact a logic change. The way that filter building and replacement works is that the thread
1. acquires the node in which the compaction occurred,
2. collects data required to build the filters,
3. releases the node,
4. builds the filters asynchronously,
5. reacquires the node and installs the new filters,
6. releases the node again and is done.

However, if the node splits somewhere in this process, then the thread may need to extend its work to the node's successor[s]. But, somewhat subtly, there are two ways in which that can happen: the first is that the split can happen before step 1, and the second is that the split can happen in between steps 3 and 5.

If the split happens before step 1, then the thread needs to run the entire process on the node and its successor[s]. This key range containing the nodes on which this needs to be done is stored in the `trunk_compact_bundle_req`.

If the split happens between steps 3 and 5, then the thread has built filters for the node as it was at step 2, and so it needs to install them into the node and its successor[s] using the key range of that node. That key range is stored in `trunk_filter_scratch`.

I feel like the code and or comments should bear this out a little more clearly, but I'm not sure how to do that immediately.

Issue #377, opened 2022-03-19
Clean-up CI-script ci/steps.lib.yml, and support for `debug` target in Makefile after PR #332 merges to /main.


gapisback:
### Describe the bug

This isn't not quite a bug, but a code cleansing to remove some build targets and invocations in CI-scripts to remove support for `make debug` target. 

Under PR #332, that whole debug-build scheme is now being changed to drive off of `BUILD_MODE` env-vars.

We need to cleanse these code-fragments:

- From `ci/steps.lib.yml`:

The setting of these `DEFAULT_` env-vars:

```
121     #@ if sanitize == "asan":
122     DEFAULT_CFLAGS: "-fsanitize=address"
123     DEFAULT_LDFLAGS: "-fsanitize=address"
[...]
127     #@ elif sanitize == "msan":
128     DEFAULT_CFLAGS: "-fsanitize=memory"
129     DEFAULT_LDFLAGS: "-fsanitize=memory"
```

This block:
```
138     #@ if is_debug:
139     - "make $MAKE_HELP debug all run-tests"
140     #@ else:
141     - "make $MAKE_HELP all run-tests"
142     #@ end
```

could just reduce to: `"make $MAKE_HELP all run-tests"`

- From `Makefile`, this block:

```
224 # This is for backwards compatibility with old CI.  Once we update CI,
225 # we can delete this target
226 debug: all
227
```

### Reproduction steps

```bash
1.Nothing to repro; this is code cleanup.
2.
3.
...
```

### Expected behavior

This isn't quite a bug ...

PR #383, opened 2022-03-20
(#377) Purge 'debug' as make target. Fix CI-build scripts.

gapisback:
This commit cleans-up CI-scripts of 'make debug' which is no
longer needed. CI jobs drive 'make' using BUILD_ env-vars to
control release v/s debug builds. Makefile no longer needs to
support 'debug' target. Update the 'help' message to list
supported targets.

----

Updated `make help` looks like follows:

```
Fusion-LocalVM:[43] $ make help
Usage: make [<target>]
Supported targets: clean all libs all-tests run-tests test-results install
Environment variables controlling the build:
[... rest of the output is unchanged ...]
```

NOTE: This change has to go in conjunction with `ci/apply-pipeline.sh` to update the CI-scripts in the CI machinery.
For now, I have applied this manually, just to make sure that this PR's CI-jobs succeed. Once they pass, I will revert the CI-scripts to those in /main. And when review passes, will re-apply then. Manual jugglery.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,8 @@ PLATFORM = linux
 PLATFORM_DIR = platform_$(PLATFORM)
 
 help::
-	@echo 'Usage: make [debug]'
+	@echo 'Usage: make [<target>]'
+	@echo 'Supported targets: clean all libs all-tests run-tests test-results install'
 
 #*************************************************************#
 # SOURCE DIRECTORIES AND FILES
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,7 +7,8 @@ PLATFORM = linux
 PLATFORM_DIR = platform_$(PLATFORM)
 
 help::
-	@echo 'Usage: make [debug]'
+	@echo 'Usage: make [<target>]'
+	@echo 'Supported targets: clean all libs all-tests run-tests test-results install'
```
gapisback, line 11: Hard-coded for now. There are `Makefile` scripts / logic that can generate this programatically, but I didn't want to go that far in this round. (Less imp, may add that in future.)

ci/steps.lib.yml
```diff
@@ -134,11 +130,7 @@ config:
     dir: #@ input_name
     args:
     - "-c"
-    #@ if is_debug:
-    - "make $MAKE_HELP debug all run-tests"
```
gapisback, line 138: We don't need this line or the `debug` target any more, as we now run with this code, called a few lines earlier:
```
116     #@ if is_debug:
117     BUILD_MODE: "debug"
118     #@ end
119
```

rtjohnso (approved):
Ship it

rtjohnso:
FWIW, I think we should feel free to merge small, straightforward patches like this one w/o getting approval.

Issue #389, opened 2022-03-24
Memtables are only half full


ajhconway:
### Describe the bug

When setting the `num_extents` memtable config option, the computation assumes that the extents are all full. However, for most workloads, they will be somewhere between half full and full. Empirically on a random insertion workload, I observed them to be roughly half full.

### Reproduction steps

```bash
Add a `platform_log` in `trunk_memtable_incorporate` that prints a sum of the `req->output_pivot_kv_byte_count` statistic.
```

### Expected behavior

Memtables should have roughly `cfg->memtable_capacity` bytes of data when full, by default this is 24MiB.

### Additional context

I think the easier and maybe best fix is just to double the `num_extents` config option in `mt_cfg`. There will be some workloads where this will cause overfull memtables, such as sequential insertion workloads, but I think this is ok.

rtjohnso:
Closing, as I believe PR #390 fixed it.

PR #390, opened 2022-03-24
set extents_per_memtable to 2 * memtable_capacity

rtjohnso:
Addresses #389.

Another way to address 389 would be to double the default memtable size.  I think that is probably the better way, since it gives users more accurate control over how much cache the memtable can consume.


ajhconway:
It's not a straight line no matter what. There are 4 memtables, so the max memory consumption is `4 * max_num_extents` already.

The other thing is that the max node size is based on `max_num_extents` I think, so we need to correct that to be relative to the logical data size of the memtable, *i.e.* it needs to be divided in half.

Issue #391, opened 2022-03-25
Unit test, splinterdb_quick_test, runs into memory leaks, caught in ASAN-build test runs.


gapisback:
### Describe the bug

In /main, at SHA f5de8cf, `bin/unit/splinterdb_quick_test`, runs into the following failures as seen in debug-mode builds.
Similar messages are seen in release builds, also. Seems like some tests are a missing out some calls to properly close writable-buffer interfaces.

```
$ Fusion-LocalVM:[15] $ build-debug-asan/bin/unit/splinterdb_quick_test

TEST 13/13 splinterdb_quick:test_iterator_custom_comparator [OK]
RESULTS: 13 tests (13 ok, 0 failed, 0 skipped) ran in 1509 ms

=================================================================
==115175==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 14 byte(s) in 1 object(s) allocated from:
    #0 0x7f34e2b77ffe in __interceptor_realloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dffe)
    #1 0x7f34e298d452 in platform_realloc src/platform_linux/platform_inline.h:430
    #2 0x7f34e298d5c1 in writable_buffer_ensure_space src/util.c:24
    #3 0x7f34e298d7d3 in writable_buffer_resize src/util.c:43
    #4 0x7f34e28cbd4c in writable_buffer_copy_slice src/util.h:203
    #5 0x7f34e28cc063 in merge_accumulator_copy_message src/data_internal.c:39
    #6 0x7f34e297a789 in btree_lookup_and_merge src/btree.c:2120
    #7 0x7f34e29c8cef in trunk_memtable_lookup src/trunk.c:3534
    #8 0x7f34e29e1c05 in trunk_lookup src/trunk.c:6349
    #9 0x7f34e298c505 in splinterdb_lookup src/splinterdb.c:786
    #10 0x55de98e4a386 in ctest_splinterdb_quick_test_custom_data_config_run tests/unit/splinterdb_quick_test.c:731
    #11 0x55de98e43446 in ctest_main tests/unit/main.c:305
    #12 0x55de98e42f08 in main tests/unit/main.c:151
    #13 0x7f34e26c90b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)

Direct leak of 7 byte(s) in 1 object(s) allocated from:
[...]

PR #393, opened 2022-03-25
(#391) Fix memory leaks found in ASAN-test runs of quick unit-tests.

gapisback:
Couple of test cases in splinterdb_quick_test.c were missing a call
to splinterdb_lookup_result_deinit() before exiting the test case.
This causes memory leaks, detected on ASAN-test-runs. This commit
fixes those problems.


gapisback:
@rtjohnso - I have a quick-fix for these ASAN-memory leak issues discovered by quick unit-tests. 

This review assignment is mainly fyi. I will put this on auto-merge as the fixes are low-risk.

See my update to issue #391 where I have explained why these ASAN-failures are not being detected on CI-test runs.

Issue #353, opened 2022-03-13
Add --verbose test-parameter option, to print verbose test-progress messages / diagnostics.


gapisback:
In order to better diagnose slow running tests, e.g `splinter_test --perf` tests [ see issue #325 ] we need extended test-execution progress messages.

To not distract regular usages of these tests, these extra instrumentation will be added under a new `--verbose` flag. 

Add that flag under this issue, and introduce one small usage of it to demonstrate that it works right.

PR #281, opened 2022-02-22
(#280) Add test to exercise various existing / new print functions

gapisback:
This commit adds a new test case that exercises various functions to print different types of of pages, BTree innards, and RC-allocator innards.

- Add a function to print the super-block
- Modularize BTree printing function to sub-fns for diff page types
- Modularize Trunk printing function to sub-fns for pivots, bundles etc
- Improve layout of allocator details, showing extent-ID, addr and count
- Rename several print-fns to _sprint*(), taking an output stream handle.
- Minor corrections to some trunk print fns for proper alignment of output

All output from this test case will now be go to the /tmp file where tests redirect stdout.

Issue #272, opened 2022-02-17
Address slow performance of unit/splinter_test test cases.


gapisback:
With the PR #224 , we are seeing very slow performance of unit-tests, due to the addition of newly refactored basic Splinter functionality test into `bin/test/splinter_test` file. 

Although this purports to be a "unit-test", we load close to 13 M records, and do several different kinds of lookups in this test case. This test is more like a "stress" test, and should probably not be included in the unit-tests list.

Here are some elapsed time metrics, when running this in `/main` (before the fix) and with this fix, on a VM-Fusion machine on a Mac. The test is inherently designed to run single-threaded, so any availability of multi-cores is not going to improve the performance

- main : OPT:       All unit tests                :   18 s [  0 h  0 m 18 s ]
- Fix.   : OPT:       All unit tests                :  175 s [  0 h  2 m 55 s ]
- main : DEBUG : All unit tests                :   30 s [  0 h  0 m 30 s ]
- Fix.    : DEBUG: All unit tests                :  426 s [  0 h  7 m  6 s ]

In Fix, the only new test is unit/splinter_test, so all drop in performance is entirely contributed by this one test.

Investigate how to refactor this test into smaller chunks, which can be presumably run in parallel (in CI) to get overall reduced run-times.

PR #394, opened 2022-03-25
(#353) Add --verbose-progress test-config option. Add config_parse_test

gapisback:
This commit adds following testing infrastructure, which will be used in
future work to diagnose slow-running perf-tests:

- Add --verbose-progress test-config option. Introduce one small new print
  msg under this option in splinter_test --perf test case
- Add new config_parse_test to verify that the config parsing logic is
  correctly parsing this boolean option. Discovered latent bugs in
  existing parsing logic, which are now fixed.
- Extend test.sh to invoke new unit-test case with required args.


tests/config.c
```diff
@@ -60,6 +60,7 @@ config_set_defaults(master_config *cfg)
       .use_stats                = FALSE,
       .reclaim_threshold        = UINT64_MAX,
       .verbose_logging_enabled  = FALSE,
+      .verbose_progress         = FALSE,
```
gapisback, line 63: [ Intentionally chose `verbose_progress` rather than `verbose_progress_enabled` as for a boolean variable, I felt, the term `_enabled` is redundant. ]

tests/config.c
```diff
@@ -226,37 +227,47 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
...
          config_has_option("no-log")
          {
-            for (cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
+            for (int cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
                cfg[cfg_idx].use_log = FALSE;
             }
          }
          config_has_option("verbose-logging")
          {
-            cfg[cfg_idx].verbose_logging_enabled = TRUE;
```
gapisback, line 253: Here is the bug in the old code. Loop variable `cfg_idx` is uninitialized.

I had copy-cloned this pattern  a few lines below, for the new `--verbose-progress` parsing.

Fortunately, unit-test discovered this bug ... as it would fail in release build but succeed in debug build.

tests/functional/splinter_test.c
```diff
@@ -950,6 +950,11 @@ splinter_perf_inserts(platform_heap_id             hid,
 
    uint64 start_time = platform_get_timestamp();
 
+   bool verbose_progress = show_verbose_test_progress(test_cfg->test_exec_cfg);
+   if (verbose_progress) {
```
gapisback, line 954: One illustrative usage ... to prove that this new config option is working right.

tests/functional/splinter_test.c
```diff
@@ -2572,10 +2579,16 @@ splinter_test(int argc, char *argv[])
...
+
    test_config *test_cfg = TYPED_ARRAY_MALLOC(hid, test_cfg, num_tables);
    for (uint8 i = 0; i < num_tables; i++) {
       test_config_set_defaults(test, &test_cfg[i]);
+
+      // All tables share the same message-generator.
       test_cfg[i].gen = &gen;
+
+      // And, all tests & tables share the same test-execution parameters.
+      test_cfg[i].test_exec_cfg = &test_exec_cfg;
```
gapisback, line 2591: Minor rearrangement of data structures.

Poke ptr to test execution config struct in `test_cfg`, ... so that we can get at the new `verbose_progress` boolean parameter.

tests/functional/splinter_test.c
```diff
@@ -2593,20 +2606,19 @@ splinter_test(int argc, char *argv[])
    data_config       *data_cfg;
    clockcache_config *cache_cfg =
       TYPED_ARRAY_MALLOC(hid, cache_cfg, num_tables);
-   test_exec_config test_exec_cfg;
-   ZERO_STRUCT(test_exec_cfg);
```
gapisback, line 2597: moved up.

tests/test_common.c
```diff
@@ -124,3 +124,9 @@ test_async_ctxt_get(trunk_handle      *spl,
 
    return ctxt;
 }
+
+inline bool
+show_verbose_test_progress(test_exec_config *test_exec_cfg)
```
gapisback, line 129: Maybe this should move to `.h` as it's defined as `inline` ? Wasn't too sure ... ???

gapisback:
Output from [CI-quick-test job](https://runway-ci.eng.vmware.com/builds/9846459) showing that the new test case is invoked and is passing:

```
est.sh: Config-params parsing test: 0 s [ 0h 0m 0s ]
Running 1 CTests, suite name 'config_parse', test case 'all'.
TEST 1/1 config_parse:test_basic_parsing [OK]
RESULTS: 1 tests (1 ok, 0 failed, 0 skipped) ran in 0 ms
```

Makefile
```diff
@@ -387,6 +390,11 @@ $(BINDIR)/$(UNITDIR)/limitations_test: $(COMMON_TESTOBJ)            \
                                        $(OBJDIR)/$(FUNCTIONAL_TESTSDIR)/test_async.o \
                                        $(LIBDIR)/libsplinterdb.so
 
+$(BINDIR)/$(UNITDIR)/config_parse_test: $(UTIL_SYS)                 \
+                                        $(COMMON_TESTOBJ)           \
+                                        $(OBJDIR)/$(FUNCTIONAL_TESTSDIR)/test_async.o \
+                                        $(LIBDIR)/libsplinterdb.so
+
```
ajhconway, line 397: Rebase artifact?

tests/config.c
```diff
@@ -128,13 +130,12 @@ platform_status
...
 {
    uint64 i;
-   uint8  cfg_idx;
    for (i = 0; i < argc; i++) {
       // Don't be mislead; this is not dead-code. See the config macro expansion
       if (0) {
          config_set_uint64("page-size", cfg, page_size)
          {
-            for (cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
+            for (uint8 cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
```
ajhconway, line 138: Let's be consistent with the types here, even though it doesn't really matter. I generally default to `uint64` in this case, so I would prefer that here.

tests/functional/splinter_test.c
```diff
@@ -2593,20 +2606,19 @@ splinter_test(int argc, char *argv[])
...
 
    rc = test_parse_args_n(splinter_cfg,
                           &data_cfg,
                           &io_cfg,
                           &al_cfg,
                           cache_cfg,
                           &log_cfg,
-                          &test_exec_cfg,
+                          &test_exec_cfg, // Extract test-execution params here
```
ajhconway, line 2616: This is a weird comment.

tests/functional/test.h
```diff
@@ -260,10 +260,12 @@ test_config_init(trunk_config           *splinter_cfg,  // OUT
 /*
  * Some command-line [config] arguments become test execution parameters.
  * Define a structure to hold these when parsing command-line arguments.
+ * This is shared by both functional and unit-test methods.
  */
 typedef struct test_exec_config {
    uint64 seed;
    uint64 num_inserts;
+   bool   verbose_progress; // --verbose-progress: During test execution
```
ajhconway, line 268: I don't understand the distinction between this config and the more general `test_config`.

tests/test_common.c
ajhconway: This should be `test_show_verbose_progress`, since it's exposed and coming from a test header.

tests/unit/splinter_test.c
```diff
@@ -138,11 +138,7 @@ CTEST_SETUP(splinter)
    data->splinter_cfg = TYPED_ARRAY_MALLOC(data->hid, data->splinter_cfg,
                                           num_tables);
 
-   data->data_cfg = TYPED_ARRAY_MALLOC(data->hid, data->data_cfg,
-                                       num_tables);
-
-   data->cache_cfg = TYPED_ARRAY_MALLOC(data->hid, data->cache_cfg,
-                                        num_tables);
+   data->cache_cfg = TYPED_ARRAY_MALLOC(data->hid, data->cache_cfg, num_tables);
```
ajhconway, line 141: What's going on with this `data_cfg` allocation going away?

ajhconway (commented):
I would like some minor changes.

The verbose progress option is nice.

I'm less sold on the config-parsing test. It's a test of test code, and it feels like it's pretty rigid.

Makefile
gapisback: Not sure what you meant here ... I found nothing wrong. Rebase is fine. 

(I just now re-indented the `\` to align on the far right for each line.)

ajhconway: It's duplicated on L390.

tests/config.c
gapisback: Well ... I was just applying the type that was used for this variable's definition on deleted old line L131.
(No change; left it as uint8.)

tests/functional/splinter_test.c
gapisback: (I thought it was self-explanatory.  This parsing routine disburses diff config params to diff locations. The comment is trying to guide the reader that this class of "test-execution" parameters are parsed into this struct. Would like to retain it.)

tests/functional/test.h
gapisback: `test_config` is a test-specific configuration, defined in `functional/splinter_test.h`.  [ I think it's probably mis-named, as it is not for *any* test, but just for this one very specific test. ]

This `test_exec_config` is a struct I had added earlier under SHA a2768f5, to address issue #272. This was done to support `--num-inserts` as a test-execution parameter.

The problem is that the existing `test_config` is only used by splinter_test, where as this new `--num-inserts` is used by unit-test `splinter_test`.

Right now, with this change, I am extending the test-execution config parameter set to include `--verbose-progress`, which currently will be used in the functional test. By adding it to this struct, it leaves open the possibility to do verbose-printing in unit-tests, too, in future.

tests/test_common.c
gapisback: I had it named like that before, but because of this nagging `test_` and then `show_` terminology, I felt the current name read better.

Anyway ... I'll change it as you remarked, but it's not my 1st choice.

tests/unit/splinter_test.c
gapisback: It's actually a funny business. For now, it's a patch-fix, but there is potentially a latent bug waiting to happen, if ever these tests are run with num-tables > 1.

The issue is that if we allocated memory here, we should free it at the end. Otherwise, there is a small memory leak and (I think) ASAN/ MSAN runs will trip up.

But, if you try to call `platform_free()` at the end of the test case on `data->data_cfg`, you will get an error.

That's because we inside this parsing function, we have this code:

```
206 static inline void
207 test_config_init(trunk_config           *splinter_cfg,  // OUT
208                  data_config           **data_cfg,      // OUT
[...]
217    *data_cfg             = test_data_config;
```

Although the code is meaning to say "structure assignment", I think all that is happening here is that `data_cfg` output param is set to the global `test_data_config`'s address.

So, upon return from `test_parse_args_n`, we would have trashed the allocated handle to `data_cfg` ptr.
And an attempt to `platform_free()` at the end will fail with some error; I forget what.

The latent bug is coming from here in `test_parse_args_n`:

```
310    for (i = 0; i < num_config; i++) {
311       test_config_init(&splinter_cfg[i],
312                        &data_cfg[i],
...
```

Here on L310, `num_config` is really supposed to be `num_tables`. 

This area is soft & buggy ... and I did not want to chase all of them down and bloat this change set.

I don't think we are running with num_tables > 1.  The right thing to do would be to exercise such tests, confirm that the behvaiour is buggy and fix the code the right way. But that's ... way out of scope.


??? Makes sense ??

Makefile
gapisback: Hmmm ... not exactly, L390 is for the previous test binary. Here is what I am seeing in my branch:

```
389 $(BINDIR)/$(UNITDIR)/limitations_test: $(COMMON_TESTOBJ)            \
390                                        $(OBJDIR)/$(FUNCTIONAL_TESTSDIR)/test_async.o \
391                                        $(LIBDIR)/libsplinterdb.so
392
393 $(BINDIR)/$(UNITDIR)/config_parse_test: $(UTIL_SYS)                                   \
394                                         $(COMMON_TESTOBJ)                             \
395                                         $(OBJDIR)/$(FUNCTIONAL_TESTSDIR)/test_async.o \
396                                         $(LIBDIR)/libsplinterdb.so
397
```

Both are defined separately. Seems ok to me [?].

tests/config.c
```diff
@@ -155,7 +156,7 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
          }
          config_set_uint64("extent-size", cfg, extent_size)
          {
-            for (cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
+            for (int cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
```
gapisback, line 159: I see what you mean by the above comment.

All these `int cfg_idx` should have been `int8` when I made this bug-fix, to match what was on old line L131.

I changed it accordingly to be `uint8`.

And, oh, btw, this variable `num_config` is misleading; it is really supposed to be `num_tables` (for which we are setting up these config structs). And num-tables is limited elsewhere to 8, and other code is using `uint8` for that variable's type.

gapisback:
@ajhconway - Made few minor corrections. Let me know if you are ok, or still would like to see some edits.

About the `config_parse_test` being rigid ... well, first of all, that rigidity is "hidden" in `test.sh`, and it's not included in `bin/unit_test`. So no one will accidentally stumble upon it.

Also, I had to write that test case as there were diffs in release v/s debug outputs, and the test case caught the bug well. So, there is some value.

We are also continually talking about perhaps over-hauling this config-parsing machinery one day. I think when that happens, this test case will be a useful dumping ground to start exercising correctness tests for that config-parsing interface.

I'd like to retain it for the reasons given above.

Makefile
ajhconway: This code has changed since I made that comment.

tests/functional/splinter_test.c
ajhconway: It's not clear to me why the reader especially cares about the `test_exec_cfg` vs any of the other configs. The format draws a lot of attention to it and it doesn't seem that important to me.

I feel as though a comment above the function saying that this parses the command line into the various configs would be more useful, since then it's obvious from context.

ajhconway (commented):
Something relatively minor that's holding me up is that I still don't understand the distinction between a `test_config` and a `test_exec_config`? Why do we need both?

tests/unit/splinter_test.c
rtjohnso: I believe the data_cfg thing is a change I made.

I modified the data_config system so that you can sub-class a data_config in the same way that we sub-class other structs:

struct my_data_config {
  data_config super;
  <additional fields go here>;
};

This means that we must never copy data_configs.  We just take the pointer given to us by the application and use that.  That also means we don't allocate memory for data_configs.

tests/functional/splinter_test.c
gapisback: I took it out. Adding an extra comment just at this invocation may require adding a similar comment in all other places where it's called.  The called function's prolog (I see) already documents that the configs are distributed to various other config sub-structs.

gapisback:
@ajhconway - let's talk tomorrow in the office about the need for `test_exec_cfg` v/s `test_config`. I may be able to explain it better if we walk-through code.

tests/unit/splinter_test.c
gapisback: But, then @rtjohnso - Do you agree with my suspicion that ... "but there is potentially a latent bug waiting to happen, if ever these tests are run with num-tables > 1."?

We can walk-thru code tomorrow when in the office.

I think what I have now in this change-set is adequately correct for the set of tests being run now.

Do you think there is any further need to revise the code as submitted for review?

gapisback:
@ajhconway - We discussed the issues around `test_exec_config` and why it exists, somewhat duplicated by `test_config`.
Agreed that teasing these apart will be larger scope, and we should re-view all those when all of test-config parsing is overhauled.

Brought-in the rename request (from PR #281 ) to this one: `PRInd()` -> `DECIMAL_STRING_WIDTH`

Issue #396, opened 2022-03-29
Consistently change all close(), deinit(), dismount() interfaces to NULL-out the input handle on exit. 


gapisback:
### Describe the bug

This isn't particularly an immediate bug, but something that can go wrong when code accesses a handle to a structure that has been closed, deinit'ed, or its memory has been freed. 

To avoid such issues, revise the interface to take an address-to-the-ptr-to struct, which gets NULL'ed out on return.

Some examples of interfaces that will be affected by this work are seen in this code-flow:

```
554 void
555 splinterdb_close(splinterdb *kvs) // IN
556 {
557    platform_assert(kvs != NULL);
558
559    trunk_unmount(&kvs->spl);
560    clockcache_deinit(&kvs->cache_handle);
561    rc_allocator_dismount(&kvs->allocator_handle);
562    io_handle_deinit(&kvs->io_handle);
563    task_system_destroy(kvs->heap_id, kvs->task_sys);
564
565    platform_free(kvs->heap_id, kvs);
566 }
```

In this example, the input `splinterdb **kvs_in` is not being NULL'ed out on exit, but there are other places where such interfaces are being called, especially in test-code. An example from `ycsb_test.c`:

```
1339    trunk_dismount(spl);
1340    clockcache_deinit(cc);
1341    platform_free(hid, cc);
1342    rc_allocator_dismount(&al);
1343    test_deinit_task_system(hid, ts);
1344    rc = STATUS_OK;
```

It will be safer to NULL out the handles upon a successful exit.

### Reproduction steps

```bash
1. (None; see description above.)
```

### Expected behavior

Safe-code, avoids accessing a stale handle.

gapisback:
FYI, @ajhconway @rtjohnso -- While reworking these interfaces, I propose to also make the following renames:
(I have aired this `dismount` -> `unmount` name change earlier, and it seemed to have passed muster.)

- `trunk_dismount()` -> `trunk_unmount()`
- `rc_allocator_dismount()` -> `rc_allocator_unmount()`
- And to change code such as in `trunk.c` : `trunk.c:   super->dismounted   = is_dismount` to always use `unmount`

Before bringing this into scope of this cleanup issue, wanted to run this by you, in case you object. 

I think it's ok to roll this into this issue as the call-sites are anyway being touched.

PR #397, opened 2022-03-29
(#396) NULL-out input handle(s) after close(). Rename dismount -> unmount(), where applicable.

gapisback:
Minor rework to tighten the interfaces which close / deinit sub-system handles:

- Rename "dismount" to "unmount"; as in 'trunk_unmount()'
- splinterdb_close(), trunk_unmount(), test_deinit_task_system(),
  task_system_destroy() now receive an addr-of-ptr to handle, and NULL-out
  the handle upon a successful close / unmount / deinit. This will cause
  a seg-fault if caller tries to access said handle after it has been closed.

-----

While developing C-example programs, I ran into some bad code (my own) while calling `splinterdb_close()`. 
I corrected the interface in my private branch, and then recollected that I had proposed to do this cleanup some weeks ago.
Peeling off this cleanup item as its own PR so that this rework can be done independently. 

It will help write out the C-example programs with a neat code-construct, calling `splinterdb_close(&splinter_handle)`, so as to inculcate proper form in application programs.

### DIFF
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -99,7 +99,7 @@ splinterdb_open(const splinterdb_config *cfg, splinterdb **kvs);
 //
 // This will flush all data to disk and release all resources
 void
-splinterdb_close(splinterdb *kvs);
+splinterdb_close(splinterdb **kvs);
 
 // Register the current thread so that it can be used with splinterdb.
 // This causes scratch space to be allocated for the thread.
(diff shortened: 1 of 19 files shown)


gapisback:
NOTE: The proposal to rename `trunk_dismount()` to `trunk_unmount()` and such like has been aired out in [this internal slack thread](https://vmware.slack.com/archives/G01MEDQ05MX/p1644886329553479?thread_ts=1644886105.189249&cid=G01MEDQ05MX). And has been accepted.

ajhconway (approved):
Ship it

Issue #354, opened 2022-03-13
Instrument perf test cases in splinter_test to print extra diagnostics, elapsed-time monitoring. (Needed to resolve slow-execution time of this perf test.)


gapisback:
To address slow running `splinter_test --perf` tests [ reported under #325 ], these test case methods have been refactored into sub-functions under issue #334. A new `--verbose` flag will be added under issue #353. 

In nightly CI-jobs, these tests never seem to complete for well over few hours. Even on Nimbus-VM runs, the `--perf` test is seen to take over 2 hours. (One example is logged under #314.) 

Here's a data point (dd. Sun 13.3.2022) from a most recent run off of /main @ SHA 8f192c1, run on machine `sdb-oss-test-vm`, with this h/w specs:

```
sdb-oss-test-vm:[27] $ osinfo.sh
Linux sdb-oss-test-vm 5.15.0-18-generic #18-Ubuntu SMP Fri Jan 21 14:57:54 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
 GenuineIntel, 32 CPUs, 62 GB,  Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz
```
The nightly run tests run-time metrics are as follows:

```
Sun Mar 13 12:53:04 UTC 2022 **** SplinterDBNightlyTest Suite Execution Times ****

Functionality Stress test 10 mil rows, 1 tables, 4 GiB cache                    :   67s [  0h  1m  7s ]
Functionality Stress test 10 mil rows, 2 tables, 4 GiB cache                    :  137s [  0h  2m 17s ]
Functionality Stress test 1 mil rows, 2 tables, 1 MiB cache                     :   10s [  0h  0m 10s ]
Functionality Stress test 1 mil rows, 4 tables, 1 MiB cache                     :   24s [  0h  0m 24s ]
Performance (sync) test 8 insert, 8 lookup, 8 range lookup threads              : 7257s [  2h  0m 57s ]
Parallel Performance (sync) test 10 pthreads                                    : 1215s [  0h 20m 15s ]
Cache Performance test , default cache size                                     :   44s [  0h  0m 44s ]
Cache Performance test , 6 GiB cache                                            :  365s [  0h  6m  5s ]
Nightly Stress & Performance Tests                                              : 9151s [  2h 32m 31s ]
```

**The single long pole test** is : Performance (sync) test 8 insert, 8 lookup, 8 range lookup threads, taking ~ 3hrs.

```
[...]

Issue #325, opened 2022-03-05
Nightly build and test job, job_nightly_main_build_test, timing out after 6 hours, due to slow-running splinter_test --perf run.


gapisback:
Recently activated main-nightly-build-test CI-jobs are timing out closer to 3h, rather than the expected 6 hours.

See [this run #6](https://runway-ci.eng.vmware.com/teams/splinterdb/pipelines/splinterdb/jobs/main-nightly-build-test-gcc/builds/6) and [run 6.1,](https://runway-ci.eng.vmware.com/teams/splinterdb/pipelines/splinterdb/jobs/main-nightly-build-test-gcc/builds/6.1) where both timed-out before the tests could complete execution.

- Run # 6 started shortly after midnight, and timed-out ~ 1:08:27, so slightly over an hour
- Run #6.1 started 6:46 a.m and timed-out ~8:01 - that's also about 75-80 mins

It almost looks like in the block below, we are returning at the end on L13, rather than the expected L10 (6h)

```
  5 #@ def get_task_timeout(quick=False, sanitize="", is_nightly=False):
  6 #@   if sanitize:
  7 #@     return "3h"
  8 #@   elif quick:
  9 #@     return "5m"
 10 #@   elif is_nightly:
 11 #@     return "6h"
 12 #@   else:
 13 #@     return "1h"
 14 #@   end
 15 #@ end
```

PR #398, opened 2022-03-31
(#354) splinter_test --perf: Fix bug in range-lookups test code.

gapisback:
Earlier refactoring of splinter_test.c [under SHA 3b7bc58] caused a
bug which is probably the cause for excessively long run-times for
this test case. (The # of range lookups done were inordinately high.)
This commit fixes that bug, and additionally:

- Adds extra progress messages under --verbose-progess flag.
- Fixes minor (divide-by-zero) bugs when this test is run with very
  small # of --num-inserts rows.
- Enables splinter_test --perf in nightly runs with --verbose-progress
  flag to run w/large data set
- Includes a quick perf run call to exercise new --verbose-progress
  and --num-inserts options.

----

NOTE : This started a change set to add --verbose` progress instrumentation to perf tests. And while triaging the slow perf-test runs, discovered (Thanks to Alex!) a bug that had crept in as part of an earlier refactoring.

That bug is also being fixed in this PR, and should address issue #325

----

Tested the re-enablement of `splinter_test --perf` done in `test.sh` (with `--tree-size-gib 4`, resulting in something like ~54 million rows) completed on a manual execution on CI-machines in about few mins. (Less than 5 m, I think.)

Am hopeful that re-enabling this test case in nightly runs will succeed now.

Am running a manual verification of this test case on CI-machines with default workload size; i.e. `--tree-size-gib 40`, which results in ~543 million rows. Depending on how long that takes, I will re-visit running this case with the default test parameter.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -233,27 +233,26 @@ function nightly_sync_perf_tests() {
     local dbname="splinter_test.perf.db"
 
     # Different #s of threads. --perf test runs in phases, where some # of
-    # threads are setup. Insert / lookup / range-lookups are run. Then, the
-    # threads are destroyed.
+    # threads are setup. Insert / lookup / range-lookups are run in separate
+    # phases. Then, the threads are destroyed.
     local nins_t=8
     local nlookup_t=8
     local nrange_lookup_t=8
     local test_descr="${nins_t} insert, ${nlookup_t} lookup, ${nrange_lookup_t} range lookup threads"
 
     # ----
-    # (#325) Commented this out till slow-performance of this test is resolved.
-    # splinter_test --perf is included in CI-runs, but just with
-    # --num-range-lookup-threads 0
-    # run_with_timing "Performance (sync) test ${test_descr}" \
-    #       "$BINDIR"/driver_test splinter_test --perf \
-    #                                           --max-async-inflight 0 \
-    #                                           --num-insert-threads ${nins_t} \
-    #                                           --num-lookup-threads ${nlookup_t} \
-    #                                           --num-range-lookup-threads ${nrange_lookup_t} \
-    #                                           --lookup-positive-percent 10 \
-    #                                           --db-capacity-gib 60 \
-    #                                           --db-location ${dbname}
-    # rm ${dbname}
+    run_with_timing "Performance (sync) test ${test_descr}" \
+            "$BINDIR"/driver_test splinter_test --perf \
+                                                --max-async-inflight 0 \
+                                                --num-insert-threads ${nins_t} \
+                                                --num-lookup-threads ${nlookup_t} \
+                                                --num-range-lookup-threads ${nrange_lookup_t} \
+                                                --lookup-positive-percent 10 \
+                                                --tree-size-gib 4 \
+                                                --db-capacity-gib 60 \
+                                                --db-location ${dbname} \
+                                                --verbose-progress
+    rm ${dbname}
 
     local npthreads=8
     local tree_size=8 # GiB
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -233,27 +233,26 @@ function nightly_sync_perf_tests() {
...
-    #       "$BINDIR"/driver_test splinter_test --perf \
-    #                                           --max-async-inflight 0 \
-    #                                           --num-insert-threads ${nins_t} \
-    #                                           --num-lookup-threads ${nlookup_t} \
-    #                                           --num-range-lookup-threads ${nrange_lookup_t} \
-    #                                           --lookup-positive-percent 10 \
-    #                                           --db-capacity-gib 60 \
-    #                                           --db-location ${dbname}
-    # rm ${dbname}
+    run_with_timing "Performance (sync) test ${test_descr}" \
```
gapisback, line 244: Re-enabling test case in nightly runs.

tests/functional/splinter_test.c
```diff
@@ -977,8 +1016,9 @@ splinter_perf_inserts(platform_heap_id             hid,
    timestamp insert_latency_max = 0;
    uint64    read_io_bytes, write_io_bytes;
    cache_io_stats(cc[0], &read_io_bytes, &write_io_bytes);
-   uint64 io_mib    = (read_io_bytes + write_io_bytes) / MiB;
-   uint64 bandwidth = io_mib / NSEC_TO_SEC(total_time);
+   uint64 io_mib = (read_io_bytes + write_io_bytes) / MiB;
+   uint64 bandwidth =
+      (NSEC_TO_SEC(total_time) ? (io_mib / NSEC_TO_SEC(total_time)) : 0);
```
gapisback, line 1021: Changes like this one fix divide-by-zero errors that occur if you run with very small `--num-inserts 10000` arg.

Now, the entire test can run with small # of rows, which greatly helps with re-validating test-code changes.

tests/functional/splinter_test.c
```diff
@@ -993,12 +1033,13 @@ splinter_perf_inserts(platform_heap_id             hid,
...
-         "\nper-splinter per-thread insert time per tuple %lu ns\n",
-         total_time * num_insert_threads / *total_inserts);
+      platform_default_log("\nTotal time=%lus, per-splinter per-thread "
+                           "insert time per tuple %lu ns\n",
+                           NSEC_TO_SEC(total_time),
+                           (total_time * num_insert_threads) / *total_inserts);
       platform_default_log(
          "splinter total insertion rate: %lu insertions/second\n",
-         SEC_TO_NSEC(*total_inserts) / total_time);
+         (total_time ? SEC_TO_NSEC(*total_inserts) / total_time : 0));
```
gapisback, line 1042: Divide-by-zero bugfix.

tests/functional/splinter_test.c
```diff
@@ -1304,14 +1401,15 @@ test_splinter_perf(trunk_config    *cfg,
       trunk_range_perf_params perf_ranges[] = {
          //               number     min                 max
          {"Small range"  , 128      , 1                , 100    },
-         {"Medium range" , 4        , 512              , 1024   },
-         {"Large range"  , 4        , (131072 - 16384) , 131072 }
+         {"Medium range" , 512      , 512              , 1024   },
+         {"Large range"  , 2048     , (131072 - 16384) , 131072 }
```
gapisback, line 1405: These two lines are the real ones fixing the slow-perf bug.

ajhconway (approved):
Looks good

PR #399, opened 2022-04-02
Set of example C-programs, to demonstrate use of SplinterDB APIs.

gapisback:
This commit adds a small number of C-sample programs, to demonstrate use of SplinterDB core APIs.
Associated changes to build this set of stand-alone programs applied to `Makefile`. A small new "test-case" is added to
`test.sh` to always execute these programs as part of default testing set.
Associated docs files are also updated to direct the user to these programs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ INCDIR               = include
 FUNCTIONAL_TESTSDIR  = $(TESTS_DIR)/functional
 UNITDIR              = unit
 UNIT_TESTSDIR        = $(TESTS_DIR)/$(UNITDIR)
+EXAMPLES_DIR         = examples
 
 SRC := $(shell find $(SRCDIR) -name "*.c")
 
@@ -39,6 +40,8 @@ TESTSRC := $(COMMON_TESTSRC) $(FUNCTIONAL_TESTSRC) $(UNIT_TESTSRC)
 #  - Skip tests that are to be invoked with specialized command-line arguments.
 # These skipped tests which will have to be run stand-alone.
 FAST_UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c" | egrep -v -e"splinter_test|config_parse_test")
+
+EXAMPLES_SRC := $(shell find $(EXAMPLES_DIR) -name "*.c")
 
 #*************************************************************#
 # CFLAGS, LDFLAGS, ETC
@@ -226,12 +229,18 @@ FAST_UNIT_TESTOBJS= $(FAST_UNIT_TESTSRC:%.c=$(OBJDIR)/%.o)
 UNIT_TESTBIN_SRC=$(filter %_test.c, $(UNIT_TESTSRC))
 UNIT_TESTBINS=$(UNIT_TESTBIN_SRC:$(TESTS_DIR)/%_test.c=$(BINDIR)/%_test)
 
+# ---- Symbols to build example sample programs
+EXAMPLES_OBJ= $(EXAMPLES_SRC:%.c=$(OBJDIR)/%.o)
+EXAMPLES_BIN_SRC=$(filter %_example.c, $(EXAMPLES_SRC))
+EXAMPLES_BINS=$(EXAMPLES_BIN_SRC:$(EXAMPLES_DIR)/%_example.c=$(BINDIR)/$(EXAMPLES_DIR)/%_example)
+
 ####################################################################
 # The main targets
 #
-all: libs all-tests $(EXTRA_TARGETS)
+all: libs all-tests all-examples $(EXTRA_TARGETS)
 libs: $(LIBDIR)/libsplinterdb.so $(LIBDIR)/libsplinterdb.a
 all-tests: $(BINDIR)/driver_test $(BINDIR)/unit_test $(UNIT_TESTBINS)
+all-examples: $(EXAMPLES_BINS)
 
 #######################################################################
 # CONFIGURATION CHECKING
@@ -331,6 +340,7 @@ $(BINDIR)/unit_test: $(FAST_UNIT_TESTOBJS) $(COMMON_TESTOBJ) $(LIBDIR)/libsplint
 # all the mini unit tests depend on these files
 $(UNIT_TESTBINS): $(OBJDIR)/$(UNIT_TESTSDIR)/main.o
 
+# -----------------------------------------------------------------------------
 # Every unit test of the form bin/unit/<test> depends on obj/tests/unit/<test>.o
 #
 # We can't use pattern rules to state this dependency pattern because
@@ -414,7 +424,7 @@ $(BINDIR)/$(UNITDIR)/config_parse_test: $(UTIL_SYS)                             
                                         $(LIBDIR)/libsplinterdb.so
 
 ########################################
-# Convenience targets
+# Convenience mini unit-test targets
 unit/util_test:                    $(BINDIR)/$(UNITDIR)/util_test
 unit/misc_test:                    $(BINDIR)/$(UNITDIR)/misc_test
 unit/btree_test:                   $(BINDIR)/$(UNITDIR)/btree_test
@@ -424,6 +434,25 @@ unit/splinterdb_quick_test:        $(BINDIR)/$(UNITDIR)/splinterdb_quick_test
 unit/splinterdb_stress_test:       $(BINDIR)/$(UNITDIR)/splinterdb_stress_test
 unit/writable_buffer_test:         $(BINDIR)/$(UNITDIR)/writable_buffer_test
 unit_test:                         $(BINDIR)/unit_test
+
+# -----------------------------------------------------------------------------
+# Every example program of the form bin/examples/<eg-prog> depends on
+# obj/examples/<eg-prog>.o
+
+#################################################################
+# The dependencies of each example program
+
+$(BINDIR)/$(EXAMPLES_DIR)/splinterdb_intro_example: $(OBJDIR)/$(EXAMPLES_DIR)/splinterdb_intro_example.o \
+                                                    $(LIBDIR)/libsplinterdb.so
+
+$(BINDIR)/$(EXAMPLES_DIR)/splinterdb_wide_values_example: $(OBJDIR)/$(EXAMPLES_DIR)/splinterdb_wide_values_example.o \
+                                                   $(LIBDIR)/libsplinterdb.so
+
+$(BINDIR)/$(EXAMPLES_DIR)/splinterdb_iterators_example: $(OBJDIR)/$(EXAMPLES_DIR)/splinterdb_iterators_example.o \
+                                                        $(LIBDIR)/libsplinterdb.so
+
+$(BINDIR)/$(EXAMPLES_DIR)/splinterdb_custom_ipv4_addr_sortcmp_example: $(OBJDIR)/$(EXAMPLES_DIR)/splinterdb_custom_ipv4_addr_sortcmp_example.o \
+                                                                       $(LIBDIR)/libsplinterdb.so
 
 #*************************************************************#
 
--- a/examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
+++ b/examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
@@ -1,0 +1,390 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * ----------------------------------------------------------------------------
+ * SplinterDB Advanced Iterators Example Program with custom sort-comparison.
+ *
+ * What's new beyond previous splinterdb_iterators_example.c?
+ *
+ * In this program, we show the application of user-specified custom
+ * key-comparison routines. The 'key' here is the 4-part IP-address, which
+ * is stored as the string seen from 'ping'; i.e. "208.80.154.232" .
+ * To illustrate the use of user-defined keys, we then provide a
+ * sort-comparison routine, which splits up the IP-address to its
+ * constituent parts, and does a numeric comparison of each 1-byte value.
+ *
+ * See:
+ * - The definition of custom splinter_data_cfg.key_compare to the
+ *   user-provided comparison function, custom_key_compare()
+ * - The ip4_ipaddr_keycmp() and ip4_split() functions that show how one
+ *   can deal with application-specific key formats.
+ * ----------------------------------------------------------------------------
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stddef.h>
+#include <string.h>
+
+#include "splinterdb/default_data_config.h"
+#include "splinterdb/splinterdb.h"
+
+/*
+ * App-specific 'defaults' that can be parameterized, eventually.
+ */
+#define DB_FILE_NAME    "splinterdb_custom_ipv4_sortcmp_example_db"
+#define DB_FILE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
+#define CACHE_SIZE_MB   64   // Size of cache; can be changed across boots
+
+// Describe the layout of fields in an IP4-address
+#define IPV4_NUM_FIELDS 4
+#define IPV4_NUM_DOTS   (IPV4_NUM_FIELDS - 1)
+
+#define IP4_MIN_KEY_SIZE ((1 * IPV4_NUM_FIELDS) + IPV4_NUM_DOTS)
+
+// Application declares the limit of key-sizes it intends to use
+#define IP4_MAX_KEY_SIZE ((3 * IPV4_NUM_FIELDS) + IPV4_NUM_DOTS)
+
+// Max # of chars in a well-formed IP4 address, including null-terminator byte
+#define IPV4_MAX_KEY_BUF_SIZE (IP4_MAX_KEY_SIZE + 1)
+
+// Key is a 4-part inet-address string, whose value is a description of
+// the IP-address: its www-name, and some ping metrics.
+typedef struct www_ping_metrics {
+   uint32     ttl_ms; // Time-to-live
+   uint32     rtt_ms; // Round trip time
+   const char www_name[30];
+} www_ping_metrics;
+
+#define WWW_PING_SIZE(p)                                                       \
+   (size_t)(offsetof(www_ping_metrics, www_name) + strlen((p)->www_name))
+
+typedef struct kv_pair {
+   char            *kv_key;
+   www_ping_metrics kv_val;
+} kv_pair;
+
+// clang-format off
+// Define a hard-coded array of key-value pairs to load
+// Mapping from inet-IP address to www.address etc.
+kv_pair inet_addr_info[] =
+{
+    //   ip-address             ttl  rtt   www-address
+      { "5.79.89.114"       , {  47, 171, "www.acm.org" } }
+    , { "208.80.154.232"    , {  52,  99, "www.wikidpedia.org" } }
+    , { "151.101.188.81"    , {  57,  28, "www.bbc.com" } }
+    , { "99.84.238.130"     , { 240,  46, "www.worldbank.org" } }
+    , { "10.113.78.20"      , {  57,  32, "www.vmware.com" } }
+    , { "34.102.136.180"    , {  116, 33, "www.eiffeltower.com" } }
+    , { "184.26.53.176"     , {  56,  33, "www.rediff.com" } }
+    , { "151.101.190.154"   , {  58,  37, "www.cnet.com" } }
+    , { "104.244.42.129"    , {  52,  74, "www.twitter.com" } }
+    , { "104.143.9.110"     , {  49,  91, "www.hongkongair.com" } }
+};
+
+int num_inet_addrs = (sizeof(inet_addr_info) / sizeof(*inet_addr_info));
+
+// clang-format on
+
+// Define min/max key values for user-defined IP4-addr keys
+#define IP4ADDR_MIN_KEY "0.0.0.0"
+#define IP4ADDR_MAX_KEY "255.255.255.255"
+
+// Function Prototypes
+static void
+configure_splinter_instance(splinterdb_config *splinterdb_cfg,
+                            data_config       *splinter_data_cfg,
+                            const char        *filename,
+                            uint64             dev_size,
+                            uint64             cache_size);
+
+int
+custom_key_compare(const data_config *cfg, slice key1, slice key2);
+
+int
+ip4_ipaddr_keycmp(const char  *key1,
+                  const size_t key1_len,
+                  const char  *key2,
+                  const size_t key2_len);
+
+int
+ip4_split(int *key_fields, const char *key, const size_t key_len);
+
+static void
+do_inserts(splinterdb *spl_handle, kv_pair *kv_pairs, int num_kv_pairs);
+
+static int
+do_iterate_from(splinterdb *spl_handle, const char *from_key);
+
+static void
+print_ping_metrics(int kctr, slice key, slice value);
+
+/*
+ * -----------------------------------------------------------------------------
+ * main() Driver for SplinterDB program for iterators interfaces.
+ * -----------------------------------------------------------------------------
+ */
+int
+main()
+{
+   printf("     **** SplinterDB Iterators example program: "
+          "Custom sort comparison  ****\n\n");
+
+   // Initialize data configuration, describing your key-value properties
+   data_config splinter_data_cfg;
+   default_data_config_init(IP4_MAX_KEY_SIZE, &splinter_data_cfg);
+
+   // -- ACTION IS HERE --
+   // Customize key-comparison with our implementation for IP4 addresses
+   // **** NOTE **** Custom key-comparision function needs to be provided
+   // up-front. Every insert will invoke this method to insert the new key
+   // in custom-sorted order.
+   sprintf(splinter_data_cfg.min_key, "%s", IP4ADDR_MIN_KEY);
+   sprintf(splinter_data_cfg.max_key, "%s", IP4ADDR_MAX_KEY);
+
+   splinter_data_cfg.min_key_length = strlen(splinter_data_cfg.min_key);
+   splinter_data_cfg.max_key_length = strlen(splinter_data_cfg.max_key);
+   splinter_data_cfg.key_compare    = custom_key_compare;
+
+   // Basic configuration of a SplinterDB instance
+   splinterdb_config splinterdb_cfg;
+   configure_splinter_instance(&splinterdb_cfg,
+                               &splinter_data_cfg,
+                               DB_FILE_NAME,
+                               (DB_FILE_SIZE_MB * 1024 * 1024),
+                               (CACHE_SIZE_MB * 1024 * 1024));
+
+   splinterdb *spl_handle = NULL; // To a running SplinterDB instance
+
+   int rc = splinterdb_create(&splinterdb_cfg, &spl_handle);
+   if (rc) {
+      printf("SplinterDB creation failed. (rc=%d)\n", rc);
+      return rc;
+   }
+
+   printf("Insert and iterate through %d {ipaddr} -> {www-url, metrics} "
+          "mapping table:\n",
+          num_inet_addrs);
+   do_inserts(spl_handle, inet_addr_info, num_inet_addrs);
+
+   // NULL start_key => Iterate through all key-value pairs
+   do_iterate_from(spl_handle, NULL);
+
+   const char *start_key = "99.84.238.130";
+   do_iterate_from(spl_handle, start_key);
+
+   // Specify a non-existent start key; Scan should return from 104.143.9.110
+   start_key = "100.101.102.103";
+   do_iterate_from(spl_handle, start_key);
+
+   splinterdb_close(&spl_handle);
+   printf("Shutdown SplinterDB instance, dbname '%s'.\n\n", DB_FILE_NAME);
+
+   return rc;
+}
+
+/*
+ * -----------------------------------------------------------------------------
+ * configure_splinter_instance()
+ *
+ * Basic configuration of a SplinterDB instance, specifying min parameters such
+ * as the device's name, device and cache sizes.
+ * -----------------------------------------------------------------------------
+ */
+static void
+configure_splinter_instance(splinterdb_config *splinterdb_cfg,
+                            data_config       *splinter_data_cfg,
+                            const char        *filename,
+                            uint64             dev_size, // in bytes
+                            uint64             cache_size)           // in bytes
+{
+   memset(splinterdb_cfg, 0, sizeof(*splinterdb_cfg));
+   splinterdb_cfg->filename   = filename;
+   splinterdb_cfg->disk_size  = dev_size;
+   splinterdb_cfg->cache_size = cache_size;
+   splinterdb_cfg->data_cfg   = splinter_data_cfg;
+   return;
+}
+
+/*
+ * -----------------------------------------------------------------------------
+ * custom_key_compare() - Implement custom key-comparison function
+ * -----------------------------------------------------------------------------
+ */
+int
+custom_key_compare(const data_config *cfg, slice key1, slice key2)
+{
+   return ip4_ipaddr_keycmp((const char *)slice_data(key1),
+                            slice_length(key1),
+                            (const char *)slice_data(key2),
+                            slice_length(key2));
+}
+
+/*
+ * -----------------------------------------------------------------------------
+ * ipaddr_keycmp() - Custom IPV4 IP-address key-comparison routine.
+ *
+ * -- ACTION IS HERE --
+ * 'key1' and 'key2' are expected to be well-formed IP4 addresses.
+ * - Extract each of the 4 parts of the IP-address
+ * - Implement comparison by numerical sort-order of each part.
+ *
+ * Returns: See below:
+ * -----------------------------------------------------------------------------
+ */
+/* Return value expected from key comparison routine */
+#define KEYCMP_RV_KEY1_LT_KEY2 ((int)-1)
+#define KEYCMP_RV_KEY1_EQ_KEY2 ((int)0)
+#define KEYCMP_RV_KEY1_GT_KEY2 ((int)1)
+
+int
+ip4_ipaddr_keycmp(const char  *key1,
+                  const size_t key1_len,
+                  const char  *key2,
+                  const size_t key2_len)
+{
+   int key1_fields[IPV4_NUM_FIELDS];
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -226,12 +229,18 @@ FAST_UNIT_TESTOBJS= $(FAST_UNIT_TESTSRC:%.c=$(OBJDIR)/%.o)
 UNIT_TESTBIN_SRC=$(filter %_test.c, $(UNIT_TESTSRC))
 UNIT_TESTBINS=$(UNIT_TESTBIN_SRC:$(TESTS_DIR)/%_test.c=$(BINDIR)/%_test)
 
+# ---- Symbols to build example sample programs
+EXAMPLES_OBJ= $(EXAMPLES_SRC:%.c=$(OBJDIR)/%.o)
+EXAMPLES_BIN_SRC=$(filter %_example.c, $(EXAMPLES_SRC))
+EXAMPLES_BINS=$(EXAMPLES_BIN_SRC:$(EXAMPLES_DIR)/%_example.c=$(BINDIR)/$(EXAMPLES_DIR)/%_example)
```
gapisback, line 235: Above diffs is a collection of changes to `Makefile` to compile and build all the example programs.

examples/example_common.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * SplinterDB - Common header file for example programs.
+ */
```
gapisback, line 6: Some of the stuff in this file is pillaged from existing splinter header files. We need this so that the example programs do not have to depend on splinterdb `.h` files.

examples/splinterdb_admin_config_example.c
```diff
@@ -0,0 +1,135 @@
...
+   splinterdb_config splinterdb_cfg;
+   configure_splinter_instance(&splinterdb_cfg,
+                               &splinter_data_cfg,
+                               APP_DB_NAME,
+                               (APP_DEVICE_SIZE_MB * K_MiB),
+                               (APP_CACHE_SIZE_MB * K_MiB));
+
+   splinterdb *spl_handle = NULL; // To a running SplinterDB instance
+
+   // -- ACTION IS HERE --
```
gapisback, line 65: Each program "builds on top" of the previous one, in some ways.

I am using this `-- ACTION IS HERE --` comment tag to direct the reader to parts of the program that is new and different than its previous on the list of programs in `README.md`.

(I don't know of any other way to high-light the new-ness in each program ... Maybe **BOLD ***** ...** but I thought that may be too too distracting.)

include/splinterdb/splinterdb.h
```diff
@@ -97,9 +97,14 @@ splinterdb_open(const splinterdb_config *cfg, splinterdb **kvs);
 
 // Close a splinterdb
 //
-// This will flush all data to disk and release all resources
+// This will flush all data to disk and release all resources.
+// Handle is NULL'ed out on a succcessful close.
 void
-splinterdb_close(splinterdb *kvs);
+splinterdb_close(splinterdb **kvs);
```
gapisback, line 103: This will be in /main once PR #396 passes review and merged to /main. (Will pull this out then.)

include/splinterdb/splinterdb.h
```diff
@@ -97,9 +97,14 @@ splinterdb_open(const splinterdb_config *cfg, splinterdb **kvs);
...
+
+// Check status of a splinterdb instance: Is it running?
+bool
+splinterdb_is_up(const splinterdb *kvs);
```
gapisback, line 107: This is a convenience testing and probe-hook for applications to call. I am bringing this in as part of the examples, but we should probably add this on its commit (for better tracking).

src/rc_allocator.c
```diff
@@ -236,6 +236,39 @@ rc_allocator_config_init(rc_allocator_config *allocator_cfg,
...
+   if (cfg->extent_capacity == 0) {
+      platform_error_log("Configured extent capacity %lu bytes is invalid.\n",
+                         cfg->extent_capacity);
+      return STATUS_BAD_PARAM;
+   }
+   if (cfg->capacity != (cfg->io_cfg->extent_size * cfg->extent_capacity)) {
+      return STATUS_BAD_PARAM;
+   }
+   return rc;
+}
```
gapisback, line 269: This was an error handling issue I ran into while building the sample programs, where some input config-params had the wrong units.

I think these are "small" enough that we can roll these in as part of this examples-PR.

If there is a strong desire to check these in separately, let me know. I can peel these off to a separate commit and integrate them first.

tests/unit/splinterdb_quick_test.c
```diff
@@ -765,7 +765,7 @@ CTEST2(splinterdb_quick, test_iterator_custom_comparator)
 {
    // We need to reconfigure Splinter with user-specified key comparator fn.
    // Tear down default instance, and create a new one.
-   splinterdb_close(data->kvsb);
+   splinterdb_close(&data->kvsb);
```
gapisback, line 768: All these will go away once PR #397 is merged to /main.

gapisback:
HI, @carlosgarciaalvarado , @ajhconway @rtjohnso - I now have a working set of C-example programs for your attention and review.

I would like to first establish the ground work and pattern for these example programs, and then bring-in newer ones in follow-on PRs. It would be good if I can get feedback on these, so that these can be integrated sometime early next week.

I have one more in the pipeline following this PR. It's built on top of `splinterdb_custom_ipv4_addr_sortcmp_example.c` that is in this PR. In that program (named `splinterdb_ping_metrics_example.c`) I have a use-case of the `UPDATE` message interface, for a live ping-program that runs in a loop hitting on few www-sites. And we show how you can do a running aggregate of telemetry data from `ping` using user-provided merge callback methods.

That is an ever more interesting example, but will need some deep review. So am peeling that off for later, and will work w/ Rob on the mechanics of that program off-line (first).

include/splinterdb/splinterdb.h
gapisback: (Have reverted these diffs, so that this change-set reduces to just the C-sample programs. Ignore this comment.)

include/splinterdb/splinterdb.h
gapisback: (Have reverted these diffs, so that this change-set reduces to just the C-sample programs. Ignore this diff.)

src/rc_allocator.c
gapisback: (Have reverted these diffs, so that this change-set reduces to just the C-sample programs. Ignore this diff.)

gapisback: Update: Above error condition checks have been separately merged under commit SHA e677a4e.

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -259,6 +259,11 @@ ip4_ipaddr_keycmp(const char  *key1,
                   const char  *key2,
                   const size_t key2_len)
 {
+   // Deal with the case of initial keys, where one of the pair maybe uninit'ed
+   if ((key1_len == 0) || (key2_len == 0)) {
+      return KEYCMP_RV_KEY1_EQ_KEY2;
```
gapisback, line 264: **Reviewers: NOTE**: - Let's discuss: This seems to be the fix required to deal with the debug assertion failure.

src/trunk.c
```diff
@@ -1492,7 +1502,9 @@ trunk_add_pivot_new_root(trunk_handle *spl,
 {
    const char *pivot_key                       = trunk_get_pivot(spl, child, 0);
    __attribute__((unused)) const char *min_key = spl->cfg.data_cfg->min_key;
-   debug_assert(trunk_key_compare(spl, pivot_key, min_key) == 0);
+   __attribute__((unused)) const int   key_cmp_rv =
+      trunk_key_compare(spl, pivot_key, min_key);
+   debug_assert((key_cmp_rv == 0), "key_cmp_rv=%d\n", key_cmp_rv);
```
gapisback, line 1507: Change will be retained as it is useful on its own.

src/trunk.c
```diff
@@ -5764,6 +5776,10 @@ trunk_range_iterator_init(trunk_handle         *spl,
    // index btrees
    uint16 height = trunk_height(spl, node);
    for (uint16 h = height; h > 0; h--) {
+      int8 key_len = *range_itor->min_key;
+      platform_default_log("Dbg> range_itor->min_key='%.*s'\n",
+                           key_len, //  (int)trunk_key_size(spl),
+                           (range_itor->min_key + 1));
```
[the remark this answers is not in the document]
gapisback: @rtjohnso -- I did have one question about this `min_key` char * data.

The 0th byte seems to be the encoded "actual key-length" value, and the user-supplied `min_key` appears from the 1st byte. Where inside the BTree lookup / comparison routines are we dealing with this business of encoded keys? 

[I would have expected to see somewhere in code we are 'decoding' this encoded-key to get at the actual key-data, but not sure if I found that spot in code. ]

No rush; Will be good to talk this thru during our review discussion.

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
gapisback: Prior to this change, [CI debug quick-tests failed with an assertion](https://runway-ci.eng.vmware.com/builds/9920756).

examples/splinterdb_apis_example.c
```diff
@@ -0,0 +1,296 @@
...
+ * Lookup a value given a key.
+ * ---------------------------------------------------------------------------
+ */
+int
+do_lookup(splinterdb *spl_handle, const char *key_data, const size_t key_len)
+{
+   splinterdb_lookup_result result;
+   splinterdb_lookup_result_init(spl_handle, &result, 0, NULL);
+
+   slice key = slice_create(key_len, key_data);
```
carlosgarciaalvarado, line 204: Do we always need to provide the key size? or is there a chance I could lookup without providing this key_size?

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -0,0 +1,456 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * ----------------------------------------------------------------------------
+ * SplinterDB Advanced Iterators Example Program with custom sort-comparison.
+ *
+ * What's new beyond previous splinterdb_iterators_example.c?
```
carlosgarciaalvarado, line 8: I think this program makes the assumption that people will consume our examples in some sort of linear fashion. 
I found this assumption very weak (for example, I rarely do that).

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
+// Copyright 2022 VMware, Inc.
```
carlosgarciaalvarado, line 1: Somehow the intro example looks more complicated than lines 72-108. Could we simplify this even by removing simple comments (e.g. line 14, 17-16).

Let us recall that the name of the variables are "comments" too!

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
...
+ */
+#define APP_DB_NAME "splinterdb_intro_db"
+
+#define APP_DEVICE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
+#define APP_CACHE_SIZE_MB  64   // Size of cache; can be changed across boots
+
+/* Application declares the limit of key-sizes it intends to use */
+#define APP_MAX_KEY_SIZE ((int)100)
+
+/*
```
carlosgarciaalvarado, line 28: Probably we could get rid of 29-34.

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
...
+ * -------------------------------------------------------------------------------
+ * main() Driver for basic SplinterDB example program.
+ * We, intentionally, do not check or show-for error handling, as this is mostly
+ * a sample program to illustrative how-to use the APIs.
+ * -------------------------------------------------------------------------------
+ */
+int
+main()
+{
+   printf("     **** SplinterDB Basic Example program ****\n\n");
```
carlosgarciaalvarado, line 38: No need for 38.

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
...
+
+   /*
+    * Initialize data configuration, using default key-comparison handling.
+    */
+   data_config splinter_data_cfg;
+   default_data_config_init(APP_MAX_KEY_SIZE, &splinter_data_cfg);
+
+   /* Basic configuration of a SplinterDB instance */
```
carlosgarciaalvarado, line 46: Is the comment above applicable here?

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
...
+   splinterdb_cfg.disk_size  = (APP_DEVICE_SIZE_MB * K_MiB);
+   splinterdb_cfg.cache_size = (APP_CACHE_SIZE_MB * K_MiB);
+   splinterdb_cfg.data_cfg   = &splinter_data_cfg;
+
+   splinterdb *spl_handle = NULL; // To a running SplinterDB instance
+
+   int rc = splinterdb_create(&splinterdb_cfg, &spl_handle);
+   ex_msg("Created SplinterDB instance, dbname '%s'.\n\n", APP_DB_NAME);
+
+   /* Insert a few kv-pairs, describing properties of fruits. */
```
carlosgarciaalvarado, line 59: what about types of trees :) .... it is kind of our thing you know!

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
...
+   const char *fruit = "apple";
+   const char *descr = "An apple a day keeps the doctor away!";
+   slice       key   = slice_create((size_t)strlen(fruit), fruit);
+   slice       value = slice_create((size_t)strlen(descr), descr);
+   rc                = splinterdb_insert(spl_handle, key, value);
+   ex_msg("Inserted key '%s'\n", fruit);
+
+   fruit = "Orange";
+   descr = "Is a good source of vitamin-C.";
+   key   = slice_create((size_t)strlen(fruit), fruit);
```
carlosgarciaalvarado, line 69: This is a comment for us to think about the API but this `slice_create` seems like an overkill for many scenarios.

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,135 @@
...
+   rc = splinterdb_iterator_status(it);
+   splinterdb_iterator_deinit(it);
+
+   ex_msg("Found %d key-value pairs\n\n", i);
+
+   splinterdb_close(&spl_handle);
+   ex_msg("Shutdown SplinterDB instance, dbname '%s'.\n\n", APP_DB_NAME);
+
+   return rc;
+}
```
carlosgarciaalvarado, line 135: Argghh.....still feels too verbose.....are there any opportunities to simplify it?

src/trunk.c
```diff
@@ -663,7 +663,7 @@ static inline void                 trunk_node_unlock               (trunk_handle
 page_handle *                      trunk_alloc                     (trunk_handle *spl, uint64 height);
 static inline char *               trunk_get_pivot                 (trunk_handle *spl, page_handle *node, uint16 pivot_no);
 static inline trunk_pivot_data    *trunk_get_pivot_data            (trunk_handle *spl, page_handle *node, uint16 pivot_no);
-static inline uint16               trunk_find_pivot                (trunk_handle *spl, page_handle *node, const char *key, lookup_type comp);
+static inline uint16               trunk__find_pivot                (trunk_handle *spl, page_handle *node, const char *key, lookup_type comp, int lineno);
```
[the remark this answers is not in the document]
carlosgarciaalvarado: Okay, thanks

carlosgarciaalvarado (approved):
Overall the examples are valuables.  I have 3 main comments:

* Please try to reduce as much as possible the number of comments and lines. The longer the example, the least useful it is.
* Focus each example on something you want to teach to our users on its own. Don't assume that they build on their own. People don't explore them linearly, they will each of them looking for something specific.
* This PR should have been shorter ~7 PRs. Many times I lost the point of what you were trying to convey with each example.

gapisback:
Overall comments:

- Try to get rid of `example_common.h`. Make every `.c` file self-contained w/o nested `.h` files.
- Just use basic C-constructs like `printf()` etc.
- Get rid of all that `APP_` stuff ... simplify the code to be self-explanatory.
- Merge the close/open construct into hello-world program. Bag the `splinterdb_admin_config_example.c`. Do the close / open before iteration chunk in hello-world program.
- Basic program:
-

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -29,18 +29,13 @@
...
 /*
  * App-specific 'defaults' that can be parameterized, eventually.
  */
-#define APP_DB_NAME "splinterdb_custom_ipv4_sortcmp_example_db"
-
-#define APP_DEVICE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
-#define APP_CACHE_SIZE_MB  64   // Size of cache; can be changed across boots
+#define DB_FILE_NAME    "splinterdb_custom_ipv4_sortcmp_example_db"
+#define DB_FILE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
+#define CACHE_SIZE_MB   64   // Size of cache; can be changed across boots
```
gapisback, line 38: Consistently reworked above 3 names (as suggested in previous review meeting). This change is now applied to all programs.

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -90,6 +85,10 @@ int num_inet_addrs = (sizeof(inet_addr_info) / sizeof(*inet_addr_info));
 
 // clang-format on
 
+// Define min/max key values for user-defined IP4-addr keys
+#define IP4ADDR_MIN_KEY "0.0.0.0"
```
gapisback, line 89: I really, only, needed to re-use this one. (See line 355 below, to handle the case of iterate-for-all kv-pairs when an initial value is not specified.)

While at it, pulled out a `#define` for the sibling, too.

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -160,40 +149,35 @@ main()
    splinterdb_config splinterdb_cfg;
    configure_splinter_instance(&splinterdb_cfg,
                                &splinter_data_cfg,
-                               APP_DB_NAME,
-                               (APP_DEVICE_SIZE_MB * K_MiB),
-                               (APP_CACHE_SIZE_MB * K_MiB));
+                               DB_FILE_NAME,
```
gapisback, line 152: This is the only program (in this set) where configuration is wrapped into its own function.

The rest of the program is quite busy so I didn't want distracting configuration code inline'ed into main().

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -160,40 +149,35 @@ main()
...
 
-   ex_msg("Insert and iterate through %d {ipaddr} -> {www-url, metrics} "
+   printf("Insert and iterate through %d {ipaddr} -> {www-url, metrics} "
           "mapping table:\n",
           num_inet_addrs);
    do_inserts(spl_handle, inet_addr_info, num_inet_addrs);
 
-   do_iterate_all(spl_handle, num_inet_addrs);
+   // NULL start_key => Iterate through all key-value pairs
+   do_iterate_from(spl_handle, NULL);
```
gapisback, line 170: It was better to simplify the "iterate_all()" as a special case of `iterate_from()` by supplying a NULL `start_key` value.

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -160,40 +149,35 @@ main()
...
 
    const char *start_key = "99.84.238.130";
    do_iterate_from(spl_handle, start_key);
 
-   start_key = "5.79.89.114";
-   do_iterate_from(spl_handle, start_key);
-
-   start_key = "10.113.78.20";
-   do_iterate_from(spl_handle, start_key);
-
```
gapisback: Deleted repeated example cases that were adding no real value. Reduces code in `main()`

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -411,10 +346,14 @@ do_iterate_all(splinterdb *spl_handle, int num_keys)
...
 do_iterate_from(splinterdb *spl_handle, const char *from_key)
 {
-   ex_msg("Iterate through all the keys starting from '%s':\n", from_key);
+   printf("Iterate through all the keys starting from '%s':\n", from_key);
 
    splinterdb_iterator *it = NULL;
 
+   // Use min-key if user has not specified any start-from-key
+   if (!from_key) {
+      from_key = IP4ADDR_MIN_KEY;
```
gapisback, line 355: @rtjohnso - We should review this carefully.

I discovered that when user-specified sort-key comparison is used, specifying `NULL_SLICE` as initial start_key for "iterate_all" does not work. Something about the `iterator_init()` isn't getting setup correctly. And scan terminates w/o returning any kv-pair.

I reasoned that if user is using custom-sorting, just as they specified `min_key` while configuring SplinterDB, they will need to also specify this `min_key` when doing iterate-for-all w/o specifying start_key.

Makes sense?

examples/splinterdb_iterators_example.c
```diff
@@ -85,72 +58,47 @@ do_iterate_from(splinterdb *spl_handle, const char *from_key);
...
    default_data_config_init(APP_MAX_KEY_SIZE, &splinter_data_cfg);
 
    // Basic configuration of a SplinterDB instance
    splinterdb_config splinterdb_cfg;
-   configure_splinter_instance(&splinterdb_cfg,
-                               &splinter_data_cfg,
-                               APP_DB_NAME,
-                               (APP_DEVICE_SIZE_MB * K_MiB),
-                               (APP_CACHE_SIZE_MB * K_MiB));
+   memset(&splinterdb_cfg, 0, sizeof(splinterdb_cfg));
```
gapisback, line 70: Removed wrapper `configuration` method and inline'd code to shorten the whole program.

gapisback:
HI, @ajhconway @rtjohnso [fyi, @carlosgarciaalvarado ]

I am back to these sample C-programs that you all looked at early April. This work was on the shelf for last several weeks.

I went back and applied many / most of the review comments I got. Mainly:

- The programs are **much shorter now**, with most code inline'ed in `main()`. Except for the very last one, `splinterdb_custom_ipv4_addr_sortcmp_example.c` which has a fair bunch of complexity, so having functions is better there.
- Deleted `splinterdb_admin_config_example.c` and merged its one useful case of close / open SplinterDb into `splinterdb_intro_example.c`
- Deleted `example_common.h` and reworked `.c` code to be self-contained w/o external `.h` dependencies.

I have squashed all my review comments into this one commit, [Rework example C-programs based on review comments.](https://github.com/vmware/splinterdb/pull/399/commits/3d777f9a44dbe07b3b1fc71d4356b555e26fc851), where you can find a fuller description of the changes applied.

I think this change-set is quite ready to go, and I'd appreciate it if you could give it a 2nd look and move it along.

There is a follow-on change-set in my queue which brings in more sample programs that I'd like to bring to the review cycle once this goes in.

Thanks!

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
```diff
@@ -0,0 +1,395 @@
...
+
+// Describe the layout of fields in an IP4-address
+#define APP_IPV4_NUM_FIELDS 4
+#define APP_IPV4_NUM_DOTS   (APP_IPV4_NUM_FIELDS - 1)
+
+// Application declares the limit of key-sizes it intends to use
+#define APP_MAX_KEY_SIZE ((3 * APP_IPV4_NUM_FIELDS) + APP_IPV4_NUM_DOTS)
+
+// Max # of chars in a well-formed IP4 address, including null-terminator byte
+#define APP_IPV4_MAX_KEY_BUF_SIZE (APP_MAX_KEY_SIZE + 1)
```
ajhconway, line 48: I would get rid of the `APP_` prefix

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
ajhconway: This seems like a bug to me

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,137 @@
...
+
+#include "splinterdb/default_data_config.h"
+#include "splinterdb/splinterdb.h"
+
+#define DB_FILE_NAME    "splinterdb_intro_db"
+#define DB_FILE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
+#define CACHE_SIZE_MB   64   // Size of cache; can be changed across boots
+
+/* Application declares the limit of key-sizes it intends to use */
+#define APP_MAX_KEY_SIZE ((int)100)
```
ajhconway, line 20: Remove `APP_`

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,137 @@
...
+
+/*
+ * -------------------------------------------------------------------------------
+ * We, intentionally, do not check or show-for error handling, as this is mostly
```
ajhconway, line 24: I don't understand this sentence. What does "show-for" mean here?

examples/splinterdb_intro_example.c
```diff
@@ -0,0 +1,137 @@
...
+   fruit = "Orange";
+   key   = slice_create((size_t)strlen(fruit), fruit);
+   rc    = splinterdb_lookup(spl_handle, key, &result);
+   rc    = splinterdb_lookup_result_value(spl_handle, &result, &value);
+   if (!rc) {
+      printf("Found key: '%s', value: '%.*s'\n",
+             fruit,
+             (int)slice_length(value),
+             (char *)slice_data(value));
+   }
```
ajhconway, line 86: Is this really the way our API works now? We should really not be using the `rc` to indicate a non-error condition, such as a negative lookup.

examples/splinterdb_intro_example.c
ajhconway: I agree with this old comment, but I think that the path towards that will be adding some convenience API functions and revisiting this example later.

examples/splinterdb_iterators_example.c
```diff
@@ -0,0 +1,168 @@
...
+
+#include "splinterdb/default_data_config.h"
+#include "splinterdb/splinterdb.h"
+
+#define DB_FILE_NAME    "splinterdb_iterators_example_db"
+#define DB_FILE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
+#define CACHE_SIZE_MB   64   // Size of cache; can be changed across boots
+
+/* Application declares the limit of key-sizes it intends to use */
+#define APP_MAX_KEY_SIZE ((int)100)
```
ajhconway, line 18: Remove `APP_`

examples/splinterdb_wide_values_example.c
```diff
@@ -0,0 +1,115 @@
...
+#define DB_FILE_NAME    "splinterdb_wide_values_example_db"
+#define DB_FILE_SIZE_MB 1024 // Size of SplinterDB device; Fixed when created
+#define CACHE_SIZE_MB   64   // Size of cache; can be changed across boots
+
+/* Application declares the limit of key-sizes it intends to use */
+#define APP_MAX_KEY_SIZE ((int)100)
+
+/* Avg value size and max value sizes we expect to deal with in this program */
+#define APP_AVG_VALUE_SIZE ((int)128)
+#define APP_MAX_VALUE_SIZE ((int)1024)
```
ajhconway, line 22: Remove `APP_`

src/trunk.c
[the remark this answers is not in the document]
ajhconway: Don't forget to back this out.

src/trunk.c
ajhconway: Use the `debug_only` macro here for `key_cmp_rv`

ajhconway (approved):
Approving assuming that minor comments are addressed.

src/trunk.c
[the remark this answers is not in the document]
gapisback: Done, @ajhconway - Have reverted the `trunk__find_pivot()` wrapper, which was added for debugging.

src/trunk.c
gapisback: Done. Nice tip, @ajhconway -- Forgot that this existed.

What's more .. the line now fits within 80-chars, so no more annoying `clang-format` re-indents on this line:

```
+   debug_only const int key_cmp_rv = trunk_key_compare(spl, pivot_key, min_key);
```

examples/splinterdb_intro_example.c
gapisback: (Noted; no changes resulting in code.)

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
gapisback: Done. Replaced all mnemonics consistently with `IP4_` prefix.

examples/splinterdb_intro_example.c
gapisback: Lazy commentary. Rephrased as:

```
 24  * We, intentionally, do not check for errors or show error handling, as this is
 25  * mostly a sample program to illustrative how-to use the APIs.
 ```

File: examples/splinterdb_custom_ipv4_addr_sortcmp_example.c (hunk 26 above)
Comment (gapisback) [reply examples/splinterdb_custom_ipv4_addr_sortcmp_example.c:355]: Hmmm ... @ajhconway - I did debug this somewhat, to see what was going wrong in `iterator_init()`.

Honestly, I did not really get to the bottom of it, and "convinced" myself that specifying `from_key` as shown in the code is the acceptable fix.

@rtjohnso -- What's your call? 

I'd like to try to merge this PR in as-is, register a potential issue and debug that off-line. If we do conclude that it's a bug, I'd like to fix the code and the example code in a follow-on. That way, this PR's progress is not blocked by a potential bug.

If you both think that this really a bug, I will go back and investigate this via a unit-test, and debug it fully.
Let me get back to you on that.

File: examples/splinterdb_intro_example.c (hunk 31 above)
Comment (gapisback) [reply examples/splinterdb_intro_example.c:86]: @ajhconway -- Currently, `splinterdb_lookup_result_value()` is implemented to return an `int`.

It calls `splinterdb_lookup_found()` defined as a `bool` function. If that were to ever fail, `splinterdb_lookup_result_value` could return a non-zero `rc`.

---

The "problem" in these interfaces is that the calling-sequence is something like this:

(Below, I added the `printf()` after `splinterdb_lookup()` to make my point.)

```
 79    rc    = splinterdb_lookup(spl_handle, key, &result);
 80    printf("splinterdb_lookup('Orange') rc=%d\n", rc);
 81    rc    = splinterdb_lookup_result_value(spl_handle, &result, &value);
 
  89    // Handling non-existent keys
 90    fruit = "Banana";
 91    key   = slice_create((size_t)strlen(fruit), fruit);
 92    rc    = splinterdb_lookup(spl_handle, key, &result);
 93    printf("splinterdb_lookup('Banana') rc=%d\n", rc);
 94    rc    = splinterdb_lookup_result_value(spl_handle, &result, &value);
 ```
 
 The output you will get when you run this is:
 
 ```
splinterdb_lookup('Orange') rc=0
Found key: 'Orange', value: 'Is a good source of vitamin-C.'
splinterdb_lookup('Banana') rc=0
Key: 'Banana' not found. (rc=22)
```

Point being: `splinterdb_lookup()` always returns 0 for both keys that exist or do not exist.
So one has to check further by calling `splinterdb_lookup_result_value`.

I agree that this is very awkward coding. I had raised similar issues in streamlining these lookup interfaces, but those were put on-hold till we get time for a revisit.

For now, I think I will leave the examples as coded till we get the time to overhaul these interfaces.

File: examples/splinterdb_iterators_example.c (hunk 32 above)
Comment (gapisback) [reply examples/splinterdb_iterators_example.c:18]: To make this quicker to resolve ... I changed all tokens named `APP_` to `USER_<something>`. E.g.

```
 17 /* Application declares the limit of key-sizes it intends to use */
 18 #define USER_MAX_KEY_SIZE ((int)100)
 19
 20 /* Avg value size and max value sizes we expect to deal with in this program */
 21 #define USER_AVG_VALUE_SIZE ((int)128)
 22 #define USER_MAX_VALUE_SIZE ((int)1024)
 ```
 
 Hope this closes out this remark.

examples/splinterdb_custom_ipv4_addr_sortcmp_example.c
gapisback: @ajhconway , @rtjohnso -- We should discuss this in our 1pm tomorrow, but for the record, here's my assessment based on debugging this issue.

----
I found what was a piece of "compensating code" in this PR that was creating / masking some issue which Alex pointed out as "the bug". I tracked it down to this piece of code in this sample program itself, in the custom-comparator function; i.e. here:

```
242 ip4_ipaddr_keycmp(const char  *key1,
243                   const size_t key1_len,
244                   const char  *key2,
245                   const size_t key2_len)
246 {
247    // Deal with the case of initial keys, where one of the pair maybe uninit'ed
248    if ((key1_len == 0) || (key2_len == 0)) {
249       return KEYCMP_RV_KEY1_EQ_KEY2;
250    }
251
```

----
On L249, we return "keys-are-equal", leading to some unexpected behaviour. 

When I took this block out, this example program runs cleanly even for the `do_iterate_from()` case when `start_key = NULL`.

----
I then wondered and debugged this code in `splinterdb_iterator_init()`:

```
800 splinterdb_iterator_init(const splinterdb     *kvs,      // IN
[...]

814    char start_key_buffer[MAX_KEY_SIZE] = {0};
815    if (!slice_is_null(start_key)) {
816       int rc = encode_key(MAX_KEY_SIZE, start_key_buffer, start_key);
817       if (rc != 0) {
818          return rc;
819       }
820    }
821
822    platform_status rc = trunk_range_iterator_init(
823       kvs->spl, range_itor, start_key_buffer, NULL, UINT64_MAX);
```

On L 823, even when `slice_is_null(start_key()` is TRUE, we are passing-down `start_key_buffer`, which would be uninitialized.

I played around with passing-down NULL to the callee here when `start_key` is NULL  [which seems to be the reasonable thing to do ]  ... and that seems to work too.

So ... now all this is very confusing to me.

It's not clear what the right interface should be. I don't want to hack-patch this ... so let's talk tomorrow.

PR #400, opened 2022-04-02
Add checks and unit-tests for few more invalid config params.

gapisback:
Convert some existing platform_assert()s that validate disk-size,
extent-size and related config params to proper error messages.
Return proper $rc, and add few negative unit-tests to validate
these user-errors. (Checks were discovered while building C-sample
programs, through wrong-code.) Sanitize text of a few few existing
error messages.


src/rc_allocator.c
```diff
@@ -236,6 +236,45 @@ rc_allocator_config_init(rc_allocator_config *allocator_cfg,
...
+      platform_error_log("Configured disk size %lu bytes is invalid.\n",
+                         cfg->capacity);
+      return STATUS_BAD_PARAM;
+   }
+   if (cfg->extent_capacity == 0) {
+      platform_error_log("Configured extent capacity %lu bytes is invalid.\n",
+                         cfg->extent_capacity);
+      return STATUS_BAD_PARAM;
+   }
+   if (cfg->capacity != (cfg->io_cfg->extent_size * cfg->extent_capacity)) {
```
rtjohnso, line 265: Not sure how I feel about this check.

This is guaranteed to be true of configs created by `rc_allocator_config_init`.

I guess this is just at init time, so it's ok.

But, if you're going to replicate constraints that are established by `config_init`, then might as well also check
- that page_size divides extent_size
- that page_capacity * page_size == capacity

src/rc_allocator.c
```diff
@@ -262,9 +301,17 @@ rc_allocator_init(rc_allocator        *al,
...
    al->heap_id     = hid;
 
+   rc = rc_allocator_valid_config(cfg);
+   if (!SUCCESS(rc)) {
+      return rc;
+   }
    platform_assert(cfg->io_cfg->page_size % 4096 == 0);
    platform_assert(cfg->capacity
-                   == cfg->io_cfg->extent_size * cfg->extent_capacity);
+                      == cfg->io_cfg->extent_size * cfg->extent_capacity,
```
rtjohnso, line 310: This assertion is now redundant.  Delete.

tests/unit/unit_tests.h
```diff
@@ -8,5 +8,6 @@
 /* Name of SplinterDB device created for unit-tests */
 #define TEST_DB_NAME "unit_tests_db"
 
-#define Mega (1024UL * 1024UL)
+#define Kilo (1024UL)
```
rtjohnso, line 11: There are already KiB, MiB, and GiB #defines from platform.  Can we not just use those?

rtjohnso (approved):

src/rc_allocator.c
gapisback: @rtjohnso - Here is what I did:

1. This one is currently not testable, as `laio_config_valid()` implements checks that we only support a fixed page-size and extent-size. So this division check is already enforced. [ I can't add any further checks as the can't be tested with the current code-flow. When we do allow different page-sizes / extent-sizes, such checks will need to be added. ]
2. Added this check, and added new test case `test_disk_size_not_integral_multiple_of_page_size` for this check.

src/rc_allocator.c
gapisback: Deleted. Also L308 came out. As did this check below, which is now caught by a real error message:

```
-   platform_assert(cfg->capacity
-                   == cfg->io_cfg->page_size * cfg->page_capacity);
```

Issue #305, opened 2022-03-02
splinter_test --perf runs into Assertion from clockcache.c:2131:clockcache_get_internal(): "allocator_get_ref(cc->al, base_addr) > 1". 


gapisback:
Following test run was seen to run into an assertion twice; once in a debug-gcc run and once in an ASAN run on CI.

```
+ bin/driver_test splinter_test --perf --max-async-inflight 0 --num-insert-threads 4 --num-lookup-threads 4 --num-range-lookup-threads 0 --tree-size-gib 2 --cache-capacity-mib 512
bin/driver_test: splinterdb_build_version d7643d50
Dispatch test splinter_test
fingerprint_size: 27
Running splinter_test with 1 caches
splinter_test: splinter performance test started with 1                tables
inserting  74% complete for table 0Assertion failed at src/clockcache.c:2131:clockcache_get_internal(): "allocator_get_ref(cc->al, base_addr) > 1". 
./test.sh: line 41:   279 Aborted                 "$@"
```

See these internal slack threads [here, 2/22/2022](https://vmware.slack.com/archives/G01MEDQ05MX/p1645550951000009) and [here, 3/2022](https://vmware.slack.com/archives/G01MEDQ05MX/p1646207005770419), where this failure was reported and discussed.

After the 1st instance, there was some speculation that this may have been due to `make` and incomplete build issues. 
Some fixes were applied to `Makefile`, addressing rules for unit-tests, but this failure emanates from `bin/driver_test`, so those fixes are not applicable.

Also, the three recent failures were all seen on CI-builds, so there should have been no issue about a `make clean` build etc.

gapisback:
**Weirdly .. this recurred** on CI-jobs in branch [agurajada/341-PgExt-size-limits](https://github.com/vmware/splinterdb/tree/agurajada/341-PgExt-size-limits), for a completely unrelated change-set.

Rebased to /main and [here's the failing CI-job](https://runway-ci.eng.vmware.com/builds/9789947): 

```
+ bin/driver_test splinter_test --perf --max-async-inflight 0 --num-insert-threads 4 --num-lookup-threads 4 --num-range-lookup-threads 0 --tree-size-gib 2 --cache-capacity-mib 512
bin/driver_test: splinterdb_build_version 30cc7439
Dispatch test splinter_test
Fingerprint size 29 too large, max value size is 5, setting to 27
fingerprint_size: 27
filter-index-size: 256 is too small, setting to 512
[...]

PR #401, opened 2022-04-06
(#305) Add instrumentation to track assertion in clockcache_get_internal()

gapisback:
Test splinter_test --perf, and some other tests, are randomly tripping up a
debug assertion in clockcache_get_internal(), which reports that we
are trying to cache a page from an unallocated extent. This problem is
not reproducible. This commit improves the assertion message to report
addresses involved, and also calls-out to a print method to dump out
allocated extents bitmap. (This should only occur very rarely, if at all.)

----

This arises from the triaging of the assertion failure reported in issue #305. Several attempts have been made to try to repro it, but no avail.

Rather than closing out the issue as "not reproducible", I am adding some instrumentation, that will only trip in debug builds, to report more data when the assertion trips. This may give us some more clues for further troubleshooting.

A request was made by @ajhconway to may be print the call-stack. The work to generate a human-readable stack, mapping addresses to fn-names is a possibility, but will take longer than we have at this time. Building that stack trace facility is a separate dev-item, which is not on the priority radar right now. For now, this minimalistic instrumentation should suffice.

### DIFF
--- a/src/allocator.h
+++ b/src/allocator.h
@@ -90,7 +90,6 @@ typedef platform_status (*alloc_super_addr_fn)(allocator        *al,
                                                uint64           *addr);
 typedef void (*remove_super_addr_fn)(allocator *al, allocator_root_id spl_id);
 typedef uint64 (*get_size_fn)(allocator *al);
-typedef uint64 (*get_num_fn)(allocator *al, uint64 addr);
 
 typedef void (*print_fn)(allocator *al);
 typedef void (*assert_fn)(allocator *al);
@@ -113,8 +112,6 @@ typedef struct allocator_ops {
    get_size_fn in_use;
 
    get_size_fn get_capacity;
-
-   get_num_fn get_extent_num;
 
    assert_fn assert_noleaks;
 
@@ -182,12 +179,6 @@ static inline uint64
 allocator_get_capacity(allocator *al)
 {
    return al->ops->get_capacity(al);
-}
-
-static inline uint64
-allocator_get_extent_num(allocator *al, uint64 page_addr)
-{
-   return al->ops->get_extent_num(al, page_addr);
 }
 
 static inline void
--- a/src/allocator.h
+++ b/src/allocator.h
@@ -90,6 +90,7 @@ typedef platform_status (*alloc_super_addr_fn)(allocator        *al,
                                                uint64           *addr);
 typedef void (*remove_super_addr_fn)(allocator *al, allocator_root_id spl_id);
 typedef uint64 (*get_size_fn)(allocator *al);
+typedef uint64 (*get_num_fn)(allocator *al, uint64 addr);
 
 typedef void (*print_fn)(allocator *al);
 typedef void (*assert_fn)(allocator *al);
@@ -113,10 +114,12 @@ typedef struct allocator_ops {
 
    get_size_fn get_capacity;
 
+   get_num_fn get_extent_num;
+
    assert_fn assert_noleaks;
 
    print_fn print_stats;
-   print_fn debug_print;
+   print_fn print_allocated;
 } allocator_ops;
 
 // To sub-class cache, make a cache your first field;
@@ -181,6 +184,12 @@ allocator_get_capacity(allocator *al)
    return al->ops->get_capacity(al);
 }
 
+static inline uint64
+allocator_get_extent_num(allocator *al, uint64 page_addr)
+{
+   return al->ops->get_extent_num(al, page_addr);
+}
+
 static inline void
 allocator_assert_noleaks(allocator *al)
 {
@@ -194,9 +203,9 @@ allocator_print_stats(allocator *al)
 }
 
 static inline void
-allocator_debug_print(allocator *al)
+allocator_print_allocated(allocator *al)
 {
-   return al->ops->debug_print(al);
+   return al->ops->print_allocated(al);
 }
 
 #endif // __ALLOCATOR_H
--- a/src/clockcache.c
+++ b/src/clockcache.c
@@ -2128,7 +2128,25 @@ clockcache_get_internal(clockcache   *cc,       // IN
    platform_status   status;
    uint64            start, elapsed;
 
-   debug_assert(allocator_get_ref(cc->al, base_addr) > 1);
+#if SPLINTER_DEBUG
+   uint8 extent_ref_count = allocator_get_ref(cc->al, base_addr);
+
+   // Dump allocated extents info for deeper debugging.
+   if (extent_ref_count <= 1) {
+      allocator_print_allocated(cc->al);
+   }
+   debug_assert((extent_ref_count > 1),
+                "Attempt to get a buffer for page addr=%lu"
+                ", page type=%d ('%s'),"
+                " from extent addr=%lu, (extent number=%lu)"
+                ", which is an unallocated extent, extent_ref_count=%u.",
+                addr,
+                type,
+                page_type_str[type],
+                base_addr,
+                allocator_get_extent_num(cc->al, base_addr),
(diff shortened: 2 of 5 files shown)


src/allocator.h
```diff
@@ -113,10 +114,12 @@ typedef struct allocator_ops {
 
    get_size_fn get_capacity;
 
+   get_num_fn get_extent_num;
+
    assert_fn assert_noleaks;
 
    print_fn print_stats;
-   print_fn debug_print;
+   print_fn print_allocated;
```
gapisback, line 122: This is a pending cleanup that I was supposed to under SHA d585788, but didn't get around to it then. Am taking this opportunity to name the virtual functions consistently with the called function.

src/allocator.h
```diff
@@ -181,6 +184,12 @@ allocator_get_capacity(allocator *al)
    return al->ops->get_capacity(al);
 }
 
+static inline uint64
+allocator_get_extent_num(allocator *al, uint64 page_addr)
+{
+   return al->ops->get_extent_num(al, page_addr);
```
gapisback, line 190: Externalizing this as virtual fn so we can print this value in `clockcache.c` when the assertion trips.

src/clockcache.c
```diff
@@ -2128,7 +2128,25 @@ clockcache_get_internal(clockcache   *cc,       // IN
    platform_status   status;
    uint64            start, elapsed;
 
-   debug_assert(allocator_get_ref(cc->al, base_addr) > 1);
+#if SPLINTER_DEBUG
+   uint8 extent_ref_count = allocator_get_ref(cc->al, base_addr);
+
+   // Dump allocated extents info for deeper debugging.
+   if (extent_ref_count <= 1) {
+      allocator_print_allocated(cc->al);
```
gapisback, line 2136: This may be more than we want ... but let's merge this and watch how things work out.

The said assertion has not reappeared in last 2-3 some weeks.

src/clockcache.c
```diff
@@ -2128,7 +2128,25 @@ clockcache_get_internal(clockcache   *cc,       // IN
...
+   }
+   debug_assert((extent_ref_count > 1),
```
gapisback, line 2138: Forcing this assertion to trip, by futzing this condition, the message will read like so, below:

```
Assertion failed at src/clockcache.c:2138:clockcache_get_internal(): "(extent_ref_count > 2)". 
Attempt to get a buffer for page addr=4096, page type=5 ('superblock'), from extent addr=0, (extent number=0),
which is an unallocated extent, extent_ref_count=2.
Aborted (core dumped)
```

src/rc_allocator.c
```diff
@@ -140,6 +139,16 @@ rc_allocator_get_capacity_virtual(allocator *a)
    return rc_allocator_get_capacity(al);
 }
 
+uint64
+rc_allocator_get_extent_num(rc_allocator *al, uint64 page_addr);
+
+uint64
+rc_allocator_get_extent_num_virtual(allocator *a, uint64 page_addr)
```
gapisback, line 146: Used on L2147 in `clockcache.c` for extra debugging info.

src/rc_allocator.h
```diff
@@ -88,6 +88,20 @@ typedef struct rc_allocator {
 } rc_allocator;
 
 
+/* Is page address 'addr' is a valid extent address */
+static inline bool
+rc_allocator_valid_extent_addr(rc_allocator *al, uint64 addr)
+{
+   return (addr % al->cfg->io_cfg->extent_size == 0);
```
gapisback, line 95: I thought I would add some validation under the assertion to check if the extent-addr is valid [etc.], but I didn't find this necessary, after all.

Having modularized this to a re-usable function, I left it in for now. If in future, we need external modules to check if an extent-addr is valid, this fn will come in handy. (We'd have to slap a virtual fn interface on top of it.)

Would like to leave this in, even though it's slightly unrelated to this fix, per se'.

src/rc_allocator.c
ajhconway: I'm not a fan of this function. We don't use this outside of the allocator, and `extent_num` isn't a descriptive name. Rather than pipe all this through for a probably temporary `debug_assert`, I'd rather just have `base_addr / cache_extent_size(cfg)` in `clockcache.c`.

ajhconway (changes requested):
I'd rather not expose `extent_num`, since it doesn't have a clear meaning outside of `rc_allocator.c`. We can just divide by `cache_extent_size(cfg)` to get the same result.

src/rc_allocator.c
gapisback: Done. Withdrew `allocator_get_extent_num()`, `typedef uint64 (*get_num_fn)` and associated changes to pipe the extent_num thru a virtual fn handle.

src/rc_allocator.h
gapisback: Also, just so you know, I have retained this sibling method, too, `rc_allocator_extent_num()`.

I want to retain both as they will be useful to call from the debugger, while examining large 'addr' values, to figure out the semantics that these two fns offer.

ajhconway (approved):
Minor changes, then ship it

src/rc_allocator.h
```diff
@@ -88,6 +88,23 @@ typedef struct rc_allocator {
...
+{
+   return ((addr % al->cfg->io_cfg->extent_size) == 0);
+}
+
+/*
+ * Convert page-address to the extent number of extent containing this page.
+ * Returns the index into the allocated extents reference count array.
+ */
+static inline uint64
+rc_allocator_extent_num(rc_allocator *al, uint64 addr)
```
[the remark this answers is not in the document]
gapisback: Changed the fn's name to `rc_allocator_extent_number`. 

I think the better name would be `rc_allocator_extentid`; as these things are really an 'extent ID', starting from 0, 1, 2, ... I had sort-of mooted bringing-in that concept thru a previous diagnostic change, but you weren't very convinced. (We can still go back to extent ID convention, sometime in the future after we talk this through. And the rework in this name will be minor.)

I wasn't exposing it as an interface thru one of the virtual functions.

But I had left it in the `.h` file so as to document this as a potentially usable extern interface. I suspect we may find need for it in the future. In any case, for now, moved this back to the `.c` file.

src/rc_allocator.h
```diff
@@ -88,6 +88,23 @@ typedef struct rc_allocator {
...
+{
+   return (addr / al->cfg->io_cfg->extent_size);
```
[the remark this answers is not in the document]
gapisback: No, `addr` can be any page address. This function maps every page in an extent to its extent ID (extent number).
Added comments to the prolog to clarify this appropriately.

src/rc_allocator.h
```diff
@@ -88,6 +88,23 @@ typedef struct rc_allocator {
 } rc_allocator;
 
 
+/* Is page address 'addr' a valid extent address? */
+static inline bool
+rc_allocator_valid_extent_addr(rc_allocator *al, uint64 addr)
```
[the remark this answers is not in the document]
gapisback: I think using `base_addr` for the argument's name is fine, but I would like to retain the name as it is written now.

In the rest of the `rc_allocator` module we do talk about allocating extents v/s pages (in mini-allocator). SO I would like to formalize that concept of extents through this name.

Over time, I have come to realize that `_is_` in a boolean function's name, like in `rc_allocator_is_valid_extent_addr` is not any more clearer than simply stating it as `rc_allocator_valid_extent_addr()`.

For instance code like : `if rc_allocator_valid_extent_addr(page_addr)`, I think reads just fine.

So have left the fn's name as written, if you don't mind.

PR #405, opened 2022-05-04
more styles for homepage

michow:
Signed-off-by: michow <michow@smalltalk.agency>


michow:
@carlosgarciaalvarado more styles for homepage

PR #414, opened 2022-05-29
use xargs -I{} instead of the deprecated xargs -i

ericlagergren:
Signed-off-by: Eric Lagergren <eric@ericlagergren.com>


ajhconway (approved):
Thanks for the PR! Looks good to me.

ajhconway:
msan timed out, but shouldn't be affected. Merging without waiting.

PR #415, opened 2022-05-29
remove dependency on malloc.h

ericlagergren:
It's non-standard and not actually used.

Signed-off-by: Eric Lagergren <eric@ericlagergren.com>


ajhconway (approved):
LGTM.

PR #416, opened 2022-06-02
Update download page to reference to main zip file

carlosgarciaalvarado:
Updating content to link to SplinterDB's main branch zip file.


gapisback (approved):
Change seems straightforward enough. I tested this off the private branch and seems like the zip file download is working well.

(I'll be more comfortable checking this out once this is merged.)

Approving this ... to get the ball rolling on this change.

Issue #120, opened 2021-11-12
Handle MESSAGE_TYPE_INVALID


ajhconway:
The `btree_insert` and `merge` code does not specially handle the `MESSAGE_TYPE_INVALID` case. We should fail the insert and handle it with some sort of error in the merge.

PR #420, opened 2022-06-08
(#120) Handle invalid user-type messages in 'btree_insert' and 'btree_pack'

deukyeon:
In `btree_insert` and `btree_pack`, invalid user-type messages are not
handled properly. This change makes those functions return an error
when they receive invalid user-type messages. An error that
`btree_insert` returns is forwarded to an application. On the other
hand, when `btree_pack` returns an error, the program will be
asserted.

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -1713,6 +1713,10 @@ btree_insert(cache              *cc,         // IN
       return STATUS_BAD_PARAM;
    }
 
+   if (message_is_invalid_user_type(msg)) {
+      return STATUS_BAD_PARAM;
+   }
+
    btree_node root_node;
    root_node.addr = root_addr;
 
@@ -2806,13 +2810,17 @@ btree_pack_setup_finish(btree_pack_req *req, slice first_key)
    btree_pack_node_init_hdr(req->cfg, req->edge[0].hdr, req->next_extent, 0);
 }
 
-static inline void
+static inline platform_status
 btree_pack_loop(btree_pack_req *req, // IN/OUT
                 slice           key, // IN
                 message         msg, // IN
                 bool           *at_end)        // IN/OUT
 {
    log_trace_key(key, "btree_pack_loop");
+
+   if (message_is_invalid_user_type(msg)) {
+      return STATUS_INVALID_STATE;
+   }
 
    if (!btree_set_leaf_entry(req->cfg,
                              req->edge[0].hdr,
@@ -2929,6 +2937,8 @@ btree_pack_loop(btree_pack_req *req, // IN/OUT
 
    iterator_advance(req->itor);
    iterator_at_end(req->itor, at_end);
+
+   return STATUS_OK;
 }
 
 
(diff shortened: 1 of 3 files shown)


src/btree.h
```diff
@@ -147,6 +147,11 @@ typedef struct btree_iterator {
    debug_code(char debug_prev_end_key[MAX_KEY_SIZE]);
 } btree_iterator;
 
+typedef enum btree_pack_err {
+   BTREE_PACK_ERR_NONE = 0,
+   BTREE_PACK_ERR_INVALID_MSG
+} btree_pack_err;
+
```
deukyeon, line 154: Because `btree_pack_loop` is a void function, `btree_pack_err` indicates which error occurs while handling `btree_pack_req`. Different error types can be added here if necessary.

In terms of the invalid enum value, `_NONE`, rather than `_INVALID` could clarify that there is no error.

src/btree.c
```diff
@@ -2990,6 +2999,17 @@ btree_pack_can_fit_tuple(btree_pack_req *req, slice key, message data)
...
+btree_pack_abort(btree_pack_req *req)
+{
+   btree_dec_ref_range(req->cc,
+                       req->cfg,
+                       req->root_addr,
+                       NULL_SLICE,
+                       NULL_SLICE,
+                       PAGE_TYPE_BRANCH);
+}
+
```
deukyeon, line 3012: `btree_pack_abort` is supposed to destroy the entire data structure. If there is more things to do, let me know.

src/btree.c
```diff
@@ -3021,6 +3041,15 @@ btree_pack(btree_pack_req *req)
    {
       iterator_get_curr(req->itor, &key, &data);
       btree_pack_loop(req, key, data, &at_end);
+
+      if (!at_end) {
```
rtjohnso, line 3045: I don't think this should be guarded by `!at_end`.

src/btree.c
```diff
@@ -2814,6 +2818,11 @@ btree_pack_loop(btree_pack_req *req, // IN/OUT
 {
    log_trace_key(key, "btree_pack_loop");
 
+   if (message_is_invalid_user_type(msg)) {
```
rtjohnso, line 2821: Why not return the error as the return value of `btree_pack_loop`?

src/btree.h
```diff
@@ -147,6 +147,11 @@ typedef struct btree_iterator {
    debug_code(char debug_prev_end_key[MAX_KEY_SIZE]);
 } btree_iterator;
 
+typedef enum btree_pack_err {
```
rtjohnso, line 150: If you have `btree_pack_loop` return the error instead of sticking it inside the `req`, then you can move this enum into `btree.c`.  It doesn't need to be part of the interface to the btree.

src/btree.h
```diff
@@ -169,6 +174,7 @@ typedef struct btree_pack_req {
    uint64 num_tuples;    // no. of tuples in the output tree
    uint64 key_bytes;     // total size of keys in tuples of the output tree
    uint64 message_bytes; // total size of msgs in tuples of the output tree
+   btree_pack_err error; // the status of error
```
rtjohnso, line 177: Just modify `btree_pack_loop` to return the error code.

rtjohnso (changes requested):

src/btree.c
deukyeon: @rtjohnso -- I felt like that an error handling like this could be possibly used in other functions related to btree packing. But now I think that returning the error is a better way for this because we don't need to change the btree interface.

src/btree.c
deukyeon: @rtjohnso -- Got it. Let me add the code.

src/btree.c
deukyeon: @rtjohnso -- Got it. As I change the code for `btree_pack_loop` to return an error, this guard is also removed.

ajhconway (approved):
This looks pretty good to me.

I'd prefer to use `platform_status` rather than defining new btree pack error codes.

rtjohnso (approved):
Fantastic.

Alex's suggestion to just use the platform_status is also fine.  If you change to do that, then you don't need my approval.  Thank you!

src/btree.c
deukyeon: @ajhconway -- I think that it is better than adding a new error type. Let me do that.

ajhconway (approved):
Looks great!

Issue #419, opened 2022-06-08
Suspected code error in splinterdb_iterator_init(), results in seg-fault and / or unpredictable behaviour in iterator init() logic


gapisback:
### Describe the bug

This issue was discovered while developing C-sample programs under PR #399. An attempt was made to extract out a stand-alone repro with a test-case.

The following simple test case results in an assertion failure from `trunk_range_iterator_init()` -> `trunk_find_pivot()`.

The test is simply doing very basic configuration of SplinterDB, then `splinterdb_create()`, followed by `splinterdb_iterator_init`. Stack is as follows:

```
#2  0x00007ffff7fa3645 in platform_assert_false (filename=0x7ffff7fad9db "src/trunk.c",
    linenumber=1373, functionname=0x7ffff7fb1ab0 <__FUNCTION__.9172> "trunk_find_pivot",
    expr=0x7ffff7fadbb2 "cmp <= 0", message=0x7ffff7fad3aa "")
    at src/platform_linux/platform.c:335
335	   abort();
(gdb)
#3  0x00007ffff7f79c5b in trunk_find_pivot (spl=0x5555558082c0, node=0x5555557a8e80,
    key=0x555555dc0f88 "", comp=less_than_or_equal) at src/trunk.c:1373
1373	            debug_assert(cmp <= 0);
(gdb)
#4  0x00007ffff7f88da2 in trunk_range_iterator_init (spl=0x5555558082c0,
    range_itor=0x555555dc0b40, min_key=0x7fffffffe590 "", max_key=0x0,
    num_tuples=18446744073709551615) at src/trunk.c:5768
5768	         trunk_find_pivot(spl, node, range_itor->min_key, less_than_or_equal);
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7d0b859 in __GI_abort () at abort.c:79
#2  0x00007ffff7fa3645 in platform_assert_false (filename=0x7ffff7fad9db "src/trunk.c",
    linenumber=1373, functionname=0x7ffff7fb1ab0 <__FUNCTION__.9172> "trunk_find_pivot",
    expr=0x7ffff7fadbb2 "cmp <= 0", message=0x7ffff7fad3aa "")
    at src/platform_linux/platform.c:335
#3  0x00007ffff7f79c5b in trunk_find_pivot (spl=0x5555558082c0, node=0x5555557a8e80,
[...]

PR #421, opened 2022-06-09
(#419) Fix bug in splinterdb_iterator_init() when min_key is configured.

gapisback:
This commit fixes a latent bug in splinterdb_iterator_init() which might
cause splinterdb_iterator_valid() to incorrectly return FALSE when
min_key is configured. Add a test case to demonstrate the problem,
albeit one that only fails in debug-mode.

As a defensive check against test/user-code attempting to add invalid
keys, add debug assertion checks in insert code-flow to validate that
keys are within the valid [min,max] key-range.

(This issue and the consequent bug-fix is rather temporary and may all
be dismantled once min/max key configuration is undone.)

### DIFF
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -292,4 +292,7 @@ splinterdb_iterator_get_current(splinterdb_iterator *iter, // IN
 int
 splinterdb_iterator_status(const splinterdb_iterator *iter);
 
+bool
+splinterdb_validate_key_in_range(const splinterdb *kvs, slice key);
+
 #endif // _SPLINTERDB_H_
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -292,7 +292,4 @@ splinterdb_iterator_get_current(splinterdb_iterator *iter, // IN
 int
 splinterdb_iterator_status(const splinterdb_iterator *iter);
 
-bool
-splinterdb_validate_key_in_range(const splinterdb *kvs, slice key);
-
 #endif // _SPLINTERDB_H_
--- a/src/splinterdb.c
+++ b/src/splinterdb.c
@@ -632,6 +632,43 @@ validate_key_length(const splinterdb *kvs, uint64 key_length)
 }
 
 /*
+ * Validate that a key being inserted is within [min, max]-key range.
+ */
+bool
+splinterdb_validate_key_in_range(const splinterdb *kvs, slice key)
+{
+   const data_config *cfg = kvs->shim_data_cfg.app_data_cfg;
+
+   int cmp_rv = 0;
+
+   // key to-be-inserted should be >= min-key
+   cmp_rv = cfg->key_compare(
+      cfg, slice_create(cfg->min_key_length, cfg->min_key), key);
+   if (cmp_rv > 0) {
+      platform_error_log("Key '%.*s' is less than configured min-key '%.*s'.\n",
+                         (int)slice_length(key),
+                         (char *)slice_data(key),
+                         (int)cfg->min_key_length,
+                         cfg->min_key);
+      return FALSE;
+   }
+
+   // key to-be-inserted should be <= max-key
+   cmp_rv = cfg->key_compare(
+      cfg, key, slice_create(cfg->max_key_length, cfg->max_key));
+   if (cmp_rv > 0) {
+      platform_error_log(
+         "Key '%.*s' is greater than configured max-key '%.*s'.\n",
+         (int)slice_length(key),
+         (char *)slice_data(key),
+         (int)cfg->max_key_length,
+         cfg->max_key);
+      return FALSE;
+   }
+   return TRUE;
+}
+
+/*
  *-----------------------------------------------------------------------------
  * splinterdb_insert_raw_message --
  *
@@ -655,6 +692,9 @@ splinterdb_insert_message(const splinterdb *kvs, // IN
    if (rc != 0) {
       return rc;
    }
+
+   debug_assert(splinterdb_validate_key_in_range(kvs, key),
+                "Attempt to insert key outside configured min/max key-range");
 
    char key_buffer[MAX_KEY_SIZE] = {0};
    rc = encode_key(sizeof(key_buffer), key_buffer, key);
@@ -812,15 +852,20 @@ splinterdb_iterator_init(const splinterdb     *kvs,      // IN
    trunk_range_iterator *range_itor = &(it->sri);
 
    char start_key_buffer[MAX_KEY_SIZE] = {0};
-   if (!slice_is_null(start_key)) {
+   bool start_key_is_null              = slice_is_null(start_key);
+   if (!start_key_is_null) {
       int rc = encode_key(MAX_KEY_SIZE, start_key_buffer, start_key);
       if (rc != 0) {
          return rc;
       }
    }
 
-   platform_status rc = trunk_range_iterator_init(
-      kvs->spl, range_itor, start_key_buffer, NULL, UINT64_MAX);
+   platform_status rc =
+      trunk_range_iterator_init(kvs->spl,
+                                range_itor,
+                                (start_key_is_null ? NULL : start_key_buffer),
+                                NULL,
+                                UINT64_MAX);
    if (!SUCCESS(rc)) {
       trunk_range_iterator_deinit(range_itor);
       platform_free(kvs->spl->heap_id, *iter);
--- a/src/splinterdb.c
+++ b/src/splinterdb.c
@@ -17,6 +17,7 @@
 
 #include "clockcache.h"
 #include "splinterdb/splinterdb.h"
+#include "splinterdb_private.h"
 #include "rc_allocator.h"
 #include "trunk.h"
 #include "btree_private.h"
--- a/src/splinterdb.c
+++ b/src/splinterdb.c
@@ -635,7 +635,7 @@ validate_key_length(const splinterdb *kvs, uint64 key_length)
  * Validate that a key being inserted is within [min, max]-key range.
  */
 bool
-splinterdb_validate_key_in_range(const splinterdb *kvs, slice key)
+validate_key_in_range(const splinterdb *kvs, slice key)
 {
    const data_config *cfg = kvs->shim_data_cfg.app_data_cfg;
 
@@ -645,11 +645,10 @@ splinterdb_validate_key_in_range(const splinterdb *kvs, slice key)
    cmp_rv = cfg->key_compare(
       cfg, slice_create(cfg->min_key_length, cfg->min_key), key);
    if (cmp_rv > 0) {
-      platform_error_log("Key '%.*s' is less than configured min-key '%.*s'.\n",
-                         (int)slice_length(key),
-                         (char *)slice_data(key),
-                         (int)cfg->min_key_length,
-                         cfg->min_key);
+      platform_error_log(
+         "Key '%s' is less than configured min-key '%s'.\n",
+         key_string(cfg, key),
+         key_string(cfg, slice_create(cfg->min_key_length, cfg->min_key)));
       return FALSE;
    }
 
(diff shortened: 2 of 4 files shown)


include/splinterdb/splinterdb.h
```diff
@@ -292,4 +292,7 @@ splinterdb_iterator_get_current(splinterdb_iterator *iter, // IN
 int
 splinterdb_iterator_status(const splinterdb_iterator *iter);
 
+bool
+splinterdb_validate_key_in_range(const splinterdb *kvs, slice key);
```
gapisback, line 296: Add this as extern interface. I used this to write a small unit-test just to sanitize my own code and verify that this validation, indeed, works.

Now, this interface will also be available publicly to end-users who may wish to validate their keys before inserts (given that Splinter is not doing it in optimized builds).

src/splinterdb.c
```diff
@@ -656,6 +693,9 @@ splinterdb_insert_message(const splinterdb *kvs, // IN
       return rc;
    }
 
+   debug_assert(splinterdb_validate_key_in_range(kvs, key),
```
gapisback, line 696: The validation only happens in debug builds.

src/splinterdb.c
```diff
@@ -812,15 +852,20 @@ splinterdb_iterator_init(const splinterdb     *kvs,      // IN
...
          return rc;
       }
    }
 
-   platform_status rc = trunk_range_iterator_init(
-      kvs->spl, range_itor, start_key_buffer, NULL, UINT64_MAX);
+   platform_status rc =
+      trunk_range_iterator_init(kvs->spl,
+                                range_itor,
+                                (start_key_is_null ? NULL : start_key_buffer),
```
gapisback, line 866: This is the real fix.

Btw, I simply could not reproduce this problem or show any symptoms of its effect in unit-tests when run in optimized build modes.

On L 854 above, `start_key_buffer[]` would be null-terminated at byte-0, so nothing offensive seems to happen in opt-builds. The bug could only be shown to manifest in debug build modes.

That's not very comforting, as now we don't know how bad the exposure might have been prior to this bug fix.

tests/unit/splinterdb_quick_test.c
```diff
@@ -437,6 +437,8 @@ CTEST2(splinterdb_quick, test_basic_iterator)
       ASSERT_EQUAL(0, rc);
       i++;
    }
+   ASSERT_EQUAL(num_inserts, i);
```
gapisback, line 440: Really ... unrelated to this fix. But while in this file, decided to augment some checks verifying that we are returning right # of kv-pairs.

tests/unit/splinterdb_quick_test.c
```diff
@@ -807,6 +809,108 @@ CTEST2(splinterdb_quick, test_iterator_custom_comparator)
    }
 }
 
+/*
+ * Exercise the utility function which validates that the key being inserted
+ * is within [min, max] range. That check was added upon discovering that
+ * we would allow invalid keys to be inserted which lead to downstream
+ * errors while exercising the iteration interfaces.
+ */
+CTEST2(splinterdb_quick, test_validate_key_in_range)
```
gapisback, line 818: minor test case to exercise newly added validation function.

tests/unit/splinterdb_quick_test.c
```diff
@@ -807,6 +809,108 @@ CTEST2(splinterdb_quick, test_iterator_custom_comparator)
...
+      data->kvsb, slice_create(strlen(test_key), test_key));
+   ASSERT_FALSE(is_valid);
+}
+
+/*
+ * Test case to verify that iterator interfaces work correctly.
+ * Prior to fix for issue #419, this test case would fail with an assertion
+ * (that is activated only) in debug mode runs.
+ */
+CTEST2(splinterdb_quick, test_iterator_init_bug)
```
gapisback, line 867: Distilled this case down to exactly the min configuration needed to repro the bug. The key line is L876 below.

gapisback:
Hi, @rtjohnso -- This PR peels out the bug-fix we investigated this afternoon.

Finalizing this PR and building the unit-test case took slightly bit longer than what I had expected.
If you can give this a quick look, I can move this forward, followed by the C-sample programs which is also ready to go.
(That other PR #399 had some buggy code that this rectifies, which will now allow me to clean-up that change set.)

include/splinterdb/splinterdb.h
rtjohnso: This should not be public.

Making this public is just setting a trap for users when we get rid of the min/max key stuff.  When that happens, this API won't even make sense any more.

rtjohnso: You can test it w/o putting it in `include`, which is the real public API.

src/splinterdb.c
```diff
@@ -631,6 +631,43 @@ validate_key_length(const splinterdb *kvs, uint64 key_length)
...
+
+   int cmp_rv = 0;
+
+   // key to-be-inserted should be >= min-key
+   cmp_rv = cfg->key_compare(
+      cfg, slice_create(cfg->min_key_length, cfg->min_key), key);
+   if (cmp_rv > 0) {
+      platform_error_log("Key '%.*s' is less than configured min-key '%.*s'.\n",
+                         (int)slice_length(key),
+                         (char *)slice_data(key),
```
rtjohnso, line 650: Keys are not C strings.  

Use the data_config API for printing keys.

rtjohnso: Even better, use the data_internal.h API for printing the keys, i.e.
```
platform_error_log("... %s ... ",
                               ...,
                               key_string(data_cfg, key),
                               ...);
```
where `key` is a slice.

rtjohnso (changes requested):

include/splinterdb/splinterdb.h
gapisback: Reworked as suggested. Renamed routine to simply `validate_key_in_range()` and removed interface from public .h file. Instead, did a slightly sloppy thing of listing the prototype in the single test-case file that needs it. (This works, and good enough for now.)

src/splinterdb.c
gapisback: @rtjohnso - Reworked as suggested.

(I'm not entirely sure that printing keys in hex-dump will aid debugging. Most common keys would likely be strings, so '%.*s' was quite helpful. Anyway -- we can re-visit this later on.)

Kindly re-review the delta-diff commit uploaded.

include/splinterdb/splinterdb.h
rtjohnso: How about creating `src/splinterdb_private.h` and put it in there?

Then have `src/splinterdb_private.h` include `include/splinterdb/splinterdb.h` and have `src/splinterdb.c` include `src/splinterdb_private.h`, similar to how things are done with the btree code.

The test would also include `src/splinterdb_private.h`

src/splinterdb.c
rtjohnso: I don't think this is a matter of taste.  It is a matter of not crashing the program.

Keys are not C strings.  They are not required to be null terminated or even contain any null bytes at all.  Passing them to a printf-style function as a "%s" is a straight-up SEGFAULT waiting to happen.

Furthermore, the appliction gets to define how keys are printed.  If they don't want hex dumps, they can define a custom function for how to print keys.

tests/unit/splinterdb_quick_test.c
```diff
@@ -67,6 +67,10 @@ check_current_tuple(splinterdb_iterator *it, const int expected_i);
 static int
 custom_key_comparator(const data_config *cfg, slice key1, slice key2);
 
+// Extern prototype defined in the library.
+bool
+validate_key_in_range(const splinterdb *kvs, slice key);
```
rtjohnso, line 72: Create `src/splinterdb_private.h` and declare this there, as described in my earlier comment.

rtjohnso (changes requested):

src/splinterdb.c
gapisback: OK, point noted & agreed. Thanks for catching this sloppy code.

include/splinterdb/splinterdb.h
gapisback: Hmmm ... @rtjohnso , I know there is a precedence here with `btree.h` but in this case, the `private.h` will initially have just this one extern proto.

I feel it's an over-kill to engineer it this way just to satisfy the one (and probably ever only) consumer of this extern prototype which is the test case. 

On a quick look, I found that somewhere ~10 files are currently #including `splinterdb.h` .  All those would need to be changed to now #include `splinterdb_private.h` -- which will [needlessly] bloat this change-set.

I am not entirely convinced that this is the preferred way, especially given that all this is for just one extern prototype visibility.

tests/unit/splinterdb_quick_test.c
gapisback: @rtjohnso - Not sure this is the preferred way. See my previous response. I feel it's an over-kill refactoring in this case.

rtjohnso:
The problem with simply declaring the function in the test is that the compiler will not check the types between the call site and the function definition.  If we ever change the function signature then the code will still compile and link, but we will get very weird and hard to diagnose failures in that test.

Either delete the test or expose the API internally so the test can see it and the compiler can check that the function definition and the call site are consistent.

Also, you do not need to change all the includes of `splinterdb.h` to be `splinterdb_private.h`.  Only the places that need the private definitions will need to change.

rtjohnso:
You could have implemented my suggestion in less time than it took to write your email.

Having declarations that are not checked by the compiler is a hard no.

Furthermore, the `splinterdb_private.h` file can grow to be useful for other unit tests down the road, similar to the `btree_private.h` file.

rtjohnso:
> At best what I am willing to consider is to add splinterdb_private.h and only list the extern prototype. And include that just in the test case file and in splinterdb.c. No nested #includes in this case as [eventually] this prototype definition will likely go away anyway, thereby, making this splinterdb_private.h file obsolete.

That is exactly what I am suggesting.  The only extra thing that I suggested is that `splinterdb_private.h` include `splinterdb.h`

gapisback:
@rtjohnso - I've reworked the change set per this commit [Address review comment: Introduce splinterdb_private.h to carry exter…](https://github.com/vmware/splinterdb/pull/421/commits/be1ec1a8daa74f6a032af800060c671bdd3b552e)

Kindly take a look.

This should address your concern of having a compiler-safe prototype definition of `validate_key_in_range` across the src and test files.

src/splinterdb.c
```diff
@@ -17,6 +17,7 @@
 
 #include "clockcache.h"
 #include "splinterdb/splinterdb.h"
```
rtjohnso, line 19: You should be able to delete the `#include "splinterdb/splinterdb.h"` once you make `splinterdb_private.h` self-contained.

tests/unit/splinterdb_quick_test.c
```diff
@@ -32,6 +32,7 @@
 #include "splinterdb/public_platform.h"
 #include "splinterdb/default_data_config.h"
 #include "splinterdb/splinterdb.h"
```
rtjohnso, line 34: Ditto on deleting `#include "splinterdb/splinterdb.h"`

rtjohnso (changes requested):
Almost there.

src/splinterdb_private.h
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * This file contains the private interfaces implemented in splinterdb.c.
+ * These definitions are provided here so that they can be shared by the
+ * source and test modules.
+ */
+#ifndef __SPLINTERDB_PRIVATE_H__
+#define __SPLINTERDB_PRIVATE_H__
+
+bool
+validate_key_in_range(const splinterdb *kvs, slice key);
```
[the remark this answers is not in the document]
gapisback: @rtjohnso -- I don't quite agree with this approach. I think it's better if the test code only includes what it needs.

In any case -- we are on opposite sides of an argument. And I want to get this code merge-in.

So, Done.

src/splinterdb.c
gapisback: I know I _should_ be able to, but I explicitly chose to not do it so as I don't think it's the better approach.

Anyway, to reiterate my previous point -- I need this code integrated to /main. And does not seem like I convince you of my reasons.

So, Done.

gapisback:
@rtjohnso -- Pushed one more, hopefully final, commit. Kindly take a relook.

src/splinterdb_private.h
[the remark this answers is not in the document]
rtjohnso: What is the rationale for having a header file that causes a compile error if you don't include another header file first?

src/splinterdb.c
rtjohnso: > On L 854 above, start_key_buffer[] would be null-terminated at byte-0, so nothing offensive seems to happen in opt-builds.

Keys are not C strings.  Keys are arbitrary sequences of bytes.  Their structure is application defined.

rtjohnso (approved):
I left a couple of extra comments adding to the discussion, but there's nothing that needs to be changed in this PR.  Ship it!

PR #431, opened 2022-08-06
Refactor Trunk Node Handles

ajhconway:
Refactors handling of trunk nodes to a new trunk_node struct, rather than using page_handles. This is now analagous to the handing of btree_nodes.

Moves the related low-level trunk node functionality to a new header src/trunk_node.h. This header will eventually hold all low-level trunk node functionality.

Moves the trunk node logging functionality to a new header src/trunk_log.h.

Moves the superblock functionality to a new header src/superblock.c. This should simplify the eventual change to calling these functions from src/splinterdb.c.

Fixes a use-after-free bug where `trunk_node_verify` is called on the child in a debug_assert after a trunk_split_leaf, which ungets the child.

### DIFF
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -629,7 +629,7 @@ trunk_node_unlock(cache *cc, trunk_node *node)
 static inline void
 trunk_alloc(cache *cc, mini_allocator *mini, uint64 height, trunk_node *node)
 {
-   node->addr = mini_alloc(mini, height, NULL_SLICE, NULL);
+   node->addr = mini_alloc(mini, height, NULL_KEY, NULL);
    debug_assert(node->addr != 0);
    node->page = cache_alloc(cc, node->addr, PAGE_TYPE_TRUNK);
    node->hdr  = (trunk_hdr *)(node->page->data);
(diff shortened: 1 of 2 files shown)


src/superblock.h
```diff
@@ -0,0 +1,113 @@
...
+   uint64      log_addr;
+   uint64      log_meta_addr;
+   uint64      timestamp;
+   bool        checkpointed;
+   bool        unmounted;
+   checksum128 checksum;
+} trunk_super_block;
+
+void
+trunk_set_super_block(trunk_handle *spl,
```
rtjohnso, line 35: I guess eliminate the `trunk_` prefix on these, if superblocks are going to become their own entity now?

src/trunk_log.h
```diff
@@ -0,0 +1,84 @@
+// Copyright 2018-2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * trunk_log.h --
```
rtjohnso, line 5: Maybe `debug` instead of `log`, since `log` is overloaded in this context?

rtjohnso (approved):
This is great!

But I assume `trunk_node.h` is still in progress?

Like, will all low-level code that basically handles marshaling / unmarshaling if the physical representation eventually go into `trunk_node.[hc]`?

gapisback:
Hi @ajhconway - I didn't realize that you had marked me as a reviewer. Will try to take a look at this later this week, after I complete some of the insert-perf benchmarking work that's due this week.

ajhconway:
> This is great!
> 
> But I assume `trunk_node.h` is still in progress?
> 
> Like, will all low-level code that basically handles marshaling / unmarshaling if the physical representation eventually go into `trunk_node.[hc]`?

I've decided to back these changes out for another PR. They were a little too incomplete in this one.

rtjohnso (approved):
No notes.

rosenhouse (commented):
Thanks for giving me the opportunity to review, but I don't have the bandwidth to develop the necessary context to review.  My only comment is that it's nice you didn't have to change any tests at all to keep everything passing -- a pure refactor like this is uncommon.

PR #432, opened 2022-08-10
clarify lifetimes in code comments

rosenhouse:
Header comments should explain to callers how long certain objects need to live.
- The user-provided `data_config` needs to live at least as long as the `splinterdb`
- The `splinterdb` needs to live at least as long as any `lookup_result` objects used with it


rtjohnso (approved):
Thank you!

PR #433, opened 2022-08-11
simplify signature of lookup_result_value

rosenhouse:
We don't need to pass the splinterdb instance in again on `splinterdb_lookup_result_value()`.

In the future, it could be stashed in the `lookup_result` when the `lookup_result` is first `init()`ed, and it could be checked at `lookup()` time.  But the `_value()` function is just an accessor.


rtjohnso (approved):
Thank you!

PR #434, opened 2022-08-15
fix alignment of lookup_result

rosenhouse:
The application-facing, opaque `splinterdb_lookup_result` and its internal representation `_splinterdb_lookup_result` should have the same alignment.


rtjohnso (approved):
Good catch!

PR #435, opened 2022-08-21
fix several page_type errors

rtjohnso:
Also change the test-results Makefile target to behave more like a normal target


rtjohnso:
Alex can you sanity check these page_types.  They looked obviously wrong, which makes me worry that I'm missing something.

ajhconway (approved):
All obviously good fixes.

PR #436, opened 2022-08-24
eliminate vestigial task handle from cache

rtjohnso:
`clockcache_init` took a handle on the task system but clockcache never used it anywhere.  Removing it makes it easier to instantiate the cache.


ajhconway (approved):
LGTM

PR #439, opened 2022-09-05
eliminate overwrites in shard_log

rtjohnso:
This patch eliminates per-page metadata that gets overwritten each time an entry it added to the log.  Specifically, it eliminates the num_entries field in shard_log_hdr.  The reason to do this is to make the log essentially append-only, which is important when the log is stored in pmem.

This patch consequently must change how the end of the log is detected.  Each log entry now has a checksum.  The log ends when we reach a malformed log entry, or an entry with a bad checksum, or we reach the end of a log extent and there is no next_extent.

The downside of this patch is that log entries are larger, due to the checksum, which might have a modest impact on throughput.  A quick test (1 run of the old cold and 1 run of the new code) using 
```driver_test splinter_test --perf --num-insert-threads 1 --log --num-inserts 50000000```
shows 1571 nsec/insert in the new code vs 1506 nsec/insert in the old code, so about a 0.5% slowdown.

### DIFF
--- a/src/shard_log.c
+++ b/src/shard_log.c
@@ -63,13 +63,6 @@ shard_log_pages_per_extent(shard_log_config *cfg)
    return cache_config_pages_per_extent(cfg->cache_cfg);
 }
 
-static inline checksum128
-shard_log_checksum(shard_log_config *cfg, page_handle *page)
-{
-   return platform_checksum128(
-      page->data + 16, shard_log_page_size(cfg) - 16, cfg->seed);
-}
-
 static inline shard_log_thread_data *
 shard_log_get_thread_data(shard_log *log, threadid thr_id)
 {
@@ -141,11 +134,12 @@ shard_log_zap(shard_log *log)
  * -------------------------------------------------------------------------
  */
 struct ONDISK log_entry {
-   uint64 generation;
-   uint16 keylen;
-   uint16 messagelen;
-   uint8  msg_type;
-   char   contents[];
+   checksum128 checksum;
+   uint64      generation;
+   uint16      keylen;
+   uint16      messagelen;
+   uint8       msg_type;
+   char        contents[];
 };
 
 static char *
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/shard_log.c [hunk 1]
```diff
@@ -362,17 +396,26 @@ shard_log_iterator_init(cache              *cc,
...
+         if (hdr->magic == magic) {
+            uint64 page_entry_count;
+            page_entry_count = num_log_entries(cfg, hdr, &page_terminal);
+            itor->num_entries += page_entry_count;
+            if (page_entry_count) {
+               num_valid_pages++;
+               next_extent_addr = shard_log_next_extent_addr(cfg, page);
+            }
+         } else {
+            page_terminal = TRUE;
```
ajhconway, line 418: Relying on `hdr->magic` here seems like a problem. What if a log extent is deallocated and then reallocated as the next extent, but no writes have been issued to it yet. Then the magic will still match and all the checksums will work too.

Another issue is what happens with a torn writes that either occur at entry boundaries or which validate part of the header? It feels like we should still have a page-wide checksum.

src/shard_log.c
```diff
@@ -386,24 +429,35 @@ shard_log_iterator_init(cache              *cc,
...
                itor->entries[entry_idx] = cursor;
                entry_idx++;
                cursor = log_entry_next(cursor);
             }
-            next_extent_addr = shard_log_next_extent_addr(cfg, page);
+            if (valid_log_entry(cfg, hdr, le)) {
+               next_extent_addr = shard_log_next_extent_addr(cfg, page);
+            } else {
+               page_terminal = TRUE;
+            }
```
ajhconway, line 458: Why is the logic different here than for the first extent?

ajhconway (commented):
I'm not sure about these changes, and I'd like to understand the motivation better.

I'd like to see log performance benchmarks to better understand what we're giving up.

Are these temporary log changes en route to a log rewrite?

rtjohnso:
Probably going to do much more substantial rewrite of log, so closing this.

Issue #444, opened 2022-09-28
Refactor and modularize inline calls in test.sh that execute tests into functions (for future code re-use)


gapisback:
### Is your feature request related to a problem? Please describe.

In `test.sh` script, there are several executions of tests in the main() portion of the script.
We should refactor those into smaller functions, like it's been done for other functions such as `nightly_test_limitations()`, `nightly_cache_perf_tests()` etc.

This sets up the script to re-deploy these tests with upcoming support for running SplinterDB using shared memory. An upcoming feature request will add support for `--use-shmem` argument to many / most tests. 

By refactoring the existing test invocations into sub-functions, it will become simpler to add-in calls to run the same set of tests using shared-memory support.

### Describe the solution you'd like

See above.

### Describe alternatives you've considered

We could duplicate calls to execute tests but that will be a code maintenance issue.

PR #445, opened 2022-09-28
(#444) Refactor chunks of test.sh into functions for code-reuse

gapisback:
The script test.sh has several inline calls to run different tests. Many of these are interesting cases to be run with an upcoming feature to run Splinter using allocated shared memory segments.

This commit rearranges and refactors inline calls to run diff tests into new functions that will now execute a collection of related tests.

Examples of new functions added are:
 - run_fast_unit_tests(), run_slower_unit_tests()
 - run_splinter_functionality_test(), run_splinter_perf_tests()

and so on. No changes are made to the set of tests that will be executed. This refactoring lays down the stage to re-execute relevant set(s) of tests using shared-memory.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -215,7 +215,9 @@ function nightly_functionality_stress_tests() {
     rm ${dbname}
 }
 
+# #############################################################################
 # Run through collection of nightly stress tests
+# #############################################################################
 function run_nightly_stress_tests() {
 
     nightly_functionality_stress_tests
@@ -272,6 +274,7 @@ function nightly_sync_perf_tests() {
 
 # #############################################################################
 # Nightly Cache Performance tests with async disabled
+# #############################################################################
 function nightly_cache_perf_tests() {
 
     local dbname="cache_test.perf.db"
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -510,6 +517,119 @@ function test_example_programs() {
     "$BINDIR"/examples/splinterdb_custom_ipv4_addr_sortcmp_example
 }
 
+# ##################################################################
+# Smoke Tests: Run a small collection of fast-running unit-tests
+# ##################################################################
+function run_fast_unit_tests() {
```
gapisback, line 523: Most of this change is to do stuff like this ... refactoring inline calls to new functions.

gapisback:
Hi @rosenhouse -- I know you have a short week but am hoping you'll be able to spend a little bit of time on Thu when you are back to review this PR.

It's a minor refactoring of `test.sh` to setup the stage to online test execution for upcoming support for shared memory in Splinter.

There is no new code, just refactoring. I could just as well push but thought you'd like to have a crack at  it.

rosenhouse (approved):
seems ok to me

Issue #448, opened 2022-09-29
Improve comments and code in task.c; Build unit-tests to exercise core interfaces of task system.


gapisback:
### Is your feature request related to a problem? Please describe.

Not exactly a problem, but some code improvements that were done while studying the task system code. These merit being integrated to /main. Actual changes are described below. This work was done while studying the design implications of supporting Splinter to run using shared memory for memory allocation (v/s current scheme of malloc()/free()). 

The task system also dips into this memory allocation, so this code cleanup and unit-testing is a useful exerciser to have in-place which comes in handy when we will bring-in support for shared segment memory management.

### Describe the solution you'd like

Few things that will be done:

- Code, comment cleanup in `task.c`. No new functionality is added
- Develop a C-Unit test to exercise core task system interfaces. Some cases have been added, and more can be developed / extended as part of review feedback.

PR #449, opened 2022-09-29
(#448) Clarify logic in task.c; Add unit-test to validate public interfaces.

gapisback:
This commit adds a bit of code cleanup to the task system module, task.c . No new functionality or change-in-logic is being done. Add new unit-test, task_system_test, to validate internal state of the task system for simple task creation / execution.

---
NOTE: to the reviewer. This change set came out of my code walk through to understand the task system and its innards.
This was done in context of the design rework to support execution of Splinter using allocated shared memory segments.
While at it, I constructed a small C-unit test to work through task system interfaces and to confirm, via tests, expected outputs.

Some of this testing will also be helpful when we "cut-over" to running tests using shared segments.

### DIFF
--- a/src/platform_linux/platform.c
+++ b/src/platform_linux/platform.c
@@ -101,6 +101,9 @@ platform_buffer_destroy(buffer_handle *bh)
    return STATUS_OK;
 }
 
+/*
+ * platform_thread_create() - External interface to create a Splinter thread.
+ */
 platform_status
 platform_thread_create(platform_thread       *thread,
                        bool                   detached,
@@ -134,6 +137,12 @@ platform_thread_join(platform_thread thread)
    ret = pthread_join(thread, &retval);
 
    return CONST_STATUS(ret);
+}
+
+platform_thread
+platform_thread_id_self()
+{
+   return pthread_self();
 }
 
 platform_status
--- a/src/task.c
+++ b/src/task.c
@@ -6,27 +6,31 @@
 
 #include "poison.h"
 
-#define MAX_HOOKS (2048)
+#define MAX_HOOKS (8)
 
 int              hook_init_done = 0;
 static int       num_hooks      = 0;
 static task_hook hooks[MAX_HOOKS];
 
 // forward declarations that aren't part of the public API of the task system
-platform_status
+static platform_status
 task_register_hook(task_hook newhook);
 
-threadid *
+static threadid *
 task_system_get_max_tid(task_system *ts);
 
-uint64 *
+static uint64 *
 task_system_get_tid_bitmask(task_system *ts);
 
-void
+static void
 task_system_io_register_thread(task_system *ts);
 // end forward declarations
 
-void
+/*
+ * task_init_tid_bitmask() - Initialize the global bitmask of active threads in
+ * the task system structure to indicate that no threads are currently active.
+ */
+static void
 task_init_tid_bitmask(uint64 *tid_bitmask)
 {
    // We use a 64 bit word to act as the thread bitmap.
@@ -40,8 +44,17 @@ task_init_tid_bitmask(uint64 *tid_bitmask)
    }
 }
 
+/*
+ * task_init_threadid() - Initialize a new thread's thread ID (index).
+
+ * The first thing a newly started thread that wants to use Splinter is expected
+ * to do is to "register itself" with the task system. This function is
+ * defined as a thread's hook and gets run as part of the thread's
+ * registration. This function initializes each thread's thread-ID, which
+ * is really an index into the array of thread-specific data structures.
+ */
 static void
-init_threadid(task_system *ts)
+task_init_threadid(task_system *ts)
 {
    threadid  tid         = INVALID_TID;
    uint64   *tid_bitmask = task_system_get_tid_bitmask(ts);
@@ -68,9 +81,15 @@ init_threadid(task_system *ts)
 
 out:
    debug_assert(tid != INVALID_TID);
+
+   // Sets thread-ID, for example tracked as thread-local storage.
    platform_set_tid(tid);
 }
 
+/*
+ * Return the max thread-index across all active tasks.
+ * Mainly intended as a testing hook.
+ */
 threadid
 task_get_max_tid(task_system *ts)
 {
@@ -88,17 +107,17 @@ task_get_max_tid(task_system *ts)
  * beginning of the main thread that uses the task, similar to how
  * __attribute__((constructor)) works.
  */
-void
+static void
 register_init_tid_hook(void)
 {
    // hooks need to be initialized only once.
    if (__sync_fetch_and_add(&hook_init_done, 1) == 0) {
-      task_register_hook(init_threadid);
+      task_register_hook(task_init_threadid);
       task_register_hook(task_system_io_register_thread);
    }
 }
 
-void
+static void
 task_run_thread_hooks(task_system *ts)
 {
    for (int i = 0; i < num_hooks; i++) {
@@ -106,22 +125,38 @@ task_run_thread_hooks(task_system *ts)
    }
 }
 
-void
+/*
+ * De-registering a task frees up the thread's index so that it can be re-used.
+ */
+static void
 task_clear_threadid(task_system *ts, threadid tid)
 {
    uint64 *tid_bitmask = task_system_get_tid_bitmask(ts);
-   platform_assert(!(*tid_bitmask & (1ULL << tid)));
+
+   uint64 bitmask_val = *tid_bitmask;
+
+   // Ensure that caller is only clearing for a thread that's in-use.
+   platform_assert(!(bitmask_val & (1ULL << tid)),
+                   "Thread [%lu] is expected to be in-use. Bitmap: 0x%lx",
+                   tid,
+                   bitmask_val);
+
    // set bit back to 1 to indicate a free slot.
    while (1) {
       uint64 tmp_bitmask = *tid_bitmask;
       uint64 new_value   = tmp_bitmask | (1ULL << tid);
       if (__sync_bool_compare_and_swap(tid_bitmask, tmp_bitmask, new_value)) {
+
+         // This is the reverse of registration, where tid is allocated by
+         // init(). So, clear this out as thread is being de-registered.
+         // (Useful for testing and used in other assertions.)
+         platform_set_tid(0);
          return;
       }
    }
 }
 
-platform_status
+static platform_status
 task_register_hook(task_hook newhook)
 {
    int my_hook_idx = __sync_fetch_and_add(&num_hooks, 1);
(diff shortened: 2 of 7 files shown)


src/task.c
```diff
@@ -10,23 +10,29 @@
 
 int              hook_init_done = 0;
 static int       num_hooks      = 0;
-static task_hook hooks[MAX_HOOKS];
+// RESOLVE: At 8 bytes per task_hook{}, we are burning up 16K of memory. Only 2 hooks
```
gapisback, line 13: @ajhconway - We should discuss and reconsider this. 

When we, eventually, move to a shared memory implementation, I suspect (expect) that this array will need to be allocated from shared memory segment. That will cost us 16K, whereas something like 64 bytes should suffice if we reduce MAX_HOOKS to 8.

src/task.c
```diff
@@ -10,23 +10,29 @@
...
+// are currently being connected. What about reducing MAX_HOOKS to, say, 8?
+static task_hook hooks[MAX_HOOKS];	
 
 // forward declarations that aren't part of the public API of the task system
-platform_status
+static platform_status
```
gapisback, line 18: Many changes in this file add `static` to functions / prototypes to document clearly that this is a static function.
Improves code readability.

src/task.c
```diff
@@ -83,12 +104,21 @@ task_get_max_tid(task_system *ts)
    return *max_tid + 1;
 }
 
+/*
+ * Return the bitmasks of tasks active. Mainly intended as a testing hook.
+ */
+uint64
+task_active_tasks_mask(task_system *ts)
```
gapisback, line 111: Added this extern accessor function, used in unit-tests for validating bitmask.

src/task.c
```diff
@@ -149,7 +196,7 @@ task_register_this_thread(task_system *ts, uint64 scratch_size)
 void
 task_deregister_this_thread(task_system *ts)
 {
-   uint64 tid = platform_get_tid();
+   threadid tid = platform_get_tid();
```
gapisback, line 199: Use the right typedef.

src/task.c
```diff
@@ -168,6 +215,17 @@ typedef struct {
...
 
+/*
+ * -----------------------------------------------------------------------------
+ * task_invoke_with_hooks() -
+ *
+ * A thread has been created with a call-back function to execute. This
+ * function invokes that function bracketted by calls to register / deregister
+ * the thread with Splinter's task system.
+ *
+ * RESOLVE: I see references to this, ... but who calls this?
```
gapisback, line 226: Let's discuss .... this, and then there seem to be a few more functions, that are written but never called.

Is this vestigial code that we could / should get rid of now, to reduce confusion ... [???].

src/task.c
```diff
@@ -185,6 +243,11 @@ task_invoke_with_hooks(void *func_and_args)
    platform_free(thread_to_start->heap_id, func_and_args);
 }
 
+/*
+ * -----------------------------------------------------------------------------
+ * task_create_thread_with_hooks() - Comment ?
```
gapisback, line 248: Help! Can you help me add a meaningful prolog?

src/task.c
```diff
@@ -215,6 +278,15 @@ task_create_thread_with_hooks(platform_thread       *thread,
    return ret;
 }
 
+/*
+ * -----------------------------------------------------------------------------
+ * task_thread_create() - Helper method to create a new task and to do the
+ *  required registration with Splinter.
+ *
+ * Currently, this is active mainly in tests. It's also used to create
+ * background tasks, a feature that is currently not enabled.
```
gapisback, line 287: @ajhconway - RESOLVE: What's the history on background threads?

Can I elicit some documentation from you and put it down here, so future readers can benefit?

src/task.c
```diff
@@ -8,25 +8,32 @@
 
 #define MAX_HOOKS (2048)
 
-int              hook_init_done = 0;
-static int       num_hooks      = 0;
+int        hook_init_done = 0;
+static int num_hooks      = 0;
+// RESOLVE: At 8 bytes per task_hook{}, we are burning up 16K of memory. Only 2
+// hooks are currently being connected. What about reducing MAX_HOOKS to, say,
+// 8?
```
ajhconway, line 15: This does feel a little much. Maybe something still hard to imagine blowing out like 64?

src/task.c
```diff
@@ -68,9 +84,15 @@ init_threadid(task_system *ts)
 
 out:
    debug_assert(tid != INVALID_TID);
+
+   // Sets thread-ID, tracked as thread-local storage.
```
ajhconway, line 88: This is only true on Linux. It could be implemented differently on another platform.

I would just rephrase this as "Sets thread-ID, for example tracked as thread-local storage."

src/task.c
```diff
@@ -98,19 +129,30 @@ register_init_tid_hook(void)
...
    uint64 *tid_bitmask = task_system_get_tid_bitmask(ts);
-   platform_assert(!(*tid_bitmask & (1ULL << tid)));
+
+   uint64 bitmask_val = *tid_bitmask;
+
+   // Ensure that caller is only clearing for a thread that's in-use.
+   platform_assert(!(bitmask_val & (1ULL << tid)),
+                   "Thread [%lu] is expected to be in-use. Bitmap: 0x%lx",
+                   tid,
+                   bitmask_val);
```
ajhconway, line 154: This is good

tests/unit/task_system_test.c
```diff
@@ -0,0 +1,349 @@
...
+   threadid main_thread_idx = platform_get_tid();
+   ASSERT_EQUAL(main_thread_idx, 0, "main_thread_idx=%lu", main_thread_idx);
+
+   // Setup thread-specific struct, needed for validation in thread's worker fn
+   thread_cfg.tasks                  = data->tasks;
+   thread_cfg.exp_thread_idx         = 1; // Main thread is at index 0.
+   thread_cfg.active_threads_bitmask = task_active_tasks_mask(data->tasks);
+
+   int rc = pthread_create(&new_thread, NULL, exec_one_thread, &thread_cfg);
+   ASSERT_EQUAL(0, rc);
```
ajhconway, line 164: This should be using `platform_thread` and `task_thread_create`.

tests/unit/task_system_test.c
```diff
@@ -0,0 +1,349 @@
...
+      // These are independent of the new thread's creation.
+      thread_cfgp->tasks          = data->tasks;
+      thread_cfgp->exp_thread_idx = tctr;
+
+      rc =
+         pthread_create(&new_thread, NULL, exec_one_of_n_threads, thread_cfgp);
+      ASSERT_EQUAL(0, rc);
+
+      thread_cfgp->this_thread_id = new_thread;
+   }
```
ajhconway, line 216: Again this should be using `platform_thread` and `task_thread_create`.

tests/unit/task_system_test.c
```diff
@@ -0,0 +1,349 @@
...
+   // Verify that the state of active-threads bitmask (managed by Splinter) has
+   // not changed just by creating this pthread. It should be the same as what
+   // we had recorded just prior to pthread_create().
+   ASSERT_EQUAL(thread_cfg->active_threads_bitmask,
+                task_bitmask_before_register);
+
+   // This is the important call to initialize thread-specific stuff in
+   // Splinter's task-system, which sets up the thread-id (index) and records
+   // this thread as active with the task system.
+   task_register_this_thread(thread_cfg->tasks, trunk_get_scratch_size());
```
ajhconway, line 251: Testing this is fine, but if we use the task thread creation machinery as I suggest above, then this call will happen as a hook.

This function is generally used only for the `main` thread or for threads otherwise created by a non-splinter context. So perhaps we should run this test on the main thread instead.

ajhconway (changes requested):
Thank you for doing this. These comments, tests and fixes are all very good.

A couple of testing changes would be good though.
1. The tests shouldn't call the pthread functions, but should call their platform and task wrappers (see comments in code).
2. We don't test `task_enqueue`, `task_perform_one` and `task_perform_all`, which is the primary way that the task system is used in the core code. It's fine if this is out of scope for this PR though.

src/task.c
gapisback: Thanks, @ajhconway - I will update the prolog accordingly.

Have opened issue #450 requesting that we have to work on re-enabling background threads support.

src/task.c
gapisback: Done.

src/platform_linux/platform.c
```diff
@@ -101,6 +101,9 @@ platform_buffer_destroy(buffer_handle *bh)
    return STATUS_OK;
 }
 
+/*
+ * platform_thread_create() - External interface to create a Splinter thread.
+ */
```
gapisback, line 106: (Added many such prologs to make the whole module easier to follow.)

src/platform_linux/platform.c
```diff
@@ -136,6 +139,12 @@ platform_thread_join(platform_thread thread)
    return CONST_STATUS(ret);
 }
 
+platform_thread
+platform_thread_id_self()
```
gapisback, line 143: Added, externalized for testing.

src/task.c
```diff
@@ -6,27 +6,31 @@
 
 #include "poison.h"
 
-#define MAX_HOOKS (2048)
+#define MAX_HOOKS (8)
```
gapisback, line 9: Reduced this .... Is this good enough or too low?  Let's discuss.

src/task.c
```diff
@@ -40,8 +44,17 @@ task_init_tid_bitmask(uint64 *tid_bitmask)
...
+
+ * The first thing a newly started thread that wants to use Splinter is expected
+ * to do is to "register itself" with the task system. This function is
+ * defined as a thread's hook and gets run as part of the thread's
+ * registration. This function initializes each thread's thread-ID, which
+ * is really an index into the array of thread-specific data structures.
+ */
 static void
-init_threadid(task_system *ts)
+task_init_threadid(task_system *ts)
```
gapisback, line 57: Renamed with `task_` prefix to be consistent with other names in this file.

src/task.c
```diff
@@ -133,23 +168,86 @@ task_register_hook(task_hook newhook)
...
+   /*
+    * For all newly created threads, the thread-index will be 0 before the hook
+    * function, run below, generates this thread's index. We do not want to
+    * register a thread twice; otherwise, it will simply burn up an index and
+    * possibly allocated scratch space.
+    * As threads generally will require scratch space to be allocated, an
+    * attempt to re-register a thread should find scratch space already
+    * allocated. Flag that.
+    */
+   platform_assert((thread_tid == 0)
```
gapisback, line 195: Added ... Useful assertion. Found and tripped on incorrect usages in test code that was resulting in nested register thread for the same task.

src/task.c
```diff
@@ -133,23 +168,86 @@ task_register_hook(task_hook newhook)
...
+task_deregister_thread(task_system *ts,
+                       const char  *file,
+                       const int    lineno,
+                       const char  *func)
 {
-   uint64 tid = platform_get_tid();
+   threadid tid = platform_get_tid();
+
+   // A registered thread should always have a non-zero tid (index).
+   if (tid == 0) {
```
gapisback, line 243: Useful check ... Found bugs in test code that was doing 2 de-registers, inadvertently.

src/task.c
```diff
@@ -168,24 +266,43 @@ typedef struct {
...
+ * the call-back function 'func' to now call Splinter APIs to do diff things.
+ * -----------------------------------------------------------------------------
+ */
 static void
 task_invoke_with_hooks(void *func_and_args)
 {
-   thread_invoke         *thread_to_start = (thread_invoke *)func_and_args;
-   platform_thread_worker func            = thread_to_start->func;
-   void                  *arg             = thread_to_start->arg;
+   thread_invoke         *thread_started = (thread_invoke *)func_and_args;
```
gapisback, line 284: Renamed variable from `thread_to_start` to `thread_started`, because by the time this function executes, it's already executing in the context of a newly fired up thread.

(Prior variable name was meaningful & correct in this fn's caller, where the thread was just going to be started.)

src/task.h
```diff
@@ -127,12 +127,25 @@ task_thread_create(const char            *name,
                    platform_thread       *thread);
 
 // Register the calling thread, allocating scratch space for it
+#define task_register_this_thread(ts, scratch_size)                            \
+   task_register_thread((ts), (scratch_size), __FILE__, __LINE__, __FUNCTION__)
```
gapisback, line 131: Changed this and on L141 below to become caller-macros, so I can pass-down code location tags to the lower level function, for improved diagnostics.

tests/unit/task_system_test.c
```diff
@@ -0,0 +1,496 @@
...
+ * in a real application code-flow). These interfaces will do all the stuff
+ * required to start a thread and set it up to perform Splinter work.
+ *
+ * This test case and its minion worker function,
+ * exec_one_thread_use_extern_apis(), is written to use external platform
+ * and task-system APIs. Otherwise, the testing done is identical to what's
+ * done in its sibling test case, test_one_thread_using_lower_apis.
+ * ------------------------------------------------------------------------
+ */
+CTEST2(task_system, test_one_thread_using_extern_apis)
```
gapisback, line 224: Added this new test case in 2nd commit, based on your review comments.

tests/unit/task_system_test.c
gapisback: @ajhconway -- Not sure I understand what to change here.

With the revised set of tests, let me know if your comment above is still relevant and do I need to readjust the test code in any way

tests/unit/task_system_test.c
```diff
@@ -0,0 +1,496 @@
...
+ * exec_one_thread_use_extern_apis() - Worker routine executed by a single
+ * thread.
+ *
+ * This worker method uses external interfaces to start a thread, thru
+ * task_invoke_with_hooks(). That interface registers the new thread with
+ * Splinter -before- invoking this worker function. Adjust the task-bitmask
+ * checks accordingly.
+ */
+static void
+exec_one_thread_use_extern_apis(void *arg)
```
gapisback, line 410: Worker fn added new in 2nd commit after your comments.

gapisback:
HI, @ajhconway -- I have revised the code and test code based on your feedback. Have done semi-major overhaul of the test code.

I think this should cover most of your feedback. The change set is back into the review cycle now

Thanks!

gapisback:
@ajhconway - I have addressed your comment (1), by reworking the test case and add a new one using just the external APIs. 

About (2): "_We don't test task_enqueue, task_perform_one and task_perform_all, which is the primary way that the task system is used in the core code. It's fine if this is out of scope for this PR though._"

I tried to work that in, but that will take me [more] time to understand how all these interfaces work and then to cobble together an execution flow which will invoke them. 

If you think that is very desirable, I can open an issue, but -- really -- I don't think I will have much time for that this month. (This work came out of Am behind my schedule to complete the PG-SplinterDB integration work, and am already running late on that.)

src/task.c
ajhconway: Good catch

tests/unit/task_system_test.c
ajhconway: Good now.

ajhconway (approved):
This looks good to me now, and I appreciate all the comments and cleanup.

One thing I would like is to preface the task tests with something to indicate that the external APIs are the preferred way of interacting and that the low level ones are not. I'm just trying to imagine a user looking at the test code as a template and trying to push them towards the external APIs.

PR #453, opened 2022-10-05
Flushing Bug Fix

ajhconway:
Fixes a bug during flushing where during the flush_fullest loop a split can occur, changing the value of num_children, which is used to trigger the end of the loop. num_children is read before the loop. The fix rechecks num_children each iteration.

Cleans up the flush_fullest mechanism to ensure (and assert) that only a non-empty flush can happen.


rtjohnso (approved):
LGTM

rosenhouse:
I know this merged already, but maybe as a follow-on: would it be practical to write a test case for this?

PR #455, opened 2022-10-07
Remove compaction byte limits and check for failed compactions

rtjohnso:
Compactions limit the result in terms of both number of tuples and in terms of number of bytes.

The tuple limit is needed because we preallocate space for the hashes for the subsequent filter build, but the byte limit is just for performance.  The code silently ignores when these limits are exceeded, resulting in dropped data.

This PR eliminates the byte-based limits, since they are just a performance policy that can be handled in the node splitting policy.

It also asserts that the tuple-based limits are not exceeded during a compaction, i.e. the system will crash instead of losing data.


ajhconway (approved):
LGTM

I might include a btree test of the `max_tuples` requirement, but it's fine to just open an issue.

PR #459, opened 2022-10-19
Expose stats in splinterdb layer

ajhconway:
Adds splinterdb_print_[insertion,lookup]_stats and splinterdb_reset stats to include/splinterdb/splinterdb.h.

### DIFF
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -307,4 +307,23 @@ splinterdb_iterator_get_current(splinterdb_iterator *iter, // IN
 int
 splinterdb_iterator_status(const splinterdb_iterator *iter);
 
+/*
+ * Statistics Printing
+ *
+ * Must set the use_stats config option.
+ *
+ * Prints insertion or lookup statistics. Both print cache statistics.
+ *
+ * Reset statistics clears all statistics, including cache statistics.
+ */
+
+void
+splinterdb_stats_print_insertion(const splinterdb *kvs);
+
+void
+splinterdb_stats_print_lookup(const splinterdb *kvs);
+
+void
+splinterdb_stats_reset(splinterdb *kvs);
+
 #endif // _SPLINTERDB_H_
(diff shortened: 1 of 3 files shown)


include/splinterdb/splinterdb.h
```diff
@@ -307,4 +307,20 @@ splinterdb_iterator_get_current(splinterdb_iterator *iter, // IN
...
+ * Prints insertion or lookup statistics. Both print cache statistics.
+ *
+ * Reset statistics clears all statistics, including cache statistics.
+ */
+
+void splinterdb_print_insertion_stats(const splinterdb *kvs);
+
+void splinterdb_print_lookup_stats(const splinterdb *kvs);
+
+void splinterdb_reset_stats(splinterdb *kvs);
```
rosenhouse, line 324: This is a super small thing, feel free to ignore, but I thought I'd share on the off chance you agree.

I have a weak preference for putting the `stats` before the verb, as a sort of hack-ish sub-namespacing for this category of functionality.  It slightly eases discovery, and makes it more natural to add other stats-related functions later.

For example if we did
```c
void splinterdb_stats_print_insertion(const splinterdb *kvs);

void splinterdb_stats_print_lookup(const splinterdb *kvs);

void splinterdb_stats_reset(splinterdb *kvs);
```
now, then if in the future we wanted to expose a structured stats-gathering feature, then it would slot in like
```c
typedef struct {
   ...
} splinterdb_stats;

void splinterdb_stats_get(splinterdb *kvs, splinterdb_stats *stats_out);
```

But again, feel free to say no, I don't feel strongly.

rosenhouse (approved):
LGTM, I'm looking forward to trying this out.

This inspired me to write up https://github.com/vmware/splinterdb/issues/461 as a possible follow-up

include/splinterdb/splinterdb.h
ajhconway: I generally prefer this sort of format too, so I've done this at the `splinterdb` layer. The old format is still there at the `trunk` layer though, so that should get changed eventually too. I've opened an issue.

PR #460, opened 2022-10-19
Robj/trunk variable sized keys

rtjohnso:
Modify trunk.c to support variable-sized keys.

Eliminates key padding in splinterdb.c.


rtjohnso:
Superceded by PR #464

Issue #458, opened 2022-10-13
Debug Assertion failure in btree_pack_post_loop() -> mini_destroy_unused() reporting mismatch in (mini->num_extents == mini->num_batches)


gapisback:
### Describe the bug

While trying to setup a manual unit-test to repro issue #457, ran into the following stack using debug build of Splinter:

```
Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=140737349751616) at ./nptl/pthread_kill.c:44
44	./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737349751616) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737349751616) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737349751616, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7d15476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7cfb7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x0000555555559c72 in platform_assert_false (filename=0x7ffff7fa19a0 "src/mini_allocator.c", linenumber=702,
    functionname=0x7ffff7fa1fb0 <__FUNCTION__.10> "mini_destroy_unused",
    expr=0x7ffff7fa1bb0 "(mini->num_extents == mini->num_batches)",
    message=0x7ffff7fa1b88 "num_extents=%lu, num_batches=%lu\n") at src/platform_linux/platform.c:344
#6  0x00007ffff7f656ef in mini_destroy_unused (mini=0x7fffffffe0a0) at src/mini_allocator.c:702
#7  0x00007ffff7f5f608 in btree_pack_post_loop (req=0x7fffffffbc40, last_key=...) at src/btree.c:2895
#8  0x00007ffff7f5fac2 in btree_pack (req=0x7fffffffbc40) at src/btree.c:2977
#9  0x00007ffff7f802ed in trunk_compact_bundle (arg=0x5555557ca2c0, scratch_buf=0x7ffff79e8ae8) at src/trunk.c:4840
#10 0x00007ffff7f97bea in task_group_perform_one (group=0x7ffff79e7800) at src/task.c:687
[...]

PR #469, opened 2022-11-06
Enable use of background threads in Splinter. Engage tests to verify basic stability of this existing feature.

gapisback:
This commit exercises use of background threads, a feature that has been implemented but hadn't been fully tested & engaged.

- Minor adjustments are done to task system initialization to manage the order in which background threads are started up.

- Fixed some checks in task_system_test to cope with the new variation that this test case could, optionally, be exercised with `--num-normal-bg-threads` | `--num-memtable-bg-threads` arguments.

- Enhanced config_parse() to recognize these two new options.
- Added a case to test.sh to exercise the task_system_test with background threads.

-  Stabilization fixes to get large inserts stress tests working. (New test withdrawn)

-  Add new test cases to test.sh, exercising background threads.

---------------
**NOTE**: Before you can run the `large_inserts_stress_test` that was leveraged to exercise background threads functionality,  you need to pull-in in-flight fix for issue https://github.com/vmware/splinterdb/issues/458; Otherwise you will run into that assertion first. 

The dev-branch were this repro has been constructed, `agurajada/478-Enable-bg-thread-feature`, pulls-in that commit already.

### DIFF
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -15,7 +15,6 @@
 
 #include "splinterdb/data.h"
 
-
 // Get a version string for this build of SplinterDB
 // Currently a git tag
 const char *
@@ -67,6 +66,23 @@ typedef struct {
    uint64 max_branches_per_node;
    uint64 use_stats;
    uint64 reclaim_threshold;
+
+   // Background threads configuration: Both have to be non-zero in order for
+   // background threads to be started. (It is an error for one to be zero
+   // while the other is non-zero.)
+   //
+   // - Memtable bg-threads work on Memtables tasks which are short but latency
+   //   sensitive. A rule of thumb is to allocate around 1 memtable bg-thread
+   //   for every 10 threads performing insertions. Too few memtable threads
+   //   can cause some insertions to have high latency.
+   // - Normal bg-threads work on task such as compacting branches in the trunk
+   //   and building filters. These tasks take longer and are less latency
+   //   sensitive. A rule of thumb is to allocate 1-2 normal background
+   //   threads for every thread performing insertions. Too few "normal"
+   //   background threads can cause disk I/O bandwidth to go underutilized.
+   uint64 num_memtable_bg_threads;
+   uint64 num_normal_bg_threads;
+
 } splinterdb_config;
 
 // Opaque handle to an opened instance of SplinterDB
--- a/src/splinterdb.c
+++ b/src/splinterdb.c
@@ -245,13 +245,14 @@ splinterdb_create_or_open(const splinterdb_config *kvs_cfg,      // IN
       goto deinit_kvhandle;
    }
 
-   uint8 num_bg_threads[NUM_TASK_TYPES] = {0}; // no bg threads
+   uint64 num_bg_threads[NUM_TASK_TYPES] = {0};
+   num_bg_threads[TASK_TYPE_MEMTABLE]    = kvs_cfg->num_memtable_bg_threads;
+   num_bg_threads[TASK_TYPE_NORMAL]      = kvs_cfg->num_normal_bg_threads;
 
    status = task_system_create(kvs->heap_id,
                                &kvs->io_handle,
                                &kvs->task_sys,
                                TRUE,
-                               FALSE,
                                num_bg_threads,
                                trunk_get_scratch_size());
    if (!SUCCESS(status)) {
@@ -373,11 +374,16 @@ splinterdb_close(splinterdb **kvs_in) // IN
    splinterdb *kvs = *kvs_in;
    platform_assert(kvs != NULL);
 
+   /*
+    * NOTE: These dismantling routines must appear in exactly the reverse
+    * order when these sub-systems were init'ed when a Splinter device was
+    * created or re-opened. Otherwise, asserts will trip.
+    */
    trunk_unmount(&kvs->spl);
    clockcache_deinit(&kvs->cache_handle);
    rc_allocator_unmount(&kvs->allocator_handle);
-   io_handle_deinit(&kvs->io_handle);
    task_system_destroy(kvs->heap_id, &kvs->task_sys);
+   io_handle_deinit(&kvs->io_handle);
 
    platform_free(kvs->heap_id, kvs);
    *kvs_in = (splinterdb *)NULL;
--- a/src/task.c
+++ b/src/task.c
@@ -26,6 +26,12 @@ static void
 task_system_io_register_thread(task_system *ts);
 // end forward declarations
 
+const char *task_type_name[] = {"TASK_TYPE_INVALID",
+                                "TASK_TYPE_MEMTABLE",
+                                "TASK_TYPE_NORMAL"};
+_Static_assert((ARRAY_SIZE(task_type_name) == NUM_TASK_TYPES),
+               "Array task_type_name[] is incorrectly sized.");
+
 /*
  * task_init_tid_bitmask() - Initialize the global bitmask of active threads in
  * the task system structure to indicate that no threads are currently active.
@@ -64,6 +70,11 @@ task_init_threadid(task_system *ts)
       uint64 tmp_bitmask = *tid_bitmask;
       // first bit set to 1 starting from LSB.
       uint64 pos = __builtin_ffsl(tmp_bitmask);
+
+      // If all threads are in-use, bitmask will be all 0s.
+      if (pos == 0) {
+         goto out;
+      }
       // builtin_ffsl returns the position plus 1.
       tid = pos - 1;
       // set bit at that position to 0, indicating in use.
@@ -80,7 +91,10 @@ task_init_threadid(task_system *ts)
    }
 
 out:
-   debug_assert(tid != INVALID_TID);
+   platform_assert((tid != INVALID_TID),
+                   "Cannot create a new thread as the limit on"
+                   " concurrent threads, %d, will be exceeded.\n",
+                   MAX_THREADS);
 
    // Sets thread-ID, for example tracked as thread-local storage.
    platform_set_tid(tid);
@@ -214,7 +228,7 @@ task_register_thread(task_system *ts,
    platform_assert(
       (ts->thread_scratch[thread_tid] == NULL),
       "[%s:%d::%s()] Scratch space found allocated at %p for thread with "
-      "index %lu.",
+      "index %lu.\n",
       file,
       lineno,
       func,
@@ -269,7 +283,8 @@ typedef struct {
 /*
  * -----------------------------------------------------------------------------
  * task_invoke_with_hooks() - Single interface to invoke a user-specified
- * call-back function, 'func', to perform Splinter work.
+ * call-back function, 'func', to perform Splinter work. Both user-threads'
+ * and background-threads' creation goes through this interface.
  *
  * A thread has been created with this function as the worker function. Also,
  * the thread-creator has registered another call-back function to execute.
@@ -291,6 +306,7 @@ task_invoke_with_hooks(void *func_and_args)
    // the actual Splinter work will be done.
    func(arg);
 
+   // For background threads', also, IO-deregistration will happen here.
    task_deregister_this_thread(thread_started->ts);
 
    platform_free(thread_started->heap_id, func_and_args);
@@ -312,7 +328,19 @@ task_create_thread_with_hooks(platform_thread       *thread,
                               platform_heap_id       hid)
 {
    platform_status ret;
-   thread_invoke  *thread_to_create = TYPED_ZALLOC(hid, thread_to_create);
+   uint64         *tid_bitmask = task_system_get_tid_bitmask(ts);
+   if (*tid_bitmask == 0) {
+      platform_error_log("Cannot create a new thread as the limit on"
+                         " concurrent threads, %d, will be exceeded.\n",
+                         MAX_THREADS);
+
+      // This is an approximate status. STATUS_LIMIT_EXCEEDED would be more
+      // accurate but that maps to ENOSPC, which will result in a misleading
+      // 'No space left on device' error message.
+      return (STATUS_BUSY);
+   }
+
+   thread_invoke *thread_to_create = TYPED_ZALLOC(hid, thread_to_create);
    if (thread_to_create == NULL) {
       return STATUS_NO_MEMORY;
    }
@@ -338,7 +366,7 @@ task_create_thread_with_hooks(platform_thread       *thread,
  *  required registration with Splinter.
  *
  * Currently, this is active mainly in tests. It's also used to create
- * background tasks, a feature that is currently not enabled.
+ * background tasks.
  * -----------------------------------------------------------------------------
  */
 platform_status
@@ -356,7 +384,8 @@ task_thread_create(const char            *name,
    ret = task_create_thread_with_hooks(
       &thr, FALSE, func, arg, scratch_size, ts, hid);
    if (!SUCCESS(ret)) {
-      platform_error_log("could not create a thread");
+      platform_error_log("Could not create a thread: %s\n",
+                         platform_status_to_string(ret));
       return ret;
    }
 
@@ -366,7 +395,13 @@ task_thread_create(const char            *name,
    return STATUS_OK;
 }
 
-/* Worker function for the background task pool. */
+/*
+ * task_worker_thread() - Worker function for the background task pool.
+ *
+ * This function is invoked when configured background threads are created.
+ * We sit in an endless-loop looking for work to do and execute the tasks
+ * enqueued.
+ */
 static void
 task_worker_thread(void *arg)
 {
@@ -430,6 +465,9 @@ task_worker_thread(void *arg)
    }
 }
 
+/*
+ * Function to terminate all background threads and clean up.
+ */
 static void
 task_group_stop_and_wait_for_threads(task_group *group)
 {
@@ -450,10 +488,12 @@ task_group_stop_and_wait_for_threads(task_group *group)
 
    uint8 num_threads = group->bg.num_threads;
 
+   // Inform the background thread that it's time to exit now.
    group->bg.stop = TRUE;
    platform_condvar_broadcast(&group->bg.cv);
    platform_condvar_unlock(&group->bg.cv);
 
+   // Allow all background threads to wrap up their work.
    for (uint8 i = 0; i < num_threads; i++) {
       platform_thread_join(group->bg.threads[i]);
    }
@@ -719,6 +759,37 @@ task_perform_one(task_system *ts)
 }
 
 /*
+ * Validate that the task system configuration is basically supportable.
+ */
+static platform_status
+task_config_valid(uint64 num_bg_threads[NUM_TASK_TYPES])
+{
+   uint64 normal_bg_threads   = num_bg_threads[TASK_TYPE_NORMAL];
+   uint64 memtable_bg_threads = num_bg_threads[TASK_TYPE_MEMTABLE];
+
+   if ((normal_bg_threads == 0) != (memtable_bg_threads == 0)) {
+      platform_error_log("Both configuration parameters for background "
+                         "threads, normal_bg_threads (%lu) "
+                         "and memtable_bg_threads (%lu) "
+                         "must be zero or be non-zero.\n",
+                         normal_bg_threads,
+                         memtable_bg_threads);
+      return STATUS_BAD_PARAM;
+   }
+
+   if ((normal_bg_threads + memtable_bg_threads) >= MAX_THREADS) {
+      platform_error_log("Total number of background threads configured"
+                         ", normal_bg_threads=%lu, memtable_bg_threads=%lu, "
+                         "must be <= %d.\n",
+                         normal_bg_threads,
+                         memtable_bg_threads,
+                         (MAX_THREADS - 1));
+      return STATUS_BAD_PARAM;
+   }
+   return STATUS_OK;
+}
+
+/*
  * -----------------------------------------------------------------------------
  * Task system initializer. Makes sure that the initial thread has an
  * adequately sized scratch space.
@@ -732,10 +803,14 @@ task_system_create(platform_heap_id    hid,
                    platform_io_handle *ioh,
                    task_system       **system,
                    bool                use_stats,
-                   bool                use_bg_threads,
-                   uint8               num_bg_threads[NUM_TASK_TYPES],
+                   uint64              num_bg_threads[NUM_TASK_TYPES],
                    uint64              scratch_size)
 {
+   platform_status rc = task_config_valid(num_bg_threads);
+   if (!SUCCESS(rc)) {
+      return rc;
+   }
+
    task_system *ts =
       TYPED_FLEXIBLE_STRUCT_ZALLOC(hid, ts, init_task_scratch, scratch_size);
 
@@ -748,10 +823,20 @@ task_system_create(platform_heap_id    hid,
    // task initialization
    register_init_tid_hook();
 
+   // Task system will create background threads when both configs are set.
+   bool use_bg_threads = ((num_bg_threads[TASK_TYPE_MEMTABLE] != 0)
+                          && (num_bg_threads[TASK_TYPE_NORMAL] != 0));
+
    ts->use_bg_threads = use_bg_threads;
    ts->heap_id        = hid;
    ts->scratch_size   = scratch_size;
    ts->init_tid       = INVALID_TID;
+
+   // Ensure that the main thread gets registered and init'ed first before
+   // any background threads are created. (Those may grab their own tids.).
+   task_run_thread_hooks(ts);
+   const threadid tid      = platform_get_tid();
+   ts->thread_scratch[tid] = ts->init_task_scratch;
 
    for (task_type type = TASK_TYPE_FIRST; type != NUM_TASK_TYPES; type++) {
       platform_status rc = task_group_init(&ts->group[type],
@@ -765,12 +850,18 @@ task_system_create(platform_heap_id    hid,
          *system = NULL;
          return rc;
       }
+      uint64 nbg_threads = num_bg_threads[type];
+      if (nbg_threads) {
+         platform_default_log("Splinter task system created %lu"
+                              " background thread%sof type '%s'.\n",
+                              nbg_threads,
+                              ((nbg_threads > 1) ? "s " : " "),
+                              task_type_name[type]);
+      }
    }
-
-   task_run_thread_hooks(ts);
-   const threadid tid      = platform_get_tid();
-   ts->thread_scratch[tid] = ts->init_task_scratch;
-
+   debug_assert((*system == NULL),
+                "Task system handle, %p, is expected to be NULL.\n",
+                *system);
    *system = ts;
    return STATUS_OK;
 }
--- a/test.sh
+++ b/test.sh
@@ -84,7 +84,7 @@ function record_elapsed_time() {
       # Provide wider test-tag for nightly tests which print verbose descriptions
       fmtstr="%-80s""${fmtstr}"
    else
-      fmtstr="%-50s""${fmtstr}"
+      fmtstr="%-70s""${fmtstr}"
    fi
 
    # Log a line in the /tmp log-file; for future cat of summary output
@@ -510,6 +510,7 @@ function test_make_run_tests() {
 # Smoke Tests: Run a small collection of fast-running unit-tests
 # ##################################################################
 function run_fast_unit_tests() {
+
    "$BINDIR"/unit/splinterdb_quick_test
    "$BINDIR"/unit/btree_test
    "$BINDIR"/unit/util_test
@@ -522,7 +523,7 @@ function run_fast_unit_tests() {
 # Run mini-unit-tests that were excluded from bin/unit_test binary:
 # Explicitly run individual cases from specific slow running unit-tests,
 # where appropriate with a different test-configuration that has been
-# found to to provide the required coverage.
+# found to provide the required coverage.
 # ##################################################################
 function run_slower_unit_tests() {
 
@@ -549,6 +550,11 @@ function run_splinter_functionality_tests() {
         "$BINDIR"/driver_test splinter_test --functionality 1000000 100 \
                                             --seed "$SEED"
 
+    run_with_timing "Functionality test, default key size, with background threads" \
+        "$BINDIR"/driver_test splinter_test --functionality 1000000 100 \
+                                            --num-normal-bg-threads 4 --num-memtable-bg-threads 2 \
+                                            --seed "$SEED"
+
     max_key_size=102
     run_with_timing "Functionality test, key size=maximum (${max_key_size} bytes)" \
         "$BINDIR"/driver_test splinter_test --functionality 1000000 100 \
@@ -572,6 +578,17 @@ function run_splinter_perf_tests() {
                                             --num-inserts 10000 \
                                             --cache-capacity-mib 512 \
                                             --verbose-progress
+
+   # Re-run small perf test configuring background threads. This scenario
+   # validates that we can configure bg- and user-threads in one go.
+   run_with_timing "Quick Performance test with bg-threads" \
+        "$BINDIR"/driver_test splinter_test --perf \
+                                            --num-insert-threads 4 \
+                                            --num-lookup-threads 4 \
+                                            --num-inserts 10000 \
+                                            --cache-capacity-mib 512 \
+                                            --num-normal-bg-threads 1 \
+                                            --num-memtable-bg-threads 1
 
    run_with_timing "Performance test" \
         "$BINDIR"/driver_test splinter_test --perf \
--- a/tests/config.c
+++ b/tests/config.c
@@ -36,7 +36,23 @@
 #define TEST_CONFIG_DEFAULT_SEED        0
 #define TEST_CONFIG_DEFAULT_NUM_INSERTS 0
 
+// By default, background threads are disabled in Splinter task system.
+// Most tests run w/o background threads. Very small # of tests exercise
+// background threads through the --num-normal-bg-threads and
+// --num-memtable-bg-threads options.
+#define TEST_CONFIG_DEFAULT_NUM_NORMAL_BG_THREADS   0
+#define TEST_CONFIG_DEFAULT_NUM_MEMTABLE_BG_THREADS 0
+
+
 // clang-format off
+/*
+ * ---------------------------------------------------------------------------
+ * Helper function to initialize master_config{} used to run tests with some
+ * useful default values. The expectation is that the input 'cfg' is zero'ed
+ * out before calling this initializer, so that all other fields will have
+ * some reasonable 0-defaults.
+ * ---------------------------------------------------------------------------
+ */
 void
 config_set_defaults(master_config *cfg)
 {
@@ -59,10 +75,12 @@ config_set_defaults(master_config *cfg)
       .max_branches_per_node    = TEST_CONFIG_DEFAULT_MAX_BRANCHES_PER_NODE,
       .use_stats                = FALSE,
       .reclaim_threshold        = UINT64_MAX,
+      .num_normal_bg_threads    = TEST_CONFIG_DEFAULT_NUM_NORMAL_BG_THREADS,
+      .num_memtable_bg_threads  = TEST_CONFIG_DEFAULT_NUM_MEMTABLE_BG_THREADS,
       .verbose_logging_enabled  = FALSE,
       .verbose_progress         = FALSE,
       .log_handle               = NULL,
-      .max_key_size                 = TEST_CONFIG_DEFAULT_KEY_SIZE,
+      .max_key_size             = TEST_CONFIG_DEFAULT_KEY_SIZE,
       .message_size             = TEST_CONFIG_DEFAULT_MESSAGE_SIZE,
       .num_inserts              = TEST_CONFIG_DEFAULT_NUM_INSERTS,
       .seed                     = TEST_CONFIG_DEFAULT_SEED,
@@ -105,6 +123,13 @@ config_usage()
    platform_error_log("\t--fanout (%d)\n", TEST_CONFIG_DEFAULT_FANOUT);
    platform_error_log("\t--max-branches-per-node (%d)\n",
                       TEST_CONFIG_DEFAULT_MAX_BRANCHES_PER_NODE);
+
+   platform_error_log("\t--num-normal-bg-threads (%d)\n",
+                      TEST_CONFIG_DEFAULT_NUM_NORMAL_BG_THREADS);
+
+   platform_error_log("\t--num-memtable-bg-threads (%d)\n",
+                      TEST_CONFIG_DEFAULT_NUM_MEMTABLE_BG_THREADS);
+
    platform_error_log("\t--stats\n");
    platform_error_log("\t--no-stats\n");
    platform_error_log("\t--log\n");
@@ -225,6 +250,15 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
          {}
          config_set_mib("reclaim-threshold", cfg, reclaim_threshold) {}
          config_set_gib("reclaim-threshold", cfg, reclaim_threshold) {}
+
+         /*
+          * These arguments will be passed through to Splinter initialization
+          * to setup Splinter task system to use background threads.
+          */
+         config_set_uint64("num-normal-bg-threads", cfg, num_normal_bg_threads);
+         config_set_uint64(
+            "num-memtable-bg-threads", cfg, num_memtable_bg_threads);
+
          config_has_option("stats")
          {
             for (uint8 cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
@@ -283,6 +317,8 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
             return STATUS_BAD_PARAM;
          }
       }
+
+      // Validate consistency of config parameters provided.
       for (uint8 cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
          if (cfg[cfg_idx].extent_size % cfg[cfg_idx].page_size != 0) {
             platform_error_log("Configured extent-size, %lu, is not a multiple "
@@ -314,6 +350,17 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
                                "experimental.\n",
                                cfg[cfg_idx].max_key_size,
                                TEST_CONFIG_MIN_KEY_SIZE);
+            return STATUS_BAD_PARAM;
+         }
+         if ((cfg[cfg_idx].num_normal_bg_threads == 0)
+             != (cfg[cfg_idx].num_memtable_bg_threads == 0))
+         {
+            platform_error_log("Both configuration parameters for background "
+                               "threads, --num-normal-bg-threads (%lu) "
+                               " and --num-memtable-bg-threads (%lu) "
+                               "must be zero or be non-zero.\n",
+                               cfg[cfg_idx].num_normal_bg_threads,
+                               cfg[cfg_idx].num_memtable_bg_threads);
             return STATUS_BAD_PARAM;
          }
       }
--- a/tests/config.h
+++ b/tests/config.h
@@ -71,13 +71,17 @@ typedef struct master_config {
    bool use_log;
 
    // splinter
-   uint64               memtable_capacity;
-   uint64               fanout;
-   uint64               max_branches_per_node;
-   uint64               use_stats;
-   uint64               reclaim_threshold;
-   bool                 verbose_logging_enabled;
-   bool                 verbose_progress;
+   uint64 memtable_capacity;
+   uint64 fanout;
+   uint64 max_branches_per_node;
+   uint64 use_stats;
+   uint64 reclaim_threshold;
+   uint64 num_normal_bg_threads;   // Both bg_threads fields have to be non-zero
+   uint64 num_memtable_bg_threads; // for background threads to be enabled
+
+   bool verbose_logging_enabled;
+   bool verbose_progress;
+
    platform_log_handle *log_handle;
 
    // data
--- a/tests/functional/splinter_test.c
+++ b/tests/functional/splinter_test.c
@@ -2383,7 +2383,9 @@ usage(const char *argv0)
       "\t%s --cache-per-table\n"
       "\t%s --parallel-perf --max-async-inflight [num] --num-pthreads [num] "
       "--lookup-positive-percent [num] --seed [num]\n"
-      "\t%s --num-bg-threads (number of background threads)\n"
+      "\t%s --num-normal-bg-threads (number of normal background threads)\n"
+      "\t\t      --num-memtable-bg-threads (number of background threads for "
+      "memtables)\n"
       "\t%s --insert-rate (inserts_done_by_all_threads in a second)\n",
       argv0,
       argv0,
@@ -2498,15 +2500,13 @@ splinter_test(int argc, char *argv[])
    uint64              test_ops;
    uint64              correctness_check_frequency;
    // Max async IOs inflight per-thread
-   uint32 num_insert_threads, num_lookup_threads;
-   uint32 num_range_lookup_threads, max_async_inflight;
-   uint32 num_pthreads    = 0;
-   uint8  num_tables      = 1;
-   bool   cache_per_table = FALSE;
-   // no bg threads by default.
-   uint8                  num_bg_threads[NUM_TASK_TYPES] = {0};
-   uint64                 insert_rate = 0; // no rate throttling by default.
-   task_system           *ts;
+   uint32                 num_insert_threads, num_lookup_threads;
+   uint32                 num_range_lookup_threads, max_async_inflight;
+   uint32                 num_pthreads    = 0;
+   uint8                  num_tables      = 1;
+   bool                   cache_per_table = FALSE;
+   uint64                 insert_rate     = 0; // no rate throttling by default.
+   task_system           *ts              = NULL;
    uint8                  lookup_positive_pct = 0;
    test_message_generator gen;
    test_exec_config       test_exec_cfg;
@@ -2602,6 +2602,7 @@ splinter_test(int argc, char *argv[])
       config_argc -= 2;
       config_argv += 2;
    }
+
    if (config_argc > 0
        && strncmp(
              config_argv[0], "--cache-per-table", sizeof("--cache-per-table"))
@@ -2611,33 +2612,7 @@ splinter_test(int argc, char *argv[])
       config_argc -= 1;
       config_argv += 1;
    }
-   if (config_argc > 0
-       && strncmp(
-             config_argv[0], "--num-bg-threads", sizeof("--num-bg-threads"))
-             == 0)
-   {
-      if (!try_string_to_uint8(config_argv[1],
-                               &num_bg_threads[TASK_TYPE_NORMAL])) {
-         usage(argv[0]);
-         return -1;
-      }
-      config_argc -= 2;
-      config_argv += 2;
-   }
-   if (config_argc > 0
-       && strncmp(config_argv[0],
-                  "--num-memtable-bg-threads",
-                  sizeof("--num-bg-threads"))
-             == 0)
-   {
-      if (!try_string_to_uint8(config_argv[1],
-                               &num_bg_threads[TASK_TYPE_MEMTABLE])) {
-         usage(argv[0]);
-         return -1;
-      }
-      config_argc -= 2;
-      config_argv += 2;
-   }
+
    if (splinter_test_parse_perf_args(&config_argv,
                                      &config_argc,
                                      &max_async_inflight,
@@ -2718,6 +2693,9 @@ splinter_test(int argc, char *argv[])
    clockcache_config *cache_cfg =
       TYPED_ARRAY_MALLOC(hid, cache_cfg, num_tables);
 
+   // no bg threads by default.
+   uint64 num_bg_threads[NUM_TASK_TYPES] = {0};
+
    rc = test_parse_args_n(splinter_cfg,
                           &data_cfg,
                           &io_cfg,
@@ -2726,6 +2704,8 @@ splinter_test(int argc, char *argv[])
                           &log_cfg,
                           &test_exec_cfg,
                           &gen,
+                          &num_bg_threads[TASK_TYPE_MEMTABLE],
+                          &num_bg_threads[TASK_TYPE_NORMAL],
                           num_tables,
                           config_argc,
                           config_argv);
@@ -2788,7 +2768,7 @@ splinter_test(int argc, char *argv[])
    bool use_bg_threads = num_bg_threads[TASK_TYPE_NORMAL] != 0;
 
    rc = test_init_task_system(
-      hid, io, &ts, splinter_cfg->use_stats, use_bg_threads, num_bg_threads);
+      hid, io, &ts, splinter_cfg->use_stats, num_bg_threads);
    if (!SUCCESS(rc)) {
       platform_error_log("Failed to init splinter state: %s\n",
                          platform_status_to_string(rc));
--- a/tests/unit/limitations_test.c
+++ b/tests/unit/limitations_test.c
@@ -44,8 +44,6 @@ CTEST_DATA(limitations)
 
    rc_allocator al;
 
-   uint8 num_bg_threads[NUM_TASK_TYPES];
-
    // Following get setup pointing to allocated memory
    trunk_config          *splinter_cfg;
    data_config           *data_cfg;
@@ -65,7 +63,7 @@ CTEST_DATA(limitations)
  */
 CTEST_SETUP(limitations)
 {
-   // This test exercises error cases, so even when everthing succeeds
+   // This test exercises error cases, so even when everything succeeds
    // it generates lots of "error" messages.
    // By default, that would go to stderr, which would pollute test output.
    // Here we ensure those expected error messages are only printed
@@ -115,6 +113,9 @@ CTEST2(limitations, test_io_init_invalid_page_size)
 
    data->cache_cfg = TYPED_ARRAY_MALLOC(data->hid, data->cache_cfg, num_tables);
 
+   uint64 num_memtable_bg_threads_unused = 0;
+   uint64 num_normal_bg_threads_unused   = 0;
+
    ZERO_STRUCT(data->test_exec_cfg);
 
    rc = test_parse_args_n(data->splinter_cfg,
@@ -125,6 +126,8 @@ CTEST2(limitations, test_io_init_invalid_page_size)
                           &data->log_cfg,
                           &data->test_exec_cfg,
                           &data->gen,
+                          &num_memtable_bg_threads_unused,
+                          &num_normal_bg_threads_unused,
                           num_tables,
                           Ctest_argc, // argc/argv globals setup by CTests
                           (char **)Ctest_argv);
@@ -185,6 +188,9 @@ CTEST2(limitations, test_io_init_invalid_extent_size)
 
    data->cache_cfg = TYPED_ARRAY_MALLOC(data->hid, data->cache_cfg, num_tables);
 
+   uint64 num_memtable_bg_threads_unused = 0;
+   uint64 num_normal_bg_threads_unused   = 0;
+
    ZERO_STRUCT(data->test_exec_cfg);
 
    rc = test_parse_args_n(data->splinter_cfg,
@@ -195,6 +201,8 @@ CTEST2(limitations, test_io_init_invalid_extent_size)
                           &data->log_cfg,
                           &data->test_exec_cfg,
                           &data->gen,
+                          &num_memtable_bg_threads_unused,
+                          &num_normal_bg_threads_unused,
                           num_tables,
                           Ctest_argc, // argc/argv globals setup by CTests
                           (char **)Ctest_argv);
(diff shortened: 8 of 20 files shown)


gapisback:
NOTE: Enabling background threads, if it does not run into the issues reported earlier, this feature does seem to show substantial gains in insert performance. 

For instance, with 4 inserting threads, and `--num-bg-threads 2` and `--num-memtable-bg-threads 1`, these two cases show:

- test_seq_key_seq_values_inserts_threaded: 25% improved throughput averaged across 4 inserting threads
- test_seq_key_seq_values_inserts_threaded_same_start_keyid: 35% improved throughput averaged across 4 inserting threads

Here is the evidence:

**Without background threads**: 

`$ build/release/bin/unit/large_inserts_bugs_stress_test --num-inserts 20000000 --num-threads 4 test_seq_key_seq_values_inserts_threaded`

```
ip-172-31-29-15:[48] $ build/release/bin/unit/large_inserts_bugs_stress_test --num-inserts 20000000 --num-threads 4 test_seq_key_seq_values_inserts_threaded
Running 1 CTests, suite name 'large_inserts_bugs_stress', test case 'test_seq_key_seq_values_inserts_threaded'.
TEST 1/2 large_inserts_bugs_stress:test_seq_key_seq_values_inserts_threaded Fingerprint size 29 too large, max value size is 5, setting to 27
fingerprint_size: 27
exec_worker_thread()::495:Thread 1  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141578, Thread-ID=1, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 3  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 40000000 (40 million) ...
exec_worker_thread()::495:Thread 4  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 60000000 (60 million) ...
OS-pid=141578, Thread-ID=4, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 2  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 20000000 (20 million) ...
OS-pid=141578, Thread-ID=2, Insert small-width sequential values of different lengths.
OS-pid=141578, Thread-ID=3, Insert small-width sequential values of different lengths.
exec_worker_thread()::601:Thread-3 Inserted 20 million KV-pairs in 25 s, 800000 rows/s
exec_worker_thread()::601:Thread-1 Inserted 20 million KV-pairs in 27 s, 740740 rows/s
exec_worker_thread()::601:Thread-2 Inserted 20 million KV-pairs in 27 s, 740740 rows/s
exec_worker_thread()::601:Thread-4 Inserted 20 million KV-pairs in 27 s, 740740 rows/s
Allocated at unmount: 3948 MiB
[OK]
TEST 2/2 large_inserts_bugs_stress:test_seq_key_seq_values_inserts_threaded_same_start_keyid Fingerprint size 29 too large, max value size is 5, setting to 27
fingerprint_size: 27
exec_worker_thread()::495:Thread 1  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141578, Thread-ID=1, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 2  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141578, Thread-ID=2, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 3  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141578, Thread-ID=3, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 4  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141578, Thread-ID=4, Insert small-width sequential values of different lengths.
exec_worker_thread()::601:Thread-4 Inserted 20 million KV-pairs in 21 s, 952380 rows/s
exec_worker_thread()::601:Thread-3 Inserted 20 million KV-pairs in 22 s, 909090 rows/s
exec_worker_thread()::601:Thread-1 Inserted 20 million KV-pairs in 22 s, 909090 rows/s
exec_worker_thread()::601:Thread-2 Inserted 20 million KV-pairs in 25 s, 800000 rows/s
Allocated at unmount: 1451 MiB
[OK]
RESULTS: 2 tests (2 ok, 0 failed, 0 skipped) ran in 53271 ms
```

**With background threads enabled**: 

`$ build/release/bin/unit/large_inserts_bugs_stress_test --num-inserts 20000000 --num-threads 4 --num-bg-threads 2 --num-memtable-bg-threads 1 test_seq_key_seq_values_inserts_threaded`

```
ip-172-31-29-15:[61] $ build/release/bin/unit/large_inserts_bugs_stress_test --num-inserts 20000000 --num-threads 4 --num-bg-threads 2 --num-memtable-bg-threads 1 test_seq_key_seq_values_inserts_threaded
Running 1 CTests, suite name 'large_inserts_bugs_stress', test case 'test_seq_key_seq_values_inserts_threaded'.
TEST 1/2 large_inserts_bugs_stress:test_seq_key_seq_values_inserts_threaded Fingerprint size 29 too large, max value size is 5, setting to 27
fingerprint_size: 27
Splinter task system created 1 background threads of type 'TASK_TYPE_MEMTABLE'.
Splinter task system created 2 background threads of type 'TASK_TYPE_NORMAL'.
exec_worker_thread()::495:Thread 4  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141859, Thread-ID=4, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 5  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 20000000 (20 million) ...
OS-pid=141859, Thread-ID=5, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 6  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 40000000 (40 million) ...
OS-pid=141859, Thread-ID=6, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 7  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 60000000 (60 million) ...
OS-pid=141859, Thread-ID=7, Insert small-width sequential values of different lengths.
exec_worker_thread()::601:Thread-7 Inserted 20 million KV-pairs in 21 s, 952380 rows/s
exec_worker_thread()::601:Thread-5 Inserted 20 million KV-pairs in 21 s, 952380 rows/s
exec_worker_thread()::601:Thread-4 Inserted 20 million KV-pairs in 21 s, 952380 rows/s
exec_worker_thread()::601:Thread-6 Inserted 20 million KV-pairs in 21 s, 952380 rows/s
Allocated at unmount: 3954 MiB
[OK]
TEST 2/2 large_inserts_bugs_stress:test_seq_key_seq_values_inserts_threaded_same_start_keyid Fingerprint size 29 too large, max value size is 5, setting to 27
fingerprint_size: 27
Splinter task system created 1 background threads of type 'TASK_TYPE_MEMTABLE'.
Splinter task system created 2 background threads of type 'TASK_TYPE_NORMAL'.
exec_worker_thread()::495:Thread 4  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141859, Thread-ID=4, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 5  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141859, Thread-ID=5, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 6  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141859, Thread-ID=6, Insert small-width sequential values of different lengths.
exec_worker_thread()::495:Thread 7  inserts 20000000 (20 million), sequential key, seqential value, KV-pairs starting from 0 (0) ...
OS-pid=141859, Thread-ID=7, Insert small-width sequential values of different lengths.
exec_worker_thread()::601:Thread-4 Inserted 20 million KV-pairs in 16 s, 1250000 rows/s
exec_worker_thread()::601:Thread-7 Inserted 20 million KV-pairs in 16 s, 1250000 rows/s
exec_worker_thread()::601:Thread-5 Inserted 20 million KV-pairs in 17 s, 1176470 rows/s
exec_worker_thread()::601:Thread-6 Inserted 20 million KV-pairs in 17 s, 1176470 rows/s
Allocated at unmount: 1084 MiB
[OK]
RESULTS: 2 tests (2 ok, 0 failed, 0 skipped) ran in 39955 ms
```

include/splinterdb/splinterdb.h
```diff
@@ -72,6 +74,8 @@ typedef struct {
    uint64 max_branches_per_node;
    uint64 use_stats;
    uint64 reclaim_threshold;
+   uint8  num_bg_threads[SPLINTERDB_NUM_TASK_TYPES]; // = {0} => no bg threads
```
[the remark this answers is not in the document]
ajhconway: Also this doesn't make clear which type is which and the enum is defined in `task.c`, so it's not straightforward to translate.

What I think we should do is have `num_memtable_bg_threads` and `num_normal_bg_threads` here and get rid of `SplinterDB_NUM_TASK_TYPES`.

I don't really like having the application provide this split to be honest. I think that the application should provide a `max_bg_threads` parameter and `task.c` should handle the load balancing.

src/task.c
```diff
@@ -26,6 +26,12 @@ static void
 task_system_io_register_thread(task_system *ts);
 // end forward declarations
 
+const char *task_type_name[] = {"TASK_TYPE_INVALID",
+                                "TASK_TYPE_MEMTABLE",
+                                "TASK_TYPE_NORMAL"};
+_Static_assert((ARRAY_SIZE(task_type_name) == NUM_TASK_TYPES),
+               "Array task_type_name[] is incorrectly sized.");
+
```
ajhconway, line 34: What's the point of the invalid task type? It interacts weirdly with the array in `splinterdb.h` and I don't see the purpose.

src/task.c
```diff
@@ -753,9 +777,18 @@ task_system_create(platform_heap_id    hid,
    ts->scratch_size   = scratch_size;
    ts->init_tid       = INVALID_TID;
 
+   // Ensure that the main threads gets registered and init'ed first before
```
ajhconway, line 780: thread

ajhconway (changes requested):
I'm glad you're kicking the tires on background threads. It's a feature we really want to support.

I only want minor changes.

I think we should kill `TASK_TYPE_INVALID` because it's awkward and I don't see the functionality. It's weird in the `num_bg_threads[]` array, and it causes the creation of another `task_group`.

I also think that at the `splinterdb.[ch]` level we shouldn't expose task types, but rather "memtable" and "normal" bg thread counts.

Ideally we should have `task.[ch]` manage the load balance between threads, but that's for another day.

include/splinterdb/splinterdb.h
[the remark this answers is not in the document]
rtjohnso: I second that we need to either define two separate `bg_threads` fields or at least expose the `#define`s for the different classes so that users can use those to set the classes.

Also 100% agree this should all be automated by task.c so the user just has to specify the max total bg threads (or maybe even just whether to use bg threads or not).  But that's probably out of scope for your PR.

src/splinterdb.c
```diff
@@ -437,13 +437,15 @@ splinterdb_create_or_open(const splinterdb_config *kvs_cfg,      // IN
       goto deinit_kvhandle;
    }
 
-   uint8 num_bg_threads[NUM_TASK_TYPES] = {0}; // no bg threads
+   uint8 *num_bg_threads = (uint8 *)kvs_cfg->num_bg_threads;
+   bool   use_bg_threads = ((num_bg_threads[TASK_TYPE_NORMAL] != 0)
+                          || (num_bg_threads[TASK_TYPE_MEMTABLE] != 0));
```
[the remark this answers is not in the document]
rtjohnso: I think Alex's point is that, once bg threads are enabled, fg threads never do any bg work at all.

So, if one of these is 0 and the other is not, then the task type with 0 bg threads will get ignored!

One option is to reject configs in which one of these is 0 and the other is not.

src/splinterdb.c
```diff
@@ -437,13 +437,15 @@ splinterdb_create_or_open(const splinterdb_config *kvs_cfg,      // IN
...
 
    status = task_system_create(kvs->heap_id,
                                &kvs->io_handle,
                                &kvs->task_sys,
                                TRUE,
-                               FALSE,
+                               use_bg_threads,
```
rtjohnso, line 448: Can we just delete this flag and have task.c infer it from the `num_bg_threads` arg?

tests/config.h
```diff
@@ -72,13 +74,17 @@ typedef struct master_config {
...
-   uint64               reclaim_threshold;
-   bool                 verbose_logging_enabled;
-   bool                 verbose_progress;
+   uint64 memtable_capacity;
+   uint64 fanout;
+   uint64 max_branches_per_node;
+   uint64 use_stats;
+   uint64 reclaim_threshold;
+   uint8  num_bg_threads;          // = 0 => no background threads
+   uint8  num_memtable_bg_threads; // = 0 => no Memtable background threads
```
rtjohnso, line 83: Just make both of these `uint64`s.

tests/unit/task_system_test.c
```diff
@@ -61,7 +61,8 @@ CTEST_DATA(task_system)
    // Config structs required, to exercise task subsystem
    io_config io_cfg;
 
-   uint8 num_bg_threads[NUM_TASK_TYPES];
+   uint8  num_bg_threads[NUM_TASK_TYPES];
```
rtjohnso, line 64: `uint64`

rtjohnso (changes requested):
See comments about exposing task types to users and using `uint64`s instead of `uint8`s.

include/splinterdb/splinterdb.h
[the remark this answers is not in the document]
gapisback: @ajhconway , @rtjohnso - Based on what we discussed in review meeting that this PR does not have to reconcile the interfaces and that we can get this 'enablement feature' in first, here is what I did to address your comments:

1. Externalized two new fields, `num_memtable_bg_threads` and `num_normal_bg_threads` to splinter_config{}.
2. Interface to `task_system_create()` remains unchanged. It will receive `use_bg_threads == TRUE` only if both fields above are non-zero. (This addresses one of the concerns that the system will be unstable if we configure only one of these task-types with bg-threads count.)
3. Config options are now renamed symmetrically to be `--num-normal-bg-threads` and `--num-memtable-bg-threads`.
4. Added checks in `config_parse()` to enforce that both config params are == 0 or both are != 0. (Pulled-in similar checks Rob added in [PR 497, here](https://github.com/vmware/splinterdb/pull/497/files#diff-b74ae5c4f955f7aaf2dabd10a7a4cdbc03e7afff831220370151d6b01b05ce43R345).)
5. Amended cases in `task_system_test.c` and calls in `test.sh` to invoke this test-case with and w/o background threads configured.

These changes should wholesale address the overall issues raised in this comment-stream.

src/task.c
gapisback: I did not try to change or correct this. The array sized `[3]` is what's existing in code. We have talked about streamlining this interface to just provide total-number-of-bg-threads, at which time ... these kinds of warts will be automatically resolved.

Hopefully.

src/task.c
gapisback: Fixed.

src/splinterdb.c
gapisback: Done, @rtjohnso . Add added unit-tests to cross-check that we are handling all cases correctly.

tests/config.h
gapisback: Done.

tests/unit/task_system_test.c
gapisback: Did not change this as that would percolate to the interface of many other functions that are taking `num_bg_threads[]` as an argument. (Also, this is existing code ... and will probably be reworked in future revs any ways.)

include/splinterdb/splinterdb.h
```diff
@@ -15,9 +15,6 @@
 
 #include "splinterdb/data.h"
 
-// Externalize # of task types that task system can support
-#define SPLINTERDB_NUM_TASK_TYPES 3
```
gapisback: Withdrew this as requested.

include/splinterdb/splinterdb.h
```diff
@@ -325,4 +323,8 @@ splinterdb_stats_print_lookup(const splinterdb *kvs);
 void
 splinterdb_stats_reset(splinterdb *kvs);
 
+// TRUE if Splinter's task system was configured to use background threads.
+bool
+splinterdb_task_system_uses_bg_threads(splinterdb *kvs);
```
gapisback, line 328: Externalized existing boolean fn, to be used as a testing hook.

src/task.c
```diff
@@ -798,7 +833,7 @@ task_system_create(platform_heap_id    hid,
          *system = NULL;
          return rc;
       }
-      nbg_threads_created += num_bg_threads[type];
+      nbg_threads_created += (use_bg_threads ? num_bg_threads[type] : 0);
```
gapisback, line 836: Now that bg-threads are configured only when both config values are non-zero, old code is a bug.

I had to reconcile like so, above, otherwise, we'd run into an endless sleep loop on L840 below ... waiting endlessly for background threads to startup.

include/splinterdb/splinterdb.h
rtjohnso: Why can't someone just get this from the splinter config?

src/task.c
```diff
@@ -474,6 +493,7 @@ task_group_deinit(task_group *group)
 static platform_status
 task_group_init(task_group  *group,
                 task_system *ts,
+                const char  *task_type, // Name of tasks in this group
```
rtjohnso, line 496: The only thing this is used for is to do the printf inside this function.  Move the printf to the caller and delete this parameter.

src/task.c
```diff
@@ -742,6 +742,32 @@ task_perform_one(task_system *ts)
...
+
+   if ((normal_bg_threads + memtable_bg_threads) > (MAX_THREADS / 2)) {
+      platform_error_log("Background thread configuration is not supported."
+                         " Total number of background threads configured"
+                         ", normal-bg-threads=%lu, memtable-bg-threads=%lu,"
+                         " must be <= %d\n",
+                         normal_bg_threads,
+                         memtable_bg_threads,
+                         (MAX_THREADS / 2));
+      return STATUS_BAD_PARAM;
```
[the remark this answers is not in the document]
rtjohnso: I'm not lovin' the arbitrary limit.  Also, why not check that the thread counts are either both 0 or both non-zero?

src/task.c
```diff
@@ -748,14 +802,28 @@ task_system_create(platform_heap_id    hid,
    // task initialization
    register_init_tid_hook();
 
+   // Treat this as a soft-error and configure the task system w/o
+   // background threads when one of these configs is 0.
+   bool use_bg_threads = ((num_bg_threads[TASK_TYPE_MEMTABLE] != 0)
```
rtjohnso, line 807: I would prefer this to be a hard error.

src/task.c
```diff
@@ -748,14 +802,28 @@ task_system_create(platform_heap_id    hid,
...
+                          && (num_bg_threads[TASK_TYPE_NORMAL] != 0));
+
    ts->use_bg_threads = use_bg_threads;
    ts->heap_id        = hid;
    ts->scratch_size   = scratch_size;
    ts->init_tid       = INVALID_TID;
 
+   // Ensure that the main thread gets registered and init'ed first before
+   // any background threads are created. (Those may grab their own tids.).
+   task_run_thread_hooks(ts);
```
rtjohnso, line 817: Doesn't this break all the code we've already written where the main thread must call register_thread?

src/task.c
```diff
@@ -765,12 +833,17 @@ task_system_create(platform_heap_id    hid,
...
    }
 
-   task_run_thread_hooks(ts);
-   const threadid tid      = platform_get_tid();
-   ts->thread_scratch[tid] = ts->init_task_scratch;
-
+   // Wait-for all background threads to start-up and register. This is not
+   // really needed right now, but it allows for reliable unit-testing to verify
+   // state of the task system when bg threads are configured.
+   threadid *max_tid = task_system_get_max_tid(ts);
```
rtjohnso, line 842: Either move this into the test or implement it by having each thread increment a counter and looping here until the counter reaches the expected value.

But please no sleeps.

tests/config.h
```diff
@@ -10,7 +10,7 @@
 #ifndef __CONFIG_H
 #define __CONFIG_H
 
-#include "clockcache.h"
+#include "clockcache.h" // Nested #include's task.h and related tokens.
```
rtjohnso, line 13: Why not include "task.h" directly?  Also, clockcache.h should not include task.h.

tests/config.h
```diff
@@ -88,6 +94,7 @@ typedef struct master_config {
    // Test-execution configuration parameters
    uint64 seed;
    uint64 num_inserts;
+   uint64 num_threads;
```
rtjohnso, line 97: This looks like it got replaced by the `num_normal_bg_threads` and `num_memtable_bg_threds` fields.  Delete this one.

include/splinterdb/splinterdb.h
gapisback: @rtjohnso - Test code uses this to cross-check whether the background threads were, indeed, created [or not] for a specific config params for bg-threads.

For example, if you specify normal_bg_threads=1 and memtable_bg_threads=0, with this fix, no background threads are created.

Test code re-confirms this behaviour. See [this test case, test_splinterdb_created_wo_background_threads] in task_system_test.c

src/task.c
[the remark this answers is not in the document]
gapisback: The both are 0 or both are non-zero is already checked in the test-parsing, `config_parse()`  in config.c (See around L359.)

I'm just trying to create some defenses and guard rails to prevent bad usages being attempted or exercised.
I was just trying to provide a mechanism to protect users from bad usages creeping.

This function exists to check cases where someone does `-num-normal-bg-threads 100000`, or 1 million; or some big big number.

The "<= (MAX_THREADS/2)" is arbitrary ... but I just wanted to implement some checks for preventing large bg-thread counts being configured.

I could change this to something on the lines of:

- check that each bg-thread config param is <= MAX_THREADS
- check that sum of both params is <= MAX_THREADS

The problem I see is that even when (sum of two configs) == MAX_THREADS, and we configure those # of bg-threads ... what do you do next when the 1st end-user threads is created? Things won't work at all.

Let me know what you see as being better, and we can revisit ...

src/task.c
gapisback: Actually, there is no change in logic here. I'm just documenting that the main thread has to register itself first.

May be... I'm missing your question [?] ...

src/task.c
gapisback: If you feel strongly about it, I'm ok changing this to a hard-error. And will rework tests accordingly.

@ajhconway -- what's your vote on this point? Should this be a soft-error (degrading to no-bg-threads enabled), or a hard-error (splinter does not come up) ?

src/task.c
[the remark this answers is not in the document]
rosenhouse: > what do you do next when the 1st end-user thread(s) is (are) created? Things won't work at all.

Could this problem instead be addressed by having a check at thread-registration time?

src/task.c
rosenhouse: Does the main thread need to call register_thread?  I thought that was taken care of automatically?  
https://github.com/vmware/splinterdb/blob/83de5e03722d61436ce247e406d87b8fa4913d85/include/splinterdb/splinterdb.h#L108-L109

test.sh
```diff
@@ -510,19 +510,28 @@ function test_make_run_tests() {
...
+
    "$BINDIR"/unit/misc_test
    "$BINDIR"/unit/limitations_test
+
+   # Default execution runs w/ no background threads configured
    "$BINDIR"/unit/task_system_test
+
+   # Just exercise with some combination of background threads to ensure
+   # that basic usage of background thread config params still works.
+   "$BINDIR"/unit/task_system_test --num-normal-bg-threads 4 --num-memtable-bg-threads  2
```
rosenhouse, line 526: I strongly prefer that "unit" tests have sensible default options, rather than requiring command line flags, so that someone can just run the executable by name and get useful coverage & feedback.
Passing flags should be the exceptional case, not the one used by the test script.

I feel less strongly about this for larger system tests.

tests/unit/task_system_test.c
```diff
@@ -349,6 +388,9 @@ exec_one_thread_use_lower_apis(void *arg)
    ASSERT_EQUAL(thread_cfg->active_threads_bitmask,
                 task_bitmask_before_register);
 
+   platform_default_log("active_threads_bitmask=0x%lx\n",
+                        thread_cfg->active_threads_bitmask);
```
rosenhouse, line 392: #494 changed all unit tests to use the `CTEST_LOG_INFO()` macro.  Any reason not to stick with that new pattern?

include/splinterdb/splinterdb.h
rosenhouse: If you remove that "silently configure" bit and instead make it a hard error, would that mean your tests wouldn't require this function?

gapisback: @rosenhouse - This is a testability viewpoint.

If we make it a hard-error when one of the 2 config-params is zero, then, we are [implicitly] assuming that for non-zero config params, that the bg-threads are created.

If I withdraw this lookup function, tests do not have a way of "verifying" that the bg-threads _were_ created as claimed.

In my book, even if I went to the hard-error route, I'd like to retain this interface as it provides a way to externalize the state of Splinter's sub-systems. Using it for testing would be one such immediate consumer. (Splinter, I think, solely lacks externally usable reporting interfaces and I kind see this a baby-step towards improving that situation.)

src/task.c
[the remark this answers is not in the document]
gapisback: We could, ... and maybe the code already does something like that already. Upon inspection, I think what will happen is that we will end up in this assert, which is only an `debug_assert()`.

```
 62 static void
 63 task_init_threadid(task_system *ts)
 64 {

 88 out:
 89    debug_assert(tid != INVALID_TID);
```

I have to verify this with a test-case as to what will happen if someone has consumed all the 63 bits for bg-threads and then a user-thread attempts to register itself.

I feel that kind of lapsed-case scenario (when someone configures all bg-threads) is something we should try to avoid, if possible. Which is what this arbitrary check in this new function is trying to do.

include/splinterdb/splinterdb.h
rosenhouse: Maybe we need to discuss this on zoom, because I think I'm seeing it differently.  A user ought to be able to easily predict when background threads will be created or not, based entirely on the config they passed in.

I think "reporting" or "metrics" interfaces are a good idea in general, I just don't think we should start designing that feature in this PR -- it should begin with a design doc.

src/task.c
gapisback: Yes, @rosenhouse  - the main thread does not have to call register_thread() itself. That happens, somewhat indirectly, when `task_system_create()` executes:

```
780 platform_status
781 task_system_create(platform_heap_id    hid,
[...]
802    // task initialization
803    register_init_tid_hook();
[...]
815    // Ensure that the main thread gets registered and init'ed first before
816    // any background threads are created. (Those may grab their own tids.).
817    task_run_thread_hooks(ts);                           <<<<<<<<<<
818    const threadid tid      = platform_get_tid();
```

Now, I'm not sure if your and Rob's questions are answered. And, if you are expecting / suggesting a rewording in some comments somewhere?

src/task.c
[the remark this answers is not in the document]
rosenhouse: I think it's good to error if we know they will definitely hit an assertion later.  But if not, then let them do it, and then let them see an error on thread_register.  This arbitrary limit seems....arbitrary.

test.sh
gapisback: @rosenhouse - The 'problem' is that this test creates one task-system right up at the top in the `SETUP()` method.

We want to test that task-system creation can be done in both ways: with and w/o bg-threads configured.
I felt that adding these invocations in `test.sh` w/ desired config options to exercise this functionality is not a bad thing. You will anyway get the coverage when you run quick unit-tests.

To do what you ask I will probably have to add a new test sub-case to tear down the task system and recreate it anew w/ diff config params.

If you insist on this point, it will probably mean a rework of this `task_system_test`. 

Let me know what you want based on this response, and I'll rework accordingly, if needed.

rosenhouse: Here's my belief: if the case is worth covering, then it should be covered when running the test without flags.  But I'm curious what others thing too.

include/splinterdb/splinterdb.h
gapisback: Agreed to ... Withdraw this interface as it is in public .h file.

src/task.c
gapisback: Nothing to change here ... leave as-is.

src/task.c
gapisback: Move print call on L528 to caller, and do not add this new param.

src/task.c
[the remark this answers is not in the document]
gapisback: Change this fn to return an error if sum of # of bg-threads > max-#-of threads.

For the user-thread case, raise & report an error if they have already exceeded bg threads and so registering a new user thread will fail.

src/task.c
gapisback: Vote is to make a hard error and check in the newly added validator fn.

test.sh
gapisback: Rework the unit test case to refactor the `task_system_create()` call into a sub-fn that is called from all cases.

tests/config.h
gapisback: Investigate ... source of `.h` comment and clean up as best as you can.

tests/config.h
gapisback: Check if `num_threads` is still needed; withdraw it if not.

include/splinterdb/splinterdb.h
```diff
@@ -67,6 +66,9 @@ typedef struct {
    uint64 max_branches_per_node;
    uint64 use_stats;
    uint64 reclaim_threshold;
+   uint64 num_memtable_bg_threads;
```
gapisback, line 69: Give some comments / guidance explaining how these config params should be set.

gapisback: Memtable bg-threads are latency sensitive. Normal bg-threads are driven for compaction, so I/O centric.

tests/unit/task_system_test.c
gapisback: @rosenhouse - This is left-over code that I did not sync-up with the interfaces of your change. Will revisit all new code and will conform to the new methods.

tests/unit/task_system_test.c
```diff
@@ -95,6 +96,11 @@ CTEST_SETUP(task_system)
    // Do minimal IO config setup, using default IO values.
    master_config master_cfg;
    config_set_defaults(&master_cfg);
+
+   // Expected args to parse: --num-normal-bg-threads, --num-memtable-bg-threads
```
gapisback, line 100: DO not have test code take arguments. Rework the body to have a separate test sub-case which will deal with specialized arguments.

src/splinterdb.c
```diff
@@ -686,3 +692,10 @@ splinterdb_stats_reset(splinterdb *kvs)
 {
    trunk_reset_stats(kvs->spl);
 }
+
+// TRUE if Splinter's task system was configured to use background threads.
+bool
+splinterdb_task_system_uses_bg_threads(splinterdb *kvs)
```
gapisback, line 698: Deleted this new lookup function.

src/task.c
gapisback: Reworked to remove the new param `task_type` to this function.

I added a print msg to the caller when # of bg-threads of a specific type is non-zero.

But from previous discussions, the leaning of the team seems to be not to have too many extraneous messages. I am ok with having this message come out, as it helps me see this creation happening in other integrations, where I troll through the logfile.

If anyone thinks this is still too much and insist that this be removed, I can do that.

src/task.c
gapisback: L836 has been deleted. NO longer needed.

src/task.c
gapisback: Removed this from here and moved the logic to a helper function, `create_task_system_with_bg_threads()` defined in test code. That fn will loop around a bit sleeping till all configured bg-threads are started up.

src/task.c
```diff
@@ -86,7 +91,10 @@ task_init_threadid(task_system *ts)
    }
 
 out:
-   debug_assert(tid != INVALID_TID);
+   platform_assert((tid != INVALID_TID),
```
gapisback, line 94: Now, we will get an error in debug & release builds. Have a new unit-test case that triggers this error.

I am reworking that case slightly to come up with a scenario where bg-threads use up all but-one-slot. Then, we should be able to create one more user thread and the next create should fail. This small test-case dev is under progress, while we speak ...

src/task.c
```diff
@@ -744,25 +755,30 @@ task_perform_one(task_system *ts)
...
+                         "threads, normal_bg_threads (%lu) "
+                         "and memtable_bg_threads (%lu) "
+                         "must be zero or be non-zero.\n",
+                         normal_bg_threads,
+                         memtable_bg_threads);
+      return STATUS_BAD_PARAM;
+   }
+
+   if ((normal_bg_threads + memtable_bg_threads) >= MAX_THREADS) {
+      platform_error_log("Total number of background threads configured"
```
gapisback, line 776: Reworked the checks on 'bad' configuration as discussed in the review.

src/task.c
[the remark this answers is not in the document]
gapisback: @rosenhouse , @rtjohnso , @ajhconway - FYI; the bulk of the comments were here. All of these have been addressed and the change set accordingly amended in this rev.

src/task.c
gapisback: Done, this is now a hard-error.

src/task.c
```diff
@@ -833,17 +845,18 @@ task_system_create(platform_heap_id    hid,
...
+      uint64 nbg_threads = num_bg_threads[type];
+      if (nbg_threads) {
+         platform_default_log("Splinter task system created %lu"
+                              " background thread%sof type '%s'.\n",
+                              nbg_threads,
+                              ((nbg_threads > 1) ? "s " : " "),
+                              task_type_name[type]);
+      }
    }
+   debug_assert((*system == NULL),
```
gapisback, line 857: This is a minor cleanliness / hygiene check. In my unit-tests, where I am destroying and recreating the task-system, I wanted to be sure that the task system interface is not inadvertently clobbering a valid task_system`ts *`. Hence, this assert.

This cascades into changes in many consumer code paths where on-stack `ts *` had to be initialized to NULL, which is a good thing to do anyway.

tests/config.c
```diff
@@ -39,8 +39,8 @@
 
 // By default, background threads are disabled in Splinter task system.
 // Most threads run w/o background threads. Very small # of tests exercise
-// background threads through --num-bg-threads and --num-memtable-bg-threads
-// options.
+// background threads through the --num-normal-bg-threads and
+// --num-memtable-bg-threads options.
```
gapisback, line 43: Corrected parameter names.

tests/unit/limitations_test.c
```diff
@@ -280,32 +285,21 @@ CTEST2(limitations, test_splinterdb_created_wo_background_threads)
...
 }
 
 /*
- * Test that SplinterDB tasks system requires non-zero config for both types
- * of task groups for background threads to be created.
+ * Test that SplinterDB can be created even when background threads use
+ * up all the slots.
  */
-CTEST2(limitations, test_splinterdb_created_w_background_threads)
+CTEST2(limitations, test_splinterdb_created_w_all_background_threads)
```
gapisback, line 302: I am reworking this newly added case slightly to create 62 bg-threads. Then, test case will show that we can create one more user thread but the next user-thread creation should fail. (Am debugging that logic right now... )

tests/unit/task_system_test.c
```diff
@@ -174,21 +158,13 @@ CTEST_TEARDOWN(task_system)
...
 CTEST2(task_system, test_basic_create_destroy)
 {
    threadid main_thread_idx = platform_get_tid();
    ASSERT_EQUAL(main_thread_idx, 0);
-   if (data->num_bg_threads_configured) {
-      ASSERT_TRUE(task_system_use_bg_threads(data->tasks));
-   } else {
-      ASSERT_FALSE(task_system_use_bg_threads(data->tasks));
-   }
+   ASSERT_FALSE(task_system_use_bg_threads(data->tasks));
```
gapisback, line 167: By default, all cases in this test will create task system w/o bg-threads.

tests/unit/task_system_test.c
```diff
@@ -330,17 +306,13 @@ CTEST2(task_system, test_multiple_threads)
 
    ZERO_ARRAY(thread_cfg);
    ASSERT_EQUAL(task_get_max_tid(data->tasks),
-                (1 + data->num_bg_threads_configured),
+                1,
```
gapisback, line 309: No more bg-threads can be created with command line args, hence, this diff.

src/splinterdb.c
rosenhouse: I still see it in the diff here.

include/splinterdb/splinterdb.h
rosenhouse: A couple small suggestions
- Note that it is an error to for one to be zero while the other is non-zero
- rather than "affect" latency of inserts, additional memtable background threads can "improve" insert latencies, or maybe more specifically "reduce insert tail latencies."
- Not sure if this really matters to app devs, but "normal" background threads work on compacting branches _and building filters_

tests/config.h
rosenhouse: @gapisback were you going to address this before merging?

tests/unit/limitations_test.c
```diff
@@ -245,6 +245,78 @@ CTEST2(limitations, test_io_init_invalid_extent_size)
...
+
+   // Task system should be setup with background threads
+   cfg.num_normal_bg_threads   = 1;
+   cfg.num_memtable_bg_threads = 1;
+
+   int rc = splinterdb_create(&cfg, &kvsb);
+   ASSERT_EQUAL(0, rc);
+
+   splinterdb_close(&kvsb);
+}
```
rosenhouse, line 296: This test case is a bit odd to me.
The rest of the file is testing "limitations" (almost all being error conditions).
The comment here says it "requires non-zero config for both types of task groups" but that sounds pretty low-level.  It seems like it's just proving that you can initialize a splinterdb with background threads.  And I sort of wonder whether this test maybe instead belongs in `splinterdb_quick_test.c` or `splinterdb_stress_test.c` as a simple "smoke" test that does this setup w/ background threads, and then does a few (300,000?) inserts to trigger at least a memtable incorporation, and then quits?   That might be a nice small demo of the happy-path for this new public API surface.

tests/unit/task_system_test.c
```diff
@@ -133,16 +130,25 @@ CTEST_SETUP(task_system)
...
+                task_bitmask,
+                "exp_bitmask=0x%x, task_bitmask=0x%x\n",
+                exp_bitmask,
+                task_bitmask);
 
    // Save it off, so it can be used for verification in a test case.
    data->active_threads_bitmask = exp_bitmask;
+   if (Ctest_verbose) {
+      platform_set_log_streams(stdout, stderr);
+   }
```
rosenhouse, line 144: Thanks for adding this here.  In the other unit test files that have it, I've put it at the top of `CTEST_SETUP()`, in part so that the users's choice for "verbose" is respected even for test-setup outputs.

gapisback:
@rosenhouse - thanks for your comments. I'm working on amended test case and will go thru your comments in a bit. Will address left-over mop in a bit. Hopefully I hear back from the others soon, so I can pull-in all delta in one pass.

tests/unit/limitations_test.c
rosenhouse: You have a fair point that we can't actually verify that any background work is happening here.

Perhaps the issue for me here is that this "limitations test" file is awkward because it's mixing concerns and creating ambiguity about where test cases ought to live.  Limitations test is trying to focus on error-cases, mostly, but it's doing so for several different components, including both low-level things (task system, io system) and for high-level API (splinterdb.h). So multiple components are mixed in here.  And tests for a single component now span multiple unit-test files.  Further, the presence of an error-case here in limitations is then being used to justify putting an "opposite" happy-path case right below it, also in "limitations test".  So there's a question of whether a happy-path case belongs in this file or in the file that holds all the other happy-path cases for that API.

This doesn't need to be fixed in this PR, but I'm starting to think that this "limitations" file ought to be broken up, and the test cases in it should be moved to the component-specific unit test files where their happy-path cases already live.  That way, all the tests for a component live together in one place, regardless of whether it's covering a happy path or an error path.

This doesn't need to be fixed in this PR.  But I think we ought to move in that direction in the future, perhaps next time we find ourself discussing whether a test-case belongs in `limitations_test.c` or elsewhere.

src/task.c
```diff
@@ -380,7 +384,8 @@ task_thread_create(const char            *name,
    ret = task_create_thread_with_hooks(
       &thr, FALSE, func, arg, scratch_size, ts, hid);
    if (!SUCCESS(ret)) {
-      platform_error_log("Could not create a thread\n");
+      platform_error_log("Could not create a thread: %s\n",
+                         platform_status_to_string(ret));
```
gapisback, line 388: Added `platform_status_to_string()` to decode ret value. In the newly amended failing test case, you will now see this message:

```
Could not create a thread: Resource temporarily unavailable
```

tests/unit/task_system_test.c
```diff
@@ -373,39 +386,73 @@ CTEST2(task_system, test_task_system_creation_with_bg_threads)
 /*
  * ------------------------------------------------------------------------
  * Test creation of task system using up the threads for background threads.
- * Verify that the next user-thread creation should fail.
+ * Verify ththe we can create just one more user-thread and that the next
+ * user-thread creation should fail with a proper error message.
  * ------------------------------------------------------------------------
  */
-CTEST2(task_system, test_use_all_threads_for_bg_threads)
+CTEST2(task_system, test_use_all_but_one_threads_for_bg_threads)
```
gapisback, line 393: Modified this previously added test case to go the extra distance of showing that you can still create a user thread when there is one more slot free. And that the next thread creation will fail once we run out of the limit.

tests/unit/task_system_test.c
```diff
@@ -373,39 +386,73 @@ CTEST2(task_system, test_task_system_creation_with_bg_threads)
...
+   thread_cfg[1].exp_thread_idx = task_get_max_tid(data->tasks);
+
+   // We've used up all threads. This thread creation should fail.
+   rc = task_thread_create("test_one_thread",
+                           exec_user_thread_loop_for_stop,
+                           &thread_cfg[1],
+                           trunk_get_scratch_size(),
+                           data->tasks,
+                           data->hid,
+                           &new_thread[1]);
```
gapisback, line 440: The failure output from here looks like:

```
sdb-fdb-build:[48] $ build/debug/bin/unit/task_system_test test_use_all_but_one_threads_for_bg_threads
Running 1 CTests, suite name 'task_system', test case 'test_use_all_but_one_threads_for_bg_threads'.
TEST 1/1 task_system:test_use_all_but_one_threads_for_bg_threads Cannot create a new thread as the limit on concurrent threads, 64, will be exceeded.
Could not create a thread: Resource temporarily unavailable
[OK]
```

gapisback:
OK, I am now done with my delta test-case update. Uploaded a new commit with changes to introduce `test_use_all_but_one_threads_for_bg_threads` test case.

Have seen @rosenhouse 's comments. Am awaiting comments from @rtjohnso and @ajhconway on this final set, and then will address all comments in one (hopefully) final pass.

test.sh
gapisback: Done, in 2 newly added sub-test-cases.

src/task.c
```diff
@@ -765,12 +833,17 @@ task_system_create(platform_heap_id    hid,
...
-   task_run_thread_hooks(ts);
-   const threadid tid      = platform_get_tid();
-   ts->thread_scratch[tid] = ts->init_task_scratch;
-
+   // Wait-for all background threads to start-up and register. This is not
+   // really needed right now, but it allows for reliable unit-testing to verify
+   // state of the task system when bg threads are configured.
+   threadid *max_tid = task_system_get_max_tid(ts);
+   while (*max_tid < nbg_threads_created) {
+      platform_sleep(USEC_TO_NSEC(100000)); // 100 msec.
```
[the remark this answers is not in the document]
gapisback: Implemented (a)

tests/unit/task_system_test.c
gapisback: Reworked unit-test cases accordingly.

src/splinterdb.c
gapisback: Yes, my omission. Thanks, for catching this ... Deleted.

src/task.c
```diff
@@ -70,6 +70,11 @@ task_init_threadid(task_system *ts)
       uint64 tmp_bitmask = *tid_bitmask;
       // first bit set to 1 starting from LSB.
       uint64 pos = __builtin_ffsl(tmp_bitmask);
+
+      // If all threads are in-use, bitmask will be all 0s.
+      if (pos == 0) {
```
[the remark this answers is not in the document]
gapisback: @rtjohnso -- Heads-up! You may want to closely review this diff. 

I verified thru a standalone C-program that `__builtin_ffsl()` returns 0 if bitmask is 0x00. So, I think a few lines below, this is probably wrong too:

```
  78       // builtin_ffsl returns the position plus 1.
  79       tid = pos - 1;
```

The check added on L75 and bailout `goto out` seems the right thing to.

Would be nice to cross-check this subtle logic ... just in case I'm still making a mistake.

src/task.c
gapisback: Update: Uploaded new commit with enhanced test case `test_use_all_but_one_threads_for_bg_threads()`.

src/task.c
gapisback: With the previous return of `STATUS_LIMIT_EXCEEDED` (which is mapped to `CONST_STATUS(ENOSPC)`, we would get a message `No space left on device` that is quite misleading

This new message is a bit more accurate as the limit on all threads in-use could go away as soon as one or more threads exit.

tests/unit/task_system_test.c
gapisback: Done.

tests/unit/limitations_test.c
gapisback: @rosenhouse - noted your concerns about +ve / -ve tests being mixed-up in this limitations_test file. 

For what it's worth, I moved these two cases to `task_system_test.c`:

- test_splinterdb_create_w_background_threads()
- test_splinterdb_create_w_all_background_threads()

Hope this is acceptable to you.

include/splinterdb/splinterdb.h
gapisback: Wordsmith'ed this a bit more.

tests/config.h
gapisback: `num-threads` was added for a new test case I had developed, `large_inserts_stress_test`, but that test has now been withdrawn.

This config param was left behind. I've deleted it now.

And gently cleaned up some definitions in `config.c` for mnemonics related to bg-threads.

tests/config.h
gapisback: @rosenhouse - I did not try to 'resolve' these nested `.h` inclusions. This comment was needlessly drawing attention and I took it out.

(A separate issue will need to be logged to clean up these nested includes. I'm not following up on that.)

src/task.c
rtjohnso: Now that I see that this function was always called, I think you moved it to a better place.  LGTM.

src/task.c
```diff
@@ -765,12 +850,18 @@ task_system_create(platform_heap_id    hid,
          *system = NULL;
          return rc;
       }
+      uint64 nbg_threads = num_bg_threads[type];
+      if (nbg_threads) {
+         platform_default_log("Splinter task system created %lu"
```
rtjohnso, line 855: I don't think we need to be noisy about this.  Other opinions?

tests/config.h
rtjohnso: I still see it?

tests/functional/splinter_test.c
```diff
@@ -2383,7 +2383,9 @@ usage(const char *argv0)
       "\t%s --cache-per-table\n"
       "\t%s --parallel-perf --max-async-inflight [num] --num-pthreads [num] "
       "--lookup-positive-percent [num] --seed [num]\n"
-      "\t%s --num-bg-threads (number of background threads)\n"
+      "\t%s --num-normal-bg-threads (number of normal background threads)\n"
```
rtjohnso, line 2386: I believe this was an old `splinter_test`-specific parameter that is now superceded by your better and more general handling in `config.[hc]`.  So I think this should go away entirely.

tests/functional/splinter_test.c
```diff
@@ -2612,8 +2614,9 @@ splinter_test(int argc, char *argv[])
       config_argv += 1;
    }
    if (config_argc > 0
-       && strncmp(
-             config_argv[0], "--num-bg-threads", sizeof("--num-bg-threads"))
+       && strncmp(config_argv[0],
```
rtjohnso, line 2617: Should be handled by `config.[hc]` parser.

tests/functional/splinter_test.c
```diff
@@ -2627,7 +2630,7 @@ splinter_test(int argc, char *argv[])
    if (config_argc > 0
        && strncmp(config_argv[0],
                   "--num-memtable-bg-threads",
-                  sizeof("--num-bg-threads"))
+                  sizeof("--num-memtable-bg-threads"))
```
rtjohnso, line 2633: Ditto.

rtjohnso (changes requested):
Mostly minor comments, but a couple of things that need to be fixed.

src/task.c
```diff
@@ -320,7 +328,15 @@ task_create_thread_with_hooks(platform_thread       *thread,
                               platform_heap_id       hid)
 {
    platform_status ret;
-   thread_invoke  *thread_to_create = TYPED_ZALLOC(hid, thread_to_create);
+   uint64         *tid_bitmask = task_system_get_tid_bitmask(ts);
+   if (*tid_bitmask == 0) {
+      platform_error_log("Cannot create a new thread as the limit on"
```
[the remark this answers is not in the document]
gapisback: @rtjohnso - I fully realize that this is racy. But trying to rework this aspect also in this enablement PR will bloat the scope. The error checks I've added now certainly improve the situation. I think it will be better to peel this off as a follow-on issue. 

What if I open a new issue to track this item?

src/task.c
gapisback: I think it's `>=` as the main thread uses slot 0. So bg-threads together can only come up to 63.

tests/functional/splinter_test.c
gapisback: I'm not sure I understand. `num-bg-threads` is being renamed to `num-normal-bg-threads` and the spelling error in the 2nd one's name is being fixed.

Not sure how all this can go away ... We should talk this thru.

tests/functional/splinter_test.c
gapisback: @rtjohnso - That seems like a general comment, no? Or are you suggesting that I take on this comment and implement it in this PR?

We've talked several times in PR reviews that this config stuff should be overhauled, majorly. Please log a new issue, if one does not already exist ... but I am not sure that overhaul belongs to this enablement PR.

tests/functional/splinter_test.c
rtjohnso: Look at your changes to config.c.  You added support for these command-line parameters to that file.

But this file used to have its own custom command-line parameter for controling bg threads.

Now that you've added support for bg-thread command-line parameters to config.c, you should remove the old special-case code from splinter_test.c and use the general code in config.c instead.

tests/functional/splinter_test.c
rtjohnso: See response above.

tests/functional/splinter_test.c
gapisback: @rtjohnso - I thought about your suggestion a bit more, and here is what I _think_ you are suggesting to do:

- Replace the inline code in `splinter_test()` with a call to `config_set_defaults(&master_cfg)`, followed by parsing cmd-line args into `master_cfg` using `config_parse()`.
- Then, pull out the bg-threads related values parsed into `master_cfg`, implementing something like `init_bg_threads_config_from_master_config()` [ like you've done elsewhere with `init_btree_config_from_master_config()` ]

Right? Well ... this won't quite work because after parsing these bg-thread args, `splinter_test()` does:

```
2646    if (splinter_test_parse_perf_args(&config_argv,
2647                                      &config_argc,
2648                                      &max_async_inflight,
...
```

I.e., we want to further parse more cmd-line arguments.

There are two problems with calling `config_parse()`:

1. It will eat up the bg-threads related config params, but will not advance argc / argv in the caller's context. We need a way to have `config_parse` advance `&config_argv, &config_argc` so that the subsequent call to `splinter_test_parse_perf_args()` is positioned correctly at the right args.
2. Calling `config_parse()` will mean caller can invoke with any other config params unrelated to bg-threads configuration. Those will be un-handled by this test, and will be misleading to the user. ( I see that this can be mitigated somewhat by a writing a new fn `init_bg_thread_config_from_master_config()`, but that is somewhat of a sub-optimal solution. )

For these reasons I did not implement your suggestion.

----
However, upon re-examination your comment did point out a flaw in my delta-commits.

In previous revs I did have one of the sub-cases in `unit/task_system_test.c` calling `config_parse()` to verify that the cmd-line arg parsing for bg-threads is working correctly.

As requested by @rosenhouse, and you agreed to that too, I've withdrawn that sequence and have inlined new test cases. That "lost" the code coverage for the parsing and error checking of bg-threads params that was added to `config _parse()`.

As it seems reasonable that we should go ahead with having bg-threads related in `master_config{}`, I'd like to retain the changes to `config_parse()`.  To verify that the right parsing is going on, as all this is not tested elsewhere, I added these small unit-test cases to `unit/config_parse_test.c`:

1. test_bg_threads_parse_configs_valid
2. test_bg_threads_parse_configs_invalid

These sub-cases do not require extra cmd-line args when the test is invoked and will provide required test coverage.

(Btw, general response is that all this config parsing stuff has to be overhauled separately through table driven logic. But that's not something I can take on right now.)

include/splinterdb/splinterdb.h
gapisback: Reworked this suggested rewrite to fold it into what was there previously. Uploaded in latest commit coming up ...

gapisback:
HI, @rtjohnso - I've gone through your comments, basically around changes in `config.c` and have responded in [this link](https://github.com/vmware/splinterdb/pull/469/files#r1045269912).

Have also pulled-in verbiage from your feedback on how to describe bg-threads config params.

@rosenhouse , @ajhconway & Rob -- If all your issues are addressed, kindly take a final look and pass this review so I can integrate this to main.

Any other / further changes -- it's best to talk things through on Tue's PR meeting as, otherwise, this exchange thru UI comment / response is becoming unmanageable for me.

gapisback:
Hi, @rtjohnso @ajhconway @rosenhouse: Here is the updated status of code review comments / feedback / follow-on items for this PR: 

1. Have addressed all review feedback and responded with inline comments. CI runs are now clean.
2. Please see [this slack post](https://app.slack.com/client/T024JFTN4/G01MEDQ05MX/thread/G01MEDQ05MX-1670797035.567259) with my analyses on what it takes to re/solve the race condition Rob pointed out during tid-initialization. I think this issue needs to be further discussed and is best peeled off as a follow-on.
3. Opened issue #499 
4. Opened issue #500 

I think this covers all points raised during the review and off-line debugging /discussions. Let me know if I'm still missing on some mop-up items.

Looking forward to your final reviews of this work so I can start work to integrate this to /main.

tests/functional/splinter_test.c
rtjohnso: Here is the problem with the arg parsing as it currently stands.  I'm pretty sure this is a real bug in the way you've changed things.

In the old code, `splinter_test.c` parsed its own command-line parameters, including --num-bg-threads, and then it called `test_config_parse` to parse the remaining parameters that are handled by `config.c`.

You have updated the `splinter_test.c` custom argument parsing to support two command line parameters, --num-normal-bg-threads and --num-memtable-bg-threads.

But you have also updated `config.c` to support the exact same command line parameters.

So now there are two difference pieces of code that could handle those command line parameters when someone runs splinter_test.  If the user puts --num-memtable-bg-threads or --num-normal-bg-threads early in their list of command line args, those arguments will get handled on lines 2617-2644 of splinter_test.c.  If they put those command line parameters later in the command-line arg list, those parameters will get processed by config.c

Now, when splinter_test.c initializes the task system on line 2795, it uses the locally declared array `num_bg_threads` (declared on line 2509).  This variable is set only by the command-line parsing code in splinter_test.c.  It is not configured by the command-line parsing code in config.c.

Thus, depending on where the user puts these command-line options, they could get silently ignored by splinter_test.c.

Hence my requested code change: delete the custom parsing code for those command-line args in splinter_test.c and convert the task_system initialization in splinter_test.c to use the values obtained from the config.c command-line-parsing code.

gapisback: Hmmm ... let me look at it some more later tomorrow. May be I did miss some double-parsing issue.

gapisback: @rtjohnso - I have reworked the code under [this recently uploaded commit](https://github.com/vmware/splinterdb/pull/469/commits/6e2b03cf43e330bebf383f4422f5e08e264c3e9f) to address this buggy behaviour.

gapisback:
HI, @rtjohnso @ajhconway @rosenhouse : I just uploaded a new commit to address the remaining outstanding issue raised by Rob in [this feedback](https://github.com/vmware/splinterdb/pull/469/files#r1045390279).

The commit you need to look at now is [this one](https://github.com/vmware/splinterdb/pull/469/commits/6e2b03cf43e330bebf383f4422f5e08e264c3e9f).

If I recall right, Rob, this is the last point you'd like to see fixed.

Gabriel -- I have reworked tests / unit-tests etc. to not require invoking them with --cmd-line-args. The commit that has this reworked code is [this one](https://github.com/vmware/splinterdb/pull/469/commits/9abf437c68a24a3eb24ad2921eb6ae4bc5188ee8). You may want to peruse those diffs to see if I've addressed your issues to your satisfaction.

Alex -- I don't think you had any major rework-needed-items. A minor rework is in the last commit, I took the opportunity to change `uint8 num_bg_threads[]` and related fn-params to consistently define them as `uint64`. That was a nag waiting to be fixed.

Let's try to cover these delta-diffs in the 1pm PR meeting and try to bring this to a closure.

rtjohnso (approved):
LGTM

tests/unit/task_system_test.c
```diff
@@ -326,6 +356,222 @@ CTEST2(task_system, test_multiple_threads)
...
+   data_config       default_data_cfg;
+
+   default_data_config_init(TEST_MAX_KEY_SIZE, &default_data_cfg);
+   create_default_cfg(&cfg, &default_data_cfg);
+
+   // Task system should be setup with all background threads
+   cfg.num_normal_bg_threads   = (MAX_THREADS - 2);
+   cfg.num_memtable_bg_threads = 1;
+
+   int rv = splinterdb_create(&cfg, &kvsb);
```
rosenhouse, line 508: Same comment as above.  These tests only need the public API.  Having them here creates confusion about which units should be covered by which unit tests.

tests/unit/task_system_test.c
rosenhouse: This output shows up even when `VERBOSE` is unset. 

 In #494 I did some work to hide such "error messages from successful tests" unless the `VERBOSE` flag is set.

Please add something like this code below, either into the test case or the file's `CTEST_SETUP` function, so that the `VERBOSE` flag is respected:

https://github.com/vmware/splinterdb/blob/83de5e03722d61436ce247e406d87b8fa4913d85/tests/unit/limitations_test.c#L68-L82

rosenhouse (changes requested):
I still have a couple small notes on unit tests.

1. `test_use_all_but_one_threads_for_bg_threads` ought to respect the `VERBOSE` env var, following the pattern in #494 
2. The tests covering the public API (`test_splinterdb_create_w_background_threads` and `test_splinterdb_create_w_all_background_threads`)  should live alongside other tests of the public API, in the `splinterdb_quick_test.c` file.

Thanks.

tests/unit/task_system_test.c
```diff
@@ -326,6 +356,222 @@ CTEST2(task_system, test_multiple_threads)
...
+   data_config       default_data_cfg;
+
+   default_data_config_init(TEST_MAX_KEY_SIZE, &default_data_cfg);
+   create_default_cfg(&cfg, &default_data_cfg);
+
+   // Task system should be setup with background threads
+   cfg.num_normal_bg_threads   = 1;
+   cfg.num_memtable_bg_threads = 1;
+
+   int rv = splinterdb_create(&cfg, &kvsb);
```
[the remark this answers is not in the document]
gapisback: Relocated these two test cases to `splinterdb_quick_test.c`.

tests/unit/task_system_test.c
gapisback: Relocated these two test cases to `splinterdb_quick_test.c`

gapisback: Also pulled in the suggested code in next comment to start of CTEST_SETUP() as suggested.

gapisback:
HI, @rosenhouse & @ajhconway - Rob has passed this review.

The last set of changes to address your feedback has been integrated. Please do clear this review before the end of your work day, so I can cleanup and check this into /main. Thanks!

rosenhouse (approved):
Thanks Aditya!

gapisback:
FYI: @ajhconway - I have rebased v/s /main, picked up a new commit, squashed all my work to one commit and cleaned up the commit message.

If I don't hear from you by my Wed a.m. ... I'd like to push in to /main, as we've already discussed the changes in the PR meeting.

PR #470, opened 2022-11-06
Fix missing initialization of rc_allocator_in_use_virtual fn in rc_allocator_ops{} virtual functions table.

gapisback:
In a separate dev-effort, while running large inserts stress tests, I ran into a signal as reported in this issue.

It's a mystery why we are not seeing this signal in other performance benchmarking efforts.

Several efforts were made to develop `large_inserts_stress_tests` to reproduce this issue, but to no avail.

The missing virtual function pointer is a clear bug from code-inspection.

Hence, I am pushing this change to a PR for integration to /main.


ajhconway (approved):
Ship it

Issue #462, opened 2022-10-22
Extend `platform_assert()` to prepend OS-pid and Splinter Thread-ID (index) for failing assertion.


gapisback:
### Is your feature request related to a problem? Please describe.

When assertion trips, and / or we run into a subsequent signal, it will be useful to also see the OS-Pid and Splinter thread-ID as part of the assertion message.

- OS-Pid is useful to correlate the PID of the failing assertion in case we want to debug core files generated for a PID.
- Thread-ID is useful to know which thread, given by `platform_get_tid()`, ran into the assertion.

One, while debugging, has to otherwise extend instance of each `platform_assert()` which fails to prepend this extra info.

By always reporting these two pieces, it will improve overall debugging-ability.

### Describe the solution you'd like

Extend `platform_assert()` to report something on the lines of:

```
S-pid=1779382, Thread-ID=0, Assertion failed at src/rc_allocator.c:554:rc_all ...
```

### Describe alternatives you've considered

Instrument / change the code of each assertion one is debugging to add this extra-info. This is unsustainable and requires code changes each time we need this info while debugging a new failing assertion.

PR #471, opened 2022-11-06
(#462) Prepend OS-pid and platform thread-ID to assert message.

gapisback:
To better debug assertion failures, this commit extends the assert message to prepend the OS-pid and thread-ID (index) of the running Splinter thread that encountered the assertion. 

A sample output from this amended assertion message looks like:

```
OS-pid=2235, Thread-ID=0, Assertion failed at tests/unit/misc_test.c:110:...
```

Couple of existing test cases are mildly amended to cope with these new fields in the assert message printed.
Existing debug_assert() also gains this functionality as it's a wrapper around platform_assert().

Currently, seeing OS-pid may not be helpful as we are running in single-process multiple-thread execution model.
In future, when we support multiple processes connecting to Splinter, this extra `OS-pid` information will be very helpful to debug signals / assertions / failures.


rtjohnso (approved):
LGTM

gapisback:
@rosenhouse , Rob @rtjohnso , has passed this fix. It's fairly straightforward but I wanted your opinion too.

What we have now with this fix is a bit verbose. I had a re-think, even while doing this work, that probably the new fields should be output in a way that they are more easily programmatically accessed.

So rather than:

(1) ```OS-pid=%d, Thread-ID=%lu, Assertion ...```

Should we go for ':'-separated fields for the two number fields, like so:?

(2) ```%d:%lu:Assertion ...```

I like the verbose output as it's self-documenting, however, there is a valid reason to go to the alternate format.

No one is parsing assertion messages right now, so probably current verbose output is acceptable.
And as this **is a message** from a failed assertion, I thought a verbose (1) is preferable.

Do you have a strong preference for (1) or (2)? If not, I'd like to stay with (1).

src/platform_linux/platform.c
```diff
@@ -356,8 +357,15 @@ platform_assert_msg(platform_log_handle *log_handle,
...
                     va_list              varargs)
 {
-   static char assert_msg_fmt[] = "Assertion failed at %s:%d:%s(): \"%s\". ";
-   platform_log(
-      log_handle, assert_msg_fmt, filename, linenumber, functionname, expr);
+   static char assert_msg_fmt[] =
+      "OS-pid=%d, Thread-ID=%lu, Assertion failed at %s:%d:%s(): \"%s\". ";
+   platform_log(log_handle,
+                assert_msg_fmt,
+                getpid(),
```
rosenhouse, line 364: Maybe use `gettid()` instead?

https://man7.org/linux/man-pages/man2/gettid.2.html

> In a single-threaded process, the thread ID is equal to the process ID (PID, as returned by `getpid(2)`).  In a multithreaded process, all threads have the same PID, but each one has a unique TID.

I'd say that's strictly more information, and it makes it useful even for multi-threaded-single-process applications too.   It also better matches the output of tools I'm familiar with, e.g. `htop` will print the `tid`.

In which case, maybe update the format string to say `OS-tid` and `Splinter-tid` or something like that...?

rosenhouse (commented):
Seems like a good idea.   I have a suggestion for making it a bit more useful.

src/platform_linux/platform.c
gapisback: @rosenhouse - I am fine extending this to also report `gettid()` but think that OS-pid is also useful on its own. (Even if it will be the same as gettid for the single-threaded process case.)

The reason I say this is that in core files produced, OS-pid is registered. And I think it's expected that we crack open core files using, something like, `gdb core -p <pid>`.

Also, if the same field is sometimes `pid` and sometimes `gettid`, then it will be confusing to remember which case we are dealing with.

I'd prefer 'Thread-ID' for the splinter-tid, as that is what it is referred to consistently inside our code base.

How about we settle on:

`OS-pid=%d, OS-tid=%lu, Thread-ID=%lu, Assertion ...`

That should satisfy all users / usages.

rosenhouse: sounds good

gapisback: @rosenhouse -- I made this change we agreed to.

I did not add any new test to verify that the new `tid` value for a thread will be different than the one for the main program. That seemed a bit overboard and unnecessary.

CI is running now ... If you can clear this review, I'll merge this when CI passes.

rosenhouse (approved):

Issue #463, opened 2022-10-23
Couple of incorrect usages of `platform_heap_id hid` in BTree unit-tests: Could potentially lead to ASAN failures.


gapisback:
### Describe the bug

During some prototype development of other features that exercise `platform_heap_id`, discovered following two instances of code snippets that are potentially flawed.

1. unit/btree_test.c: In `leaf_split_tests()`, use of `scratch` arg as `hid` parameter to `btree_leaf_incorporate_tuple()` is potentially flawed. Way deep down in writable buffer code, we mistakenly think that this 'scratch' space handle is the heap-handle. And code flows into attempts to free memory from this handle. Currently, nothing is majorly wrong, but instrumentation to track allocs / frees are tripping up messages, indicating that this call-site is the source of the aberrant usage.

```
416       bool success = btree_leaf_incorporate_tuple(
417          cfg, scratch, hdr, key, bigger_msg, &spec, &generation);
418       if (success) {
```

2. unit/btree_stress_test.c: In this test, the `insert_thread_params{} ` struct has two instances of heap_id variables:

```
 29 typedef struct insert_thread_params {
 30    cache           *cc;
 31    btree_config    *cfg;
 32    platform_heap_id heap_id;
[...]
 38    platform_heap_id hid;
 39 } insert_thread_params;
```

This causes code to be confused as to which one to use. E.g., here in `insert_tests()` function we mistakenly use both handles; one to alloc and one to free.

```
312    uint8 *keybuf      = TYPED_MANUAL_MALLOC(hid, keybuf, keybuf_size);
313    uint8 *msgbuf      = TYPED_MANUAL_MALLOC(hid, msgbuf, msgbuf_size);
314
[...]
330    platform_free(heap_id, keybuf);
331    platform_free(heap_id, msgbuf);
```

Currently it's not a real issue as both heap_ids are NULL, so we pretty much end up using `malloc` and `free`.

In some future prototype work, this mixed-up usages of `platform_heap_id` is causing ASAN runs to flag memory allocation errors.

### Reproduction steps

1. Nothing can be seen in /main with ASAN runs. These are latent issues which will surface if we ever change heap-memory handling.
2.
3.
...

### Expected behavior

1. Potential fix is to use `hid` rather than `scratch` when invoking said function.
2. Potential fix is to purge `heap_id` variable from `insert_thread_params{}` and consistently use only `hid` variable.

gapisback:
[...]

PR #472, opened 2022-11-07
(#463) Fix incorrect usages of heap-ID in BTree unit tests.

gapisback:
This commit cleans up some potentially erroneous usages of platform_heap_id in BTree unit tests. 

None of these are truly problematic at this time as heap_id is always NULL, so we end up using malloc(). However, these incorrect usages tripped up failures in on-going development which added some memory-tracking instrumentation.

These are latent issues which will surface if we ever change heap-memory handling.


tests/unit/btree_stress_test.c
```diff
@@ -29,13 +29,12 @@
...
    cache           *cc;
    btree_config    *cfg;
-   platform_heap_id heap_id;
+   platform_heap_id hid;
    btree_scratch   *scratch;
    mini_allocator  *mini;
    uint64           root_addr;
    int              start;
    int              end;
-   platform_heap_id hid;
```
gapisback, line 38: There are two instances of this variable, causing confusion below -- we ended up using one for allocation and one for free.

(This is probably a bug that sneaked-in when I initially ported these tests to C-Unit framework.)

tests/unit/btree_stress_test.c
```diff
@@ -45,13 +44,12 @@ insert_thread(void *arg);
...
 insert_tests(cache           *cc,
              btree_config    *cfg,
-             platform_heap_id heap_id,
+             platform_heap_id hid,
              btree_scratch   *scratch,
              mini_allocator  *mini,
              uint64           root_addr,
              int              start,
-             int              end,
-             platform_heap_id hid);
```
gapisback, line 54: Strangely, the same bug crept into this interface, too. Delete this arg, retain the `hid` arg on L47 above.

tests/unit/btree_stress_test.c
```diff
@@ -327,8 +323,8 @@ insert_tests(cache           *cc,
          ASSERT_TRUE(FALSE, "Failed to insert 4-byte %ld\n", i);
       }
    }
-   platform_free(heap_id, keybuf);
-   platform_free(heap_id, msgbuf);
+   platform_free(hid, keybuf);
```
gapisback, line 326: In the old code (not seen in this diff block), allocation was done using `hid` but free is being done using `heap_id`.

For current scheme in /main, this is a non-issue. But while supporting new under-development feature for different memory models, this mixed usage becomes problematic. It could lead to memory leaks, too!

tests/unit/btree_test.c
```diff
@@ -414,7 +414,7 @@ leaf_split_tests(btree_config    *cfg,
       slice key = slice_create(1, &i);
 
       bool success = btree_leaf_incorporate_tuple(
-         cfg, scratch, hdr, key, bigger_msg, &spec, &generation);
+         cfg, hid, hdr, key, bigger_msg, &spec, &generation);
```
gapisback, line 417: The reasons behind this fix has been aired out in internal [slack thread, here](https://vmware.slack.com/archives/G01MEDQ05MX/p1665524515654859?thread_ts=1665524450.401499&cid=G01MEDQ05MX).

Quoting from that response, here is my reasoning for this fix:

----
scratch is btree_scratch   *scratch declared as on-stack scratch buffers in CTEST_DATA().

On L445, passing down scratch as hid is benign now, as nothing is done to free this memory in shmem code-path.

By default, w/o using shared memory, in platform_free_from_heap() we will naturally fall through to free().

But in that function, due to above usage, because hid is non-NULL, we end up going to splinter_shm_free() which does nothing currently.

I just had enabled tracing in platform_free() to see who's calling it, and found that  `writable_buffer_deinit()` -> `platform_free()` trips that trace message that something is being freed from shared-memory, even though test was run without use of --use-shmem flag.

Walking back ... I found this location on L445 which seems problematic. (edited)

rtjohnso (approved):
Thank you for fixing these embarrassing bugs!

PR #477, opened 2022-11-18
make compaction failures carry on

rtjohnso:
I'd be happy to extend this to also add a flag in trunk nodes to record when a node needs a split (due to a prior failure) so that a split will be forced next time we flush to it.

I _think_ the current splitting policy will already cause the node to split on the next flush, but it would be reasonable to add some mechanism to force that to happen no matter what the splitting policy is.

### DIFF
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -4627,7 +4627,7 @@ trunk_compact_bundle_cleanup_iterators(trunk_handle           *spl,
    for (uint64 i = 0; i < num_branches; i++) {
       trunk_btree_skiperator_deinit(spl, &skip_itor_arr[i]);
    }
-   debug_only(memset(skip_itor_arr, 0, num_branches * sizeof(*skip_itor_arr)));
+   debug_code(memset(skip_itor_arr, 0, num_branches * sizeof(*skip_itor_arr)));
 }
 
 /*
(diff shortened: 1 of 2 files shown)


src/trunk.c
```diff
@@ -4616,6 +4616,19 @@ trunk_btree_pack_req_init(trunk_handle   *spl,
                        spl->heap_id);
 }
 
+static void
+trunk_compact_bundle_cleanup_iterators(trunk_handle           *spl,
```
gapisback, line 4620: Minor point on the name of this function. It calls two other functions, below, one of which is named `destroy` and the other is named `deinit`. And this one is named `cleanup`.

Can we settle on naming this, say, just `trunk_compact_bundle_deinit_iterators()`, so that we are at least consistent with the name of one of the callees?

src/trunk.c
```diff
@@ -4616,6 +4616,19 @@ trunk_btree_pack_req_init(trunk_handle   *spl,
...
+                                       merge_iterator        **merge_itor,
+                                       uint64                  num_branches,
+                                       trunk_btree_skiperator *skip_itor_arr)
+{
+   platform_status rc = merge_iterator_destroy(spl->heap_id, merge_itor);
+   platform_assert_status_ok(rc);
+   for (uint64 i = 0; i < num_branches; i++) {
+      trunk_btree_skiperator_deinit(spl, &skip_itor_arr[i]);
+   }
+}
```
gapisback, line 4630: I like / prefer to see such `deinit` functions NULL'ing out their input handles, so that we don't have stale ptrs in the caller upon returning.

How would you feel about NULL'ing out input `skip_itor_arr` before exiting?

src/trunk.c
```diff
@@ -4838,9 +4851,12 @@ trunk_compact_bundle(void *arg, void *scratch_buf)
    }
 
    platform_status pack_status = btree_pack(&pack_req);
-   platform_assert(SUCCESS(pack_status),
-                   "platform_status of btree_pack: %d\n",
-                   pack_status.r);
+   if (!SUCCESS(pack_status)) {
+      platform_default_log("btree_pack failed: %d\n", pack_status.r);
```
gapisback, line 4855: For easier debugging, please consider changing this to print `"... failed: %d (%s)\n"` and print `platform_status_to_string()` for the `%s` arg.

gapisback:
Just for posterity, this PR addresses the issues reported in [this internal slack thread](https://vmware.slack.com/archives/G01MEDQ05MX/p1668704310333599?thread_ts=1668704254.666049&cid=G01MEDQ05MX).

gapisback:
@rtjohnso - I have given this a look and based on what we discussed this afternoon, as you'd like to merge this sooner, I'm passing this review. Have added very minor comments suggesting / requesting minor changes.

If you do add the mechanism of tracking this event with a bit and need to add handling for that on-disk bit, and if you need me to re-review that delta, do let me know. (Will get to it only next week, now.)

gapisback (approved):
@rtjohnso -- the changes are minor and am passing this so you can move this along.

If you add the new status-bit handling and need me to re-look, that will only happen next week.

Thanks.

src/trunk.c
rtjohnso: I agree with you that this is slightly annoying, but there is a logic to the naming deinit vs. destroy.  init/deinit functions operate on an object that is already allocated (e.g. on the stack).  create/destroy allocate/deallocate the object in addition to initializing/deinitializing it.  I think it is good to have different names for those different styles of functions.

Unfortunately, this cleanup function does both!  So I think renaming it to deinit or destroy would both be wrong.  So I'm leaving it as cleanup for now.

src/trunk.c
rtjohnso: Done (in debug mode)

src/trunk.c
rtjohnso: Nice!  Didn't know about that.

Issue #481, opened 2022-11-22
Memory allocated for large fingerprint array is not free()'ed in some code paths.


gapisback:
This issue was discovered accidentally while executing concurrent large-inserts workloads against Splinter. The actual test exercise was performed in a prototype dev-branch, which has support for some experimental features (shared memory support). 

A specialized mechanism was added to track large memory fragments and to recycle them after `free()` so that we do not run into OOM from shared memory. Using instrumentation and diagnostics to trace allocations / free / re-allocation of these large fragments, we discovered that after running a large concurrent inserts workloads, when SplinterDB is shutdown (gracefully), a whole set of large fragments are still marked as "in-use". Normally, these large fragments should be freed after tree compaction is completed.

This issue reports the suspected memory leak due to a missing-`free()` by way of showing the diagnostics that lead to this discovery:

--------------
1. **List of large fragments still appearing as "in-use" at the end of the workload**:

```
[OS-pid=461420, ThreadID=0, src/platform_linux/laio.c:318::io_handle_deinit()], Track freed fragment of size=1409024 bytes, at slot=0, addr=0x7f08a677c380, allocated_to_pid=461420, allocated_to_tid=0, shm_num_frags_inuse=69, shm_large_frag_hip=0x7f092c283780
  Fragment at slot=4, addr=0x7f08c062c7c0, size=25165824 is in-use, allocated_to_pid=461533, allocated_to_tid=3
  Fragment at slot=6, addr=0x7f08c2274200, size=25165824 is in-use, allocated_to_pid=461533, allocated_to_tid=3
  Fragment at slot=7, addr=0x7f08c3a742c0, size=25165824 is in-use, allocated_to_pid=461534, allocated_to_tid=1
[...]
  Fragment at slot=73, addr=0x7f092687c3c0, size=25165824 is in-use, allocated_to_pid=461580, allocated_to_tid=3
  Fragment at slot=75, addr=0x7f0929283780, size=25165824 is in-use, allocated_to_pid=461581, allocated_to_tid=2
Deallocated SplinterDB shared memory segment at 0x7f08a6772000, shmid=262154. Used=2268167616 bytes (~2.11 GiB, ~13 %), Free=14911695296 bytes (~13.88 GiB, ~86 %).
SplinterDB shared memory stats: nfrees=3870, nf_search_skipped=2142 (55 %), Large fragments in-use HWM=75 (found in-use=69), consumed=2254558784 bytes (~2.9 GiB).
[...]

PR #483, opened 2022-11-22
(#481) Fix missing free for fingerprint array in trunk_compact_bundle()

gapisback:
This commit fixes a missing free for memory allocated to the fingerprint array. The problem was seen to occur in highly concurrent inserts stress workload developed as part of concurrent POC-development and is difficult to reproduce in standalone Splinter.

In trunk_compact_bundle(), for a specific code-flow when we found no newly split siblings to replace with a new branch, we release the memory allocated for the trunk_compact_bundle_req{} struct, and exit processing. But the memory allocated (elsewhere, previously) for the fingerprint array hanging off this req struct remains unfreed.

This fix was identified through trace diagnostics and code walk through. And has been independently verified to fix this memory leak in the stress workloads that were used to generate this error.

### DIFF
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -4988,6 +4988,7 @@ trunk_compact_bundle(void *arg, void *scratch_buf)
          spl->stats[tid].compaction_time_wasted_ns[height] +=
             platform_timestamp_elapsed(compaction_start);
       }
+      platform_free(spl->heap_id, req->fp_arr);
       platform_free(spl->heap_id, req);
    } else {
       if (spl->cfg.use_stats) {


gapisback:
@rtjohnso - Am assigning this PR to you, as discussed earlier. You helped me triage this fix last week.

I have run few variations of `build/release-msan/bin/driver_test splinter_test` for `--perf` option but it is not triggering this code path. I suspect we need some specific data load to trigger the condition `if (num_replacements == 0)` under which the missing `free()` occurs.

I have verified that this fix in my dev-branch where the large concurrent inserts work load is running, so am comfortable with this change (even w/o its own stand-alone repro test case).

rtjohnso (approved):
Ship it.

ajhconway (approved):
Good catch

Issue #288, opened 2022-02-24
Unit tests log messages should be more easily discovered and viewed; e.g. all output to stdout


gapisback:
This issue requests for some enhancements in the way outputs from unit tests are generated. Currently, they all go to files in `tmp`; i.e. `/tmp/unit_test.stderr` and `/tmp/unit_test.stdout`. Feedback received on this is as follows:

---
Request the ability to have the unit test output get printed to the terminal, somehow….

In doing this refactoring work, I’m trying to test-drive it, so I’m running individual unit tests like bin/unit/splinterdb_test repeatedly.  And they often fail with informative messages, e.g. an assertion trips.  

But the assertion doesn’t appear in the terminal.  Instead, I repeatedly have to run this 3-part command
```
rm /tmp/unit_test.std* && bin/unit/splinterdb_test ; cat /tmp/unit_test.std*
```

If I don’t do the rm then the output is contaminated with prior test output.

On past projects, I got accustomed to having both:

1.  test stdout and stderr are always printed when a test fails.
2. one can invoke the unit test binary with -v to always print the full output, regardless of exit status. 

It would be really nice if at least one of those two things worked on this project.

--- **Another use-case**: when running the unit tests in the debugger.  

It’s nice to see assertion text or other messages printed to the terminal, rather that having to go cat this other random file to see what’s happened since the last step/breakpoint/whatever.

---
**Proposed solution** to investigate and productize:

- Figure out if we can better manage the swizzling of platform output file handles that CTests do
- Provide an optional `-v` | `--verbose` flag under which these file handles are redirected to stdout. This way all output will come to the terminal.
- Investigate if this solution would address the debugging use-case (reported internally)

PR #487, opened 2022-11-29
Enable application to silence info logs

### LINKED ISSUE #441: Informational messages should be hidden by default
## Background
Currently, every time a program uses SplinterDB, some messages are printed to stdout and stderr, e.g.
```
Fingerprint size 29 too large, max value size is 5, setting to 27
fingerprint_size: 27
```
When the allocator is mounted from an existing device or unmounted on exit, additional messages are printed, e.g.
```
Allocated 4 extents at mount: (0 MiB)
Allocated at unmount: 1 MiB
```

## User story
As an application developer, I expect to be in full control of my program's stdout and stderr streams.  I'd like informational messages from a library to be silent by default, and I'd like a way to control where to send both informational messages and error messages from a library.

## Suggested changes
1. The two `platform_error_log` calls here
   - https://github.com/vmware/splinterdb/blob/cdc18e73a75be2ce1e27edeba6b830e75cb46e80/src/trunk.c#L9069-L9070
   - https://github.com/vmware/splinterdb/blob/cdc18e73a75be2ce1e27edeba6b830e75cb46e80/src/trunk.c#L9108-L9109
  should instead be `platform_default_log` calls

2. On library initialization:
   - `Platform_default_log_handle` should default to `/dev/null`
   - `Platform_error_log_handle` can continue to default to stderr.
 
3. The public API should grow a new function with a signature like:
    ```c
    void splinterdb_set_log_streams(FILE* info_stream, FILE* error_stream)
    ```
    so that an application can choose to override those defaults.

rosenhouse:
Fixes #441 , mostly.

Info messages now default to `/dev/null`.  Error messages still go to stderr.

The app may customize that by calling a new function `splinterdb_logs_set_streams()`.

A future PR could provide command-line or env-var control over _test_ output.

### DIFF
--- a/tests/functional/test_dispatcher.c
+++ b/tests/functional/test_dispatcher.c
@@ -26,6 +26,8 @@ usage(void)
 int
 test_dispatcher(int argc, char *argv[])
 {
+   platform_set_log_streams(stdout, stderr);
+
    platform_default_log("%s: %s\n", argv[0], BUILD_VERSION);
    // check first arg and call the appropriate test
    if (argc > 1) {
--- a/include/splinterdb/platform_linux/public_platform.h
+++ b/include/splinterdb/platform_linux/public_platform.h
@@ -67,9 +67,22 @@ typedef int32 bool;
 #endif
 typedef uint8 bool8;
 
-// See platform.c
+// By default, platform_default_log() messages are sent to /dev/null
+// and platform_error_log() messages go to stderr.
+//
+// Use platform_set_log_streams() to send the log messages elsewhere.
 typedef FILE                platform_log_handle;
-extern platform_log_handle *Platform_default_log_handle; // stdout FILE handle
-extern platform_log_handle *Platform_error_log_handle;   // stderr FILE handle
+extern platform_log_handle *Platform_default_log_handle;
+extern platform_log_handle *Platform_error_log_handle;
+
+// Set the streams where informational and error messages will be printed.
+//
+// These default to /dev/null and stderr, respectively.
+//
+// For example, to send info messages to stdout and errors to stderr, run:
+//    platform_set_log_streams(stdout, stderr);
+void
+platform_set_log_streams(platform_log_handle *info_stream,
+                         platform_log_handle *error_stream);
 
 #endif // __PUBLIC_PLATFORM_H
--- a/src/merge.c
+++ b/src/merge.c
@@ -408,12 +408,12 @@ merge_iterator_create(platform_heap_id hid,
    if (!out_itor || !itor_arr || !cfg || num_trees < 0
        || num_trees >= ARRAY_SIZE(merge_itor->ordered_iterator_stored))
    {
-      platform_default_log("merge_iterator_create: bad parameter merge_itor %p"
-                           " num_trees %d itor_arr %p cfg %p\n",
-                           out_itor,
-                           num_trees,
-                           itor_arr,
-                           cfg);
+      platform_error_log("merge_iterator_create: bad parameter merge_itor %p"
+                         " num_trees %d itor_arr %p cfg %p\n",
+                         out_itor,
+                         num_trees,
+                         itor_arr,
+                         cfg);
       return STATUS_BAD_PARAM;
    }
 
--- a/src/platform_linux/platform.c
+++ b/src/platform_linux/platform.c
@@ -12,17 +12,32 @@ __thread threadid xxxtid;
 bool platform_use_hugetlb = FALSE;
 bool platform_use_mlock   = FALSE;
 
-// By default, platform_default_log() messages go to stdout, and
-// platform_error_log() messages go to stderr. These can be changed to module-
-// or test-specific log files, by overriding these settings.
+// By default, platform_default_log() messages are sent to /dev/null
+// and platform_error_log() messages go to stderr (see below).
+//
+// Use platform_set_log_streams() to send the log messages elsewhere.
 platform_log_handle *Platform_default_log_handle = NULL;
 platform_log_handle *Platform_error_log_handle   = NULL;
 
 // This function is run automatically at library-load time
 void __attribute__((constructor)) platform_init_log_file_handles(void)
 {
-   Platform_default_log_handle = stdout;
+   FILE *dev_null_file = fopen("/dev/null", "w");
+   platform_assert(dev_null_file != NULL);
+
+   Platform_default_log_handle = dev_null_file;
    Platform_error_log_handle   = stderr;
+}
+
+// Set the streams where informational and error messages will be printed.
+void
+platform_set_log_streams(platform_log_handle *info_stream,
+                         platform_log_handle *error_stream)
+{
+   platform_assert(info_stream != NULL);
+   platform_assert(error_stream != NULL);
+   Platform_default_log_handle = info_stream;
+   Platform_error_log_handle   = error_stream;
 }
 
 platform_status
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -9032,7 +9032,7 @@ trunk_config_init(trunk_config        *trunk_cfg,
    size_t max_value_size        = 64 - __builtin_clzll(max_value);
 
    if (filter_fingerprint_size > 32 - max_value_size) {
-      platform_error_log(
+      platform_default_log(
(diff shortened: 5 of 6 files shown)


include/splinterdb/splinterdb.h
```diff
@@ -321,4 +321,18 @@ splinterdb_stats_print_lookup(const splinterdb *kvs);
...
+// These default to stdout and stderr, respectively.
+//
+// A program may use this function to override the defaults, to send
+// the messages to a file, or /dev/null or elsewhere.
+//
+// Typical usage would be to silence info messages, but leave error
+// messages going to stderr:
+//   splinterdb_log_streams(fopen("/dev/null", "w"), stderr);
+void
+splinterdb_log_streams(platform_log_handle *info_stream,
```
gapisback, line 335: (As you remarked elsewhere) I prefer the name to be `splinterdb_set_log_streams()`. The term `_set_` makes it clear as to what's going on. (Also will take away the potential ambiguity of `_log_streams` implying we are `logging'` something. `log` is also a verb.)

src/splinterdb.c
```diff
@@ -686,3 +686,11 @@ splinterdb_stats_reset(splinterdb *kvs)
 {
    trunk_reset_stats(kvs->spl);
 }
+
+void
```
gapisback, line 690: A minor single-line comment will help w/ readability. (I understand that it's a minimalistic function.)

include/splinterdb/splinterdb.h
gapisback: Add-on comment: Btw, I see this comment elsewhere in `platform.c`:

```
platform_linux/platform.c:16:// platform_error_log() messages go to stderr. These can be changed to module-
platform_linux/platform.c-17-// or test-specific log files, by overriding these settings.
platform_linux/platform.c-18-platform_log_handle *Platform_default_log_handle = NULL;
```

It will be useful to sneak-in a reference to your new method by updating L18 to something like:

```
or test-specific log files, by using the splinterdb_set_log_streams() interface.
```

gapisback:
@rosenhouse - I found few other usages of `platform_error_log` which could also be changed to `platform_default_log`, while are you doing this kind of clean-up. Kindly consider this list.

These ones seem like they **should** be changed:

```
../tests/functional/test_functionality.c:51:   platform_error_log("Saw a total of %lu keys\n", count);
```

These ones below, which report usage messages ... bit less unclear if they should be changed.

```
../tests/functional/splinter_test.h:138:      platform_error_log("Test Configuration:\n");
../tests/functional/splinter_test.h:139:      platform_error_log("\t--tree-size-gib\n");
```

```
../tests/config.c:76:   platform_error_log("Configuration: (default)\n");
../tests/config.c:77:   platform_error_log("\t--page-size (%d)\n", TEST_CONFIG_DEFAULT_PAGE_SIZE);
../tests/config.c:78:   platform_error_log("\t--extent-size (%d)\n",
[...]
```

```
../tests/functional/splinter_test.c:2369:   platform_error_log(
../tests/functional/splinter_test.c-2370-      "Usage:\n"

../tests/functional/test_dispatcher.c-12-usage(void)
../tests/functional/test_dispatcher.c-13-{
../tests/functional/test_dispatcher.c:14:   platform_error_log("List of tests:\n");
../tests/functional/test_dispatcher.c:15:   platform_error_log("\tbtree_test\n");
../tests/functional/test_dispatcher.c:16:   platform_error_log("\tfilter_test\n");

```

Why can't these usage reporting messages be `platform_default_log`? (One possible issue is that if the appln-developer uses your new `_set_` interface to reset default log-stream to /dev/null then they will not see these messages. Not sure how that can be handled.)



An [unrelated] side-cleanup would be several instances like this one:

```
trunk.c:7575:         platform_error_log("trunk_verify: empty bundle\n");
trunk.c:7576:         platform_error_log("addr: %lu\n", addr);

trunk.c:7591:            platform_error_log("trunk_verify: start_branch mismatch\n");
trunk.c:7592:            platform_error_log("addr: %lu\n", addr);
```

Why not merge both calls into one, so we get one line of output? There are several such lines to merge ... so it's ok if you choose to not do it now, and I'd be happy to clean this up off-line in a separate issue.

gapisback:
@rosenhouse - About your question regarding what the default log stream should be: I prefer it to be `stdout`, the way you have left it unchanged with this PR.

I have found many messages sent to stdout to be helpful to troubleshoot the system and its workings, especially in more 'complex' integrations like the one we are embarking on in POCs.

gapisback (approved):
I have left minor comments which can be easily incorporated (I think). Passing this review so as to unblock you. If you need me to re-review, let me know, and I'll do so.

include/splinterdb/splinterdb.h
rosenhouse: On the name: I've changed it to `splinterdb_logs_set_streams`.  This includes the `_set_` that you asked for, but retains using `_logs_` as a sort of sub-namespacing.  That is intended to follow the pattern that Alex and I discussed recently here: https://github.com/vmware/splinterdb/pull/459#discussion_r1000921275.  That way, any future functions related to message logs can have the same prefix.

rosenhouse: Good catch on the code comment in `platform.c`, I've updated it to mention the new function.

src/splinterdb.c
rosenhouse: Thanks, done in latest.

rosenhouse:
@ajhconway @rtjohnso @gapisback I've updated this PR based on our meeting yesterday:
-  Info messages now default to `/dev/null`
- `driver_test` overrides that default, sending them (back) to `stdout` to maintain existing behavior.  However, **it doesn't use the new function to do this, see comments in `test_dispatcher.c`, but I'd like feedback on this part.**
-  The unit tests were already overriding the streams, sending them to separate files.  That behavior is left as-is, and can be addressed separately (see #288)

Please take another look, thanks!

gapisback:
@rosenhouse - Took a re-look. I'm fine with this round of changes you've done.

About `test_dispatcher.c` manually setting the `Platform_default_log_handle = stdout;`, I'm not too hung up on this.

If you _could_ use the new `logs_set_` interface, that might become another example usage for someone to discover while browsing the code. There **is** some value in that discovery.

Were there any build issues that prevented you from using the new _set_ interface? If not, and if you still have the energy, might be good to change the code to use the interface (so we, in a way, discourage leaking globals around).

rosenhouse:
Yes, you're right, I shouldn't be so worried about layering violations in the tests, given that we already dislike how the tests are organized.  Fixed on latest.

tests/functional/test_dispatcher.c
```diff
@@ -26,6 +26,9 @@ usage(void)
 int
 test_dispatcher(int argc, char *argv[])
 {
+   // Info messages go to stdout
```
gapisback, line 29: (Sorry, very minor nit.)

You could remove this comment entirely as it's implied by the interface and the call itself.

Or, expand it to state both parts: `// Info messages go to stdout, and error messages to stderr`

The way it is right now it's high-lighting the former, making the reader wonder what's so special about info-msgs.

rosenhouse: Sure thing, fixed in latest.

include/splinterdb/splinterdb.h
```diff
@@ -321,4 +321,14 @@ splinterdb_stats_print_lookup(const splinterdb *kvs);
...
 splinterdb_stats_reset(splinterdb *kvs);
 
+// Set the streams where informational and error messages will be printed.
+//
+// These default to /dev/null and stderr, respectively.
+//
+// To capture info messages on stdout, run:
+//    splinterdb_logs_set_streams(stdout, stderr);
+void
+splinterdb_logs_set_streams(platform_log_handle *info_stream,
```
rtjohnso, line 331: Not a deal breaker, but maybe this goes in `public_platform.h`?

rtjohnso (approved):
LGTM, but consider moving `splinterdb_logs_set_streams` into `public_platform.h`.  Whatever you decide on that is fine.

include/splinterdb/splinterdb.h
rosenhouse: Fair.  Should it also get renamed to `platform_set_log_streams()`?

PR #489, opened 2022-11-29
Fix Default Filter Config and Remove Superfluous Printing

ajhconway:
Changes the default filter config so that the system doesn't print a warning.

Removes allocator, trunk, and task test printing.

### DIFF
--- a/src/rc_allocator.c
+++ b/src/rc_allocator.c
@@ -473,10 +473,6 @@ rc_allocator_mount(rc_allocator        *al,
          al->stats.curr_allocated++;
       }
    }
-   platform_default_log(
-      "Allocated %lu extents at mount: (%lu MiB)\n",
-      al->stats.curr_allocated,
-      B_TO_MiB(al->stats.curr_allocated * cfg->io_cfg->extent_size));
    return STATUS_OK;
 }
 
@@ -485,10 +481,6 @@ void
 rc_allocator_unmount(rc_allocator *al)
 {
    platform_status status;
-
-   platform_default_log(
-      "Allocated at unmount: %lu MiB\n",
-      B_TO_MiB(al->stats.curr_allocated * al->cfg->io_cfg->extent_size));
 
    // persist the ref counts upon unmount.
    uint32 io_size =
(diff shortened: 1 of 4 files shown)


src/rc_allocator.c
```diff
@@ -473,10 +473,6 @@ rc_allocator_mount(rc_allocator        *al,
          al->stats.curr_allocated++;
       }
    }
-   platform_default_log(
```
gapisback, line 476: I could go either way about deleting these messages. But, ... With reworked fix for PR #487 where we will default to `/dev/ null/` for `stdout`, it should not really matter that these info messages coming out, right?

What's your overall direction in such cases -- Should we all just not add any info-messages at all?

(I do find some info-messages quite useful and am personally more tolerant of messages.)

src/splinterdb.c
```diff
@@ -88,7 +88,7 @@ splinterdb_config_set_defaults(splinterdb_config *cfg)
       cfg->filter_index_size = 256;
    }
    if (!cfg->filter_remainder_size) {
-      cfg->filter_remainder_size = 6;
+      cfg->filter_remainder_size = 4;
```
gapisback, line 91: I haven't checked the logic here, but if this change is what will avoid printing the fingerprint array related message, then this change is fine.

gapisback (approved):
Quick review of minor messages cleanup. One minor comment / question. Looks good overall, nothing of any major concern.

rosenhouse (approved):
Let's pull this in first, since #487 will take me a bit longer to finish.

src/rc_allocator.c
ajhconway: I don't really know the answer, but these ones in particular aren't super useful. My question back is what sorts of notifications are actually useful?

There are places where we assert no space leaks, and that feels like the right way to monitor this sort of thing.

src/splinterdb.c
ajhconway: Yes, this was happening automatically in the config auditor, and now it's just the default.

PR #490, opened 2022-11-29
tweak example running policy

rtjohnso:
Adds a new target, run-example, which runs as part of CI but not part of any version of "run-tests".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ PLATFORM_DIR = platform_$(PLATFORM)
 
 help::
 	@echo 'Usage: make [<target>]'
-	@echo 'Supported targets: clean all libs all-tests run-tests test-results install'
+	@echo 'Supported targets: clean all libs all-tests run-tests test-results run-examples install'
 
 #*************************************************************#
 # SOURCE DIRECTORIES AND FILES
--- a/Makefile
+++ b/Makefile
@@ -484,6 +484,9 @@ run-tests: all-tests
 test-results: all-tests
 	INCLUDE_SLOW_TESTS=true BINDIR=$(BINDIR) ./test.sh 2>&1 | tee ./test-results
 
+run-examples: all-examples
+		for i in $(EXAMPLES_BINS); do $$i || exit; done
+
 INSTALL_PATH ?= /usr/local
 
 install: $(LIBDIR)/libsplinterdb.so $(LIBDIR)/libsplinterdb.a
--- a/ci/steps.lib.yml
+++ b/ci/steps.lib.yml
@@ -130,7 +130,7 @@ config:
     dir: #@ input_name
     args:
     - "-c"
-    - "make $MAKE_HELP all run-tests"
+    - "make $MAKE_HELP all run-tests run-examples"
 #@ end
 
 ---
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ## What is SplinterDB?
 SplinterDB is a key-value store designed for high performance on fast storage devices.
 
-See
+See [this talk at the CMU Database Seminar](https://www.youtube.com/watch?v=1gOlXfbiT_Y) and this paper:
 > Alexander Conway, Abhishek Gupta, Vijay Chidambaram, Martin Farach-Colton, Richard P. Spillane, Amy Tai, Rob Johnson:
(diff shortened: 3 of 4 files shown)


rosenhouse (approved):
I like it.

Makefile
```diff
@@ -484,6 +484,9 @@ run-tests: all-tests
 test-results: all-tests
 	INCLUDE_SLOW_TESTS=true BINDIR=$(BINDIR) ./test.sh 2>&1 | tee ./test-results
 
+run-examples: all-examples
```
gapisback, line 487: Suggestion: If you are introducing this new target, please also include it in `help` output:

```
sdb-fdb-build:[39] $ make help
Usage: make [<target>]
Supported targets: clean all libs all-tests run-tests test-results install
```

Will be good to see `... run-tests run-examples ...` in above output.

Makefile
```diff
@@ -484,6 +484,9 @@ run-tests: all-tests
...
+		for i in $(EXAMPLES_BINS); do $$i || exit; done
```
gapisback, line 488: Any reason why you preferred synthesizing the list of example programs like so, above, rather than inlining the order in the now-deleted function, `test_example_programs()`?

One thing I noticed is that this `for i in ...` will generate the list in some order which may not match the order in which these example programs were sequenced. Upon testing in your branch, I found the order to be:

```
for i in build/release/bin/examples/splinterdb_wide_values_example build/release/bin/examples/splinterdb_intro_example build/release/bin/examples/splinterdb_iterators_example build/release/bin/examples/splinterdb_custom_ipv4_addr_sortcmp_example; do $i || exit; done
```

The `intro` example comes after the `splinterdb_wide_values_example`, which to a newbie user trying to learn the system is non-intuitive.

My thought is this: There are two candidate consumers of these examples: The end-user and the testing-system.

How about we make a small amendment to `Makefile` to support two targets: `test-examples` and `run-examples`?

The former is what you have in the PR right now. This will run all example programs found in the bin dir. And it is sufficient to just call the new modified `test-examples` target as part of nightly CI target (as you have it in the `ci/steps.lib.yml` file).

The latter is end-user oriented. Yes; It requires a manual update to add a new example program to the list. 

But this, then, will also give us a way to **not** have every end-user run some example programs that have an impact. (For instance, I heard both Alex & you remark that an example program that always does a `ping` to an inet-address should not be part of the test `run-examples` target.)

End-user can still run some example program that is omitted from the `run-examples` list to see how they perform; so no functionality is lost.

I tried out a small change to your `Makefile` and the following seems to do the job:

```
@@ -484,9 +484,17 @@ run-tests: all-tests
 test-results: all-tests
        INCLUDE_SLOW_TESTS=true BINDIR=$(BINDIR) ./test.sh 2>&1 | tee ./test-results

-run-examples: all-examples
+test-examples: all-examples
                for i in $(EXAMPLES_BINS); do $$i || exit; done

+# Target for end-user to run example programs in some useful order.
+run-examples: all-examples
+               for i in splinterdb_intro_example \
+               splinterdb_wide_values_example \
+               splinterdb_iterators_example \
+               splinterdb_custom_ipv4_addr_sortcmp_example; \
+         do $(BINDIR)/$(EXAMPLES_DIR)/$$i || exit; done
+
```

ci/steps.lib.yml
```diff
@@ -130,7 +130,7 @@ config:
     dir: #@ input_name
     args:
     - "-c"
-    - "make $MAKE_HELP all run-tests"
+    - "make $MAKE_HELP all run-tests run-examples"
```
gapisback, line 133: If you accept my proposal to add a new target to `Makefile` to run all examples, then this would change to `test-examples`.

(That name makes sense to me given that this invocation is in CI scripts which are there for testing the system.)

gapisback (approved):
In principle the changes are fine.

I have a proposal for an improvement over this solution, which will make the usability better w/o compromising testability.

Will leave it up to the author to consider and amend the change set accordingly.

Makefile
rtjohnso: If there is a specific order you want users to view these examples, how about naming them to make the order clear, e.g.

`splinterdb_exampe_01_intro`
`splinterdb_example_02_wide_values`
etc...

As it is, it is clear only that intro probably comes before the others, but the order in which a beginner should read the others is not at all clear. 

The above naming change would provide guidance both to users that find the examples by browsing the source tree and by users who find them by running `make run-examples`.

PR #493, opened 2022-12-01
When file open fails, return error code instead of crashing

### LINKED ISSUE #482: Friendlier error message when database file can't be created....
When you try to create a database in a directory that doesn't exist, you get a pretty inscrutable assertion: 
```
OS-pid=7, OS-tid=7, Thread-ID=0, Assertion failed at src/platform_linux/laio.c:100:io_handle_init(): "io->fd != -1".
```

Given that could be a common user-error, we should make it more helpful.  Ideally we could return an errno that an application could interpret.  Also maybe log something helpful.

rosenhouse:
Fixes #482

### DIFF
--- a/src/platform_linux/laio.c
+++ b/src/platform_linux/laio.c
@@ -97,7 +97,11 @@ io_handle_init(laio_handle         *io,
    } else {
       io->fd = open(cfg->filename, cfg->flags);
    }
-   platform_assert(io->fd != -1);
+   if (io->fd == -1) {
+      platform_error_log(
+         "open() '%s' failed: %s\n", cfg->filename, strerror(errno));
+      return CONST_STATUS(errno);
+   }
 
    req_size =
       sizeof(io_async_req) + cfg->async_max_pages * sizeof(struct iovec);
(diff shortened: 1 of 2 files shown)


rtjohnso (approved):
LGTM.

src/platform_linux/laio.c
```diff
@@ -97,7 +97,11 @@ io_handle_init(laio_handle         *io,
    } else {
       io->fd = open(cfg->filename, cfg->flags);
    }
-   platform_assert(io->fd != -1);
+   if (io->fd == -1) {
+      platform_error_log(
+         "open %s failed: %s\n", cfg->filename, strerror(errno));
```
gapisback, line 102: Minor nit suggestion. Can you rephrase the message as: ```open() '%s' failed ...```

The `()` will make it clear that a system call to some fn named `open` failed. The enclosing `''` is conventional, in my experience, to enclose string parameters so if (ever) the parameter was an empty string `""`, you'd see it immediately. (Minor stylistic suggestion.)

gapisback (approved):
Cleanup fix to handle bad file `open()` returns is fine. Thanks for doing this!

src/platform_linux/laio.c
rosenhouse: Fixed on latest, thanks.

PR #494, opened 2022-12-02
Simplify output handling in unit tests

### LINKED ISSUE #288: Unit tests log messages should be more easily discovered and viewed; e.g. all output to stdout

rosenhouse:
- Unit tests respect a new `VERBOSE` environment variable, which ensures output goes to `stdout`/`stderr`.

- When `VERBOSE` is unset, or set to `0` or `false`, then:
    - `stdout` will show only 1 status line per test
    - `stderr` will be silent, except for error messages

- When `VERBOSE` is set to any other value, then library and unit test info messages go to `stdout`.

- When a unit test fails, the user will see a suggestion to set `VERBOSE=1` to see all the detailed output.
  (TODO: the suggestion doesn't appear on a crash, but that'd be nice)

- Library log lines from unit tests no longer go to `/tmp/unit_test.std[out,err]`

- `Platform_default_log_handle` and `Platform_error_log_handle` are no longer part of the public API surface.  Applications (and tests that mimic apps) must use `platform_set_log_streams()` instead (see #487)

- Cleans up some tests to avoid a direct dependency on `platform.h`:
  - Removes some references to `Platform_*_log_handle` globals and instead calls new `platform_set_log_streams()`
  - Removes some calls to `platform_default_log()` and instead use a dedicated `CTEST_LOG_INFO()` macro.
  - This isn't complete, but is a start

Fixes #288

-----

Here's a side-by-side comparison, running in quiet mode (default) on the left, and with `VERBOSE=1` on the right.
image

and here's what it looks like when a test fails and you haven't set VERBOSE yet:
image


ci/steps.lib.yml
```diff
@@ -105,6 +105,7 @@ config:
     INCLUDE_SLOW_TESTS: #@ str(not quick).lower()
     RUN_NIGHTLY_TESTS: #@ str(test_nightly).lower()
     BUILD_VERBOSE: "1"
+    VERBOSE: "1"
```
rtjohnso, line 108: Any reason to make this an environment variable instead of a command-line option?

include/splinterdb/platform_linux/public_platform.h
```diff
@@ -67,17 +67,12 @@ typedef int32 bool;
 #endif
 typedef uint8 bool8;
 
-// By default, platform_default_log() messages are sent to /dev/null
-// and platform_error_log() messages go to stderr.
-//
-// Use platform_set_log_streams() to send the log messages elsewhere.
-typedef FILE                platform_log_handle;
```
rtjohnso, line 74: Excellent idea to remove these from the public api.

rtjohnso (approved):
I generally prefer command-line args over environment variables, but env vars are fine if there's a technical challenge to implementing this as a command-line arg.

ci/steps.lib.yml
rosenhouse: Yeah, a flag would be preferable for ease of typing and for discoverability.  I just couldn't figure out how to make it work: the interactions between `unit/main.c`, `config.h`, `unit/config_parse_test.c`, and `unit/splinter_test.c` were just too messy for me to untangle in the couple hours that I spent on this.  I'd welcome some help on that from @gapisback or anyone else.

tests/unit/limitations_test.c
```diff
@@ -66,8 +66,19 @@ CTEST_DATA(limitations)
...
-   Platform_default_log_handle = fopen("/tmp/unit_test.stdout", "a+");
-   Platform_error_log_handle = fopen("/tmp/unit_test.stderr", "a+");
+   // This test exercises error cases, so even when everthing succeeds
+   // it generates lots of "error" messages.
+   // By default, that would go to stderr, which would pollute test output.
+   // Here we ensure those _expected error messages_ are only printed
+   // when the caller opts in, by setting the VERBOSE env var.
+   if (Ctest_verbosity) {
+      platform_set_log_streams(stdout, stderr);
+      CTEST_LOG_INFO("\nVerbose mode on, successful tests may print messages that appear to be errors\n");
```
ajhconway, line 76: I would add "..., exercising error cases, so..." into the output message to give a little context

tests/unit/misc_test.c
```diff
@@ -56,9 +56,26 @@ test_vfprintf_usermsg(platform_log_handle *log_handle,
...
-CTEST_SETUP(misc) {}
+CTEST_SETUP(misc)
+{
+   // Tests in this file exercise log-printing using real-looking error
+   // messages. It can be surprising to see error messages printed from a
+   // successful test run. So lets send those messages to /dev/null unless
+   // VERBOSE=1.
+   if (Ctest_verbosity) {
+      data->log_output = stdout;
+      CTEST_LOG_INFO("\nVerbose mode on, successful tests may print messages "
```
ajhconway, line 72: same here

ajhconway (approved):
LGTM

tests/unit/limitations_test.c
rosenhouse: Good point, fixed in latest.

tests/unit/misc_test.c
rosenhouse: thanks, fixed in latest.

tests/unit/btree_stress_test.c
```diff
@@ -234,23 +234,15 @@ CTEST2(btree_stress, test_random_inserts_concurrent)
    if (!iterator_tests(
           (cache *)&data->cc, &data->dbtree_cfg, root_addr, nkvs, data->hid))
    {
-      platform_default_log("invalid ranges in original tree\n");
+      CTEST_ERR("invalid ranges in original tree\n");
    }
 
-   /* platform_default_log("\n\n\n"); */
```
gapisback, line 240: I'm not so sure about deleting this dead code and the old lines on L249 onwards below.

Seems like the code author had felt a need that this kind of debugging would be handy if ever this one had to troubleshoot failures in this test's execution.

I'm generally tolerant about such code being left behind, as it does help some future engineer who might have to debug this stuff. (Rob didn't complain, surprisingly ... as he's the author.)

What if we (a) either retain this dead-code in comments or (b) enclose it under if (FALSE) { }? The code is there for troubleshooting and doesn't get in anyone's way.

tests/unit/main.c
```diff
@@ -35,6 +35,9 @@ static const char *testcase_name = NULL;
 
 typedef int (*ctest_filter_func)(struct ctest *);
 
+// If 0, then CTEST_LOG_INFO() and CTEST_LOG() will no-op.  Defaults to 0.
+int Ctest_verbosity = 0;
```
gapisback, line 39: Any reason  why this is declared as `int` rather than as `bool` ? (I think I dimly recall some issue with using `bool` as a type in ctests' `main.c` ... Maybe that's why this is an `int` ?)

You may want to doc that this is really being treated as a boolean variable.

It would be good to be consistent elsewhere where this variable is checked. In some places code does `if (Ctest_verbosity)` and in others it does `if (Ctest_verbosity > 0)`.

tests/unit/splinterdb_stress_test.c
```diff
@@ -132,11 +129,11 @@ CTEST2(splinterdb_stress, test_naive_range_delete)
       ASSERT_EQUAL(0, rc);
    }
 
-   platform_default_log("loaded %u k/v pairs\n", num_inserts);
+   CTEST_LOG_INFO("loaded %u k/v pairs\n", num_inserts);
 
    uint32 num_rounds = 5;
    for (uint32 round = 0; round < num_rounds; round++) {
-      platform_default_log("range delete round %d...\n", round);
+      CTEST_LOG_INFO("range delete round %d...\n", round);
```
gapisback, line 136: My comment is not specifically tied to this diff, or the diff below on L164.

I just wondered ... how would a new test author know which interface to use for printing messages.

Unit-tests: I noticed that in `splinter_test.c`, we are still using `platform_default_log()`, while in other files we are using `CTEST_LOG_INFO()`.

Functional-tests: are all using `platform_default_log()`.

I am wondering ... how can we provide guidelines on what to use where. May be a `README` in `tests/` sub-dir with some general guidelines on what to use where (`platform_default_log()`, `CTEST_LOG_INFO()` and `CTEST_ERR()`) will be helpful. I will certainly forget these details and will benefit from archived how-to instructions for future reference.

tests/unit/splinterdb_stress_test.c
```diff
@@ -164,7 +161,7 @@ exec_worker_thread(void *w)
 
    pthread_t thread_id = pthread_self();
 
-   fprintf(stderr, "Writing lots of data from thread %lu\n", thread_id);
+   CTEST_LOG_INFO("Writing lots of data from thread %lu\n", thread_id);
```
gapisback, line 164: This and the diff on L179 is a good catch. Thanks!

gapisback (approved):
Hi, @rosenhouse - Thanks for taking an issue off my plate and improving the usage landscape somewhat!

I have marked-up minor comments and suggestions; nothing major, so assume that you will be able to address them accordingly.

A general comment is that it will be a bit confusing for a new test-code author to figure out which interface to use where, `functional` and `unit` tests code. (I myself will forget and will end up searching what we did in other tests and copy from there.) 

I suspect there will be a bit more churn in this area, especially relating to interface and handling of (current) env-var `VERBOSE` and / or `--verbose-progress` argument.

I think it will be helpful to having coding-level guidelines perhaps, in [testing.md](https://github.com/vmware/splinterdb/blob/main/docs/testing.md) or in [unit_testing.md](https://github.com/vmware/splinterdb/blob/main/docs/unit-testing.md). 

We can develop this off-line once the dust settles on this area.

tests/unit/ctest.h
```diff
@@ -79,6 +79,19 @@ struct ctest {
...
+
+// For immedate logging to stdout
+// (contrast with CTEST_LOG which does buffered/delayed logging to stderr)
+#define CTEST_LOG_INFO(...)                                                    \
+   do {                                                                        \
+      if (Ctest_verbosity) {                                                   \
+         fprintf(stdout, __VA_ARGS__);                                         \
+         fflush(stdout);                                                       \
+      }                                                                        \
+   } while (0)
```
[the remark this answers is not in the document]
rosenhouse: That seems dangerous to me.

Suppose someone forgets the `;` and writes
```c
CTEST_LOG_INFO("foo")
something_important();
```

With the `do/while` implementation of the macro, the compiler will error: 
```
tests/unit/splinterdb_quick_test.c:862:25: fatal error: expected ';' after do/while statement
   CTEST_LOG_INFO("foo")
                        ^
                        ;
```

But with your proposed `if/else` implementation, it compiles cleanly, and then `something_important()` only gets called when `VERBOSE` is unset.

tests/unit/splinterdb_stress_test.c
rosenhouse: Yes, that was a problem before this PR (multiple ways to print between tests).  I think `splinter_test.c` is a special case, because it is testing private APIs.  That said, based on this feedback, I've made the following updates to this PR:
1. updated it to use the `CTEST_LOG_INFO()` macro as much as I could easily... though there are other places where it uses private APIs to print to the `Platform_default_log_stream` and I'm not going to touch that in this PR.  
2. Made it respect the verbose flag by calling `platform_set_log_streams()` the right way
3. added a note to the top of that test file now to warn folks not to follow patterns established there.

tests/unit/main.c
rosenhouse: Good point.  Changed to `_Bool` on latest.

rosenhouse:
Thanks for the review.  I've tried to improve the "new test author experience" part by updating `splinter_test` as much as was practical in a short-time, and leaving a note up top to caution against following patterns there.  And then there's at least a clearer distinction between "unit" versus "functional".

And yes, let's merge this as-is, and we'll defer the flag thing to later.

tests/unit/btree_stress_test.c
rosenhouse: I strongly dislike dead code.  I'm going to delete it.  No one is speaking up to claim this code as theirs, and I'm guessing it could be added back by someone who's making changes.

ci/steps.lib.yml
rosenhouse: Yes, we'll return to this in future work

PR #496, opened 2022-12-04
Copy-on-write Trunk

ajhconway:
This changeset implements copy-on-write for trunk nodes, which includes several high-level changes. This PR still needs to be rebased onto main, but the the purpose is to discuss high- and low-level design decisions.

Changes in this PR:

1. Trunk root lock. A distributed RW lock is used to access/change the current root.

2. Flush from root. Flushes proceed from the root and cascade immediately rather than being triggered at the beginning of `trunk_compact_bundle`.

3. Copy-on-write. Trunk nodes cannot be modified directly, and instead are change via a copy-on-write of the root-to-node path together with a change of the root node.

4. Garbage Collection for unlinked branches and filters. After a copy-on-write, the nodes on the old path will be unreferenced. This PR does not GC the trunk nodes themselves, but it includes a GC path to dereference the replaced branches and filters.

5. `platform_batch_rwlock`. Replaces distributed locks using dummy cache pages with a batched distributed RW lock implementation in `platform.[ch]`.

### DIFF
--- a/src/platform_linux/platform.c
+++ b/src/platform_linux/platform.c
@@ -219,6 +219,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
    return CONST_STATUS(ret);
 }
 
+/*
+ *-----------------------------------------------------------------------------
+ * Batch Read/Write Locks
+ *
+ * These are generic distributed reader/writer locks with an intermediate claim
+ * state. These offer similar semantics to the locks in the clockcache, but in
+ * these locks read locks cannot be held with write locks.
+ *
+ * These locks are allocated in batches of PLATFORM_CACHELINE_SIZE / 2, so that
+ * the write lock and claim bits, as well as the distributed read counters, can
+ * be colocated across the batch.
+ *-----------------------------------------------------------------------------
+ */
+
+void
+platform_batch_rwlock_init(platform_batch_rwlock *lock)
+{
+   ZERO_CONTENTS(lock);
+}
+
+/*
+ *-----------------------------------------------------------------------------
+ * lock/unlock
+ *
+ * Drains and blocks all gets (read locks)
+ *
+ * Caller must hold a claim
+ * Caller cannot hold a get (read lock)
+ *-----------------------------------------------------------------------------
+ */
+
+void
+platform_batch_rwlock_lock(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   platform_assert(lock->write_lock[lock_idx].claim);
+   debug_only uint8 was_locked =
+      __sync_lock_test_and_set(&lock->write_lock[lock_idx].lock, 1);
+   debug_assert(!was_locked,
+                "platform_batch_rwlock_lock: Attempt to lock a locked page.\n");
+
+   uint64 wait = 1;
+   for (uint64 i = 0; i < MAX_THREADS; i++) {
+      while (lock->read_counter[i][lock_idx] != 0) {
+         platform_sleep_ns(wait);
+         wait = wait > 2048 ? wait : 2 * wait;
+      }
+      wait = 1;
+   }
+}
+
+void
+platform_batch_rwlock_unlock(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   __sync_lock_release(&lock->write_lock[lock_idx].lock);
+}
+
+/*
+ *-----------------------------------------------------------------------------
+ * try_claim/claim/unlock
+ *
+ * A claim blocks all other claimants (and therefore all other writelocks,
+ * because writelocks are required to hold a claim during the writelock).
+ *
+ * Cannot hold a get (read lock)
+ *
+ * try_claim returns whether the claim succeeded
+ *-----------------------------------------------------------------------------
+ */
+
+bool
+platform_batch_rwlock_try_claim(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   if (__sync_lock_test_and_set(&lock->write_lock[lock_idx].claim, 1)) {
+      return FALSE;
+   }
+   return TRUE;
+}
+
+void
+platform_batch_rwlock_claim(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   uint64 wait = 1;
+   while (!platform_batch_rwlock_try_claim(lock, lock_idx)) {
+      platform_sleep_ns(wait);
+      wait = wait > 2048 ? wait : 2 * wait;
+   }
+}
+
+void
+platform_batch_rwlock_unclaim(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   __sync_lock_release(&lock->write_lock[lock_idx].claim);
+}
+
+/*
+ *-----------------------------------------------------------------------------
+ * get/unget
+ *
+ * Acquire a read lock
+ *-----------------------------------------------------------------------------
+ */
+
+void
+platform_batch_rwlock_get(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   threadid tid = platform_get_tid();
+   while (1) {
+      uint64 wait = 1;
+      while (lock->write_lock[lock_idx].lock) {
+         platform_sleep_ns(wait);
+         wait = wait > 2048 ? wait : 2 * wait;
+      }
+      debug_only uint8 old_counter =
+         __sync_fetch_and_add(&lock->read_counter[tid][lock_idx], 1);
+      debug_assert(old_counter == 0);
+      if (!lock->write_lock[lock_idx].lock) {
+         return;
+      }
+      old_counter = __sync_fetch_and_sub(&lock->read_counter[tid][lock_idx], 1);
+      debug_assert(old_counter == 1);
+   }
+   platform_assert(0);
+}
+
+void
+platform_batch_rwlock_unget(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   threadid         tid = platform_get_tid();
+   debug_only uint8 old_counter =
+      __sync_fetch_and_sub(&lock->read_counter[tid][lock_idx], 1);
+   debug_assert(old_counter == 1);
+}
+
+
 platform_status
 platform_histo_create(platform_heap_id       heap_id,
                       uint32                 num_buckets,
--- a/src/platform_linux/platform_types.h
+++ b/src/platform_linux/platform_types.h
@@ -81,6 +81,45 @@ typedef struct {
 // Spin lock
 typedef pthread_spinlock_t platform_spinlock;
 
+// Distributed Batch RW Lock
+typedef struct {
+   volatile uint8 lock;
+   volatile uint8 claim;
+} platform_claimlock;
+
+typedef struct {
+   platform_claimlock write_lock[PLATFORM_CACHELINE_SIZE / 2];
+   volatile uint8     read_counter[MAX_THREADS][PLATFORM_CACHELINE_SIZE / 2];
+} PLATFORM_CACHELINE_ALIGNED platform_batch_rwlock;
+
+_Static_assert(sizeof(platform_batch_rwlock)
+                  == PLATFORM_CACHELINE_SIZE * (MAX_THREADS / 2 + 1),
+               "Missized platform_batch_rwlock\n");
+
+void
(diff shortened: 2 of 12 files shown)


ajhconway (commented):
Some comments from my self-review.

At a high level, I'm concerned about how long write locks are held on the trunk root. These block reads, so reducing the duration that these locks are held is important.

The main place that this comes up is during memtable incorporation, where the root is flushed and then flushes happen recursively, including node splits and especially (slow) leaf splits. On the other hand, if we do not flush recursively, then each flush will generate a new CoW path, which feels wasteful. Maybe we need a claim here so as to block new write while allowing readers?

src/trunk.c
```diff
@@ -3241,162 +3485,184 @@ trunk_try_continue_incorporate(trunk_handle *spl, uint64 next_generation)
...
+      req->bundle_no);
+   trunk_close_log_stream_if_enabled(spl, &stream);
+   task_enqueue(
+      spl->ts, TASK_TYPE_NORMAL, trunk_bundle_build_filters, req, TRUE);
 
-   // X. Dec-ref the now-incorporated memtable
+   /*
+    * Decrement the now-incorporated memtable ref count and recycle if no
+    * references
+    */
```
ajhconway, line 3665: This garbage collection here is unsafe. Need to lock old root first.

ajhconway: Nevermind, this is fine.

src/trunk.c
```diff
@@ -1054,6 +1056,204 @@ trunk_btree_config(trunk_handle *spl)
...
+   trunk_default_log_if_enabled(
+      spl, "Node copy %lu -> %lu\n", node->addr, node_copy->addr);
+}
+
+/*
+ * Makes a copy of the child indicated by pdata and replaces the parent's
+ * pointer with one to the new child. Returns the new child's page_handle *.
+ */
+static inline void
+trunk_copy_node_and_add_to_parent(trunk_handle     *spl,    // IN
```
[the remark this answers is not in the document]
ajhconway: Changed.

src/trunk.c
```diff
@@ -2605,29 +2866,15 @@ trunk_replace_bundle_branches(trunk_handle             *spl,
                               trunk_branch             *repl_branch,
                               trunk_compact_bundle_req *req)
 {
+   trunk_hdr *hdr = node->hdr;
```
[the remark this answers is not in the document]
ajhconway: Done.

src/trunk.c
```diff
@@ -3241,162 +3485,184 @@ trunk_try_continue_incorporate(trunk_handle *spl, uint64 next_generation)
...
- *  9. Unlock the root
+ *  1. Lock lookup lock (blocks lookups, which must obtain a read lock on the
+ *     lookup lock), root, and Trunk Modification Lock.
+ *  2. Add the memtable to the root as a new compacted bundle
+ *  4. If root is full, flush until it is no longer full. Also flushes any full
+ *     descendents.
+ *  5. If necessary, split the root
+ *  6. Unlock the root
+ *  7. Decrement the now-incorporated memtable ref count and recycle if no
+ *     references
```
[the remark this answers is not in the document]
ajhconway: Done.

src/trunk.c
```diff
@@ -4146,12 +4502,39 @@ trunk_room_to_flush(trunk_handle     *spl,
...
+}
+
 /*
  * flush flushes from parent to the child indicated by pdata.
  *
+ * FLUSH FAILURE DISABLED TEMPORARILY (WILL ASSERT)
  * Failure can occur if there is not enough space in the child.
  *
- * NOTE: parent must be write locked
+ * NOTE: parent must be write locked and TML must be held.
```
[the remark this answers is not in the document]
ajhconway: Done.

src/trunk.c
```diff
@@ -4831,41 +5226,42 @@ trunk_compact_bundle(void *arg, void *scratch_buf)
...
     */
-   uint64 addr             = req->addr;
    uint64 num_replacements = 0;
-   uint64 generation;
-   do {
-      trunk_node_claim(spl->cc, &node);
-      trunk_node_lock(spl->cc, &node);
-
+   bool   should_continue  = TRUE;
+   while (should_continue) {
```
rtjohnso, line 5241: This is basically a `key_compare`.  As above, I think the code will be easier to understand if the predicate is directly in the while condition.

rtjohnso (changes requested):
In addition to the questions and requests below, there's a general comment:

Can you go through and audit new functions to ensure that anything that can fail does in fact return a platform_status and that all the call sites handle failure?  I know existing code doesn't always do that, so you could say that is something that should be addressed in a general error-handling PR, but I'd like us to try to start writing new code that handles errors already.  Feel free to push back on this.

src/platform_linux/platform_types.h
```diff
@@ -81,6 +81,45 @@ typedef struct {
...
 typedef pthread_spinlock_t platform_spinlock;
 
+// Distributed Batch RW Lock
+typedef struct {
+   volatile uint8 lock;
+   volatile uint8 claim;
+} platform_claimlock;
+
+typedef struct {
+   platform_claimlock write_lock[PLATFORM_CACHELINE_SIZE / 2];
```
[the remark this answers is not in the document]
ajhconway: The idea is that we pack the 2-byte structs into a cache line

src/trunk.c
```diff
@@ -3241,162 +3555,179 @@ trunk_try_continue_incorporate(trunk_handle *spl, uint64 next_generation)
...
+    * Lock the lookup lock, blocking lookups.
+    * Transition memtable state and increment memtable generation (blocks
+    * lookups from accessing the memtable that's being incorporated).
+    */
+   memtable_lock_lookup_lock(spl->mt_ctxt);
+   memtable *mt = trunk_get_memtable(spl, generation);
+   // Normally need to hold incorp_mutex, but debug code and also guaranteed no
+   // one is changing gen_to_incorp (we are the only thread that would try)
+   debug_assert(generation == memtable_generation_to_incorporate(spl->mt_ctxt));
+   memtable_transition(
```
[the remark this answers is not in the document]
ajhconway: We have never used `MEMTABLE_STATE_INCORPORATING`. It should just be removed.

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+ *
+ * Caller must hold a claim
+ * Caller cannot hold a get (read lock)
+ *-----------------------------------------------------------------------------
+ */
+
+void
+platform_batch_rwlock_lock(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   platform_assert(lock->write_lock[lock_idx].claim);
```
gapisback, line 239: It will be useful to peel this one-line off into its own 
`platform_batch_rwlock_claimheld(platform_batch_rwlock *lock, uint64 lock_idx)`
inline function which implements the rule doc'ed on L231.

As this interface grows, we will find other places where "claim-held" rule will need to be checked.

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+ *-----------------------------------------------------------------------------
+ * Batch Read/Write Locks
+ *
+ * These are generic distributed reader/writer locks with an intermediate claim
+ * state. These offer similar semantics to the locks in the clockcache, but in
+ * these locks read locks cannot be held with write locks.
+ *
+ * These locks are allocated in batches of PLATFORM_CACHELINE_SIZE / 2, so that
+ * the write lock and claim bits, as well as the distributed read counters, can
+ * be colocated across the batch.
```
gapisback, line 215: Having gone thru this batch of new `platform_batch_rwlock*()` interfaces added with this PR, I think it may be useful to pre-stage just this interface as a stand-alone change-set, augmented with its own set of C-Unit tests.

This kind of API is very amenable to unit-tests. I'm happy to contribute here to peel this off, build unit-tests and work on pre-staging this integration.

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+ */
+
+void
+platform_batch_rwlock_lock(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   platform_assert(lock->write_lock[lock_idx].claim);
+   debug_only uint8 was_locked =
+      __sync_lock_test_and_set(&lock->write_lock[lock_idx].lock, 1);
+   debug_assert(!was_locked,
+                "platform_batch_rwlock_lock: Attempt to lock a locked page.\n");
```
gapisback, line 243: Suggestions: May be it's better to use `%s` for the 1st arg and print using `__FUNCTION__`.

Also, is it really locking a page? Should "lock a locked page" be re-worded?
Will be helpful to print `lock_idx` and whatever else will be useful to identify the "lock"-think we were trying to get a lock on.

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+ * Caller cannot hold a get (read lock)
+ *-----------------------------------------------------------------------------
+ */
+
+void
+platform_batch_rwlock_lock(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   platform_assert(lock->write_lock[lock_idx].claim);
+   debug_only uint8 was_locked =
+      __sync_lock_test_and_set(&lock->write_lock[lock_idx].lock, 1);
```
gapisback, line 241: As implementations become more complex, such kind of checks will really protect us from concurrency issues.

If the actual bit of this "lock"-call overhead is not too much, ... why not test for this in release code, too?

Concurrency and race condition dynamics, often times, change significantly in release binaries whereas they may not show up in debug builds. My experience has been that tightening such checks in lock mgr primitives and keeping them active release builds, also, has a larger pay-off for stress usages.

gapisback, line 241: This is a candidate to be separated out into an bool inline fn, say, `platform_batch_rwlock_lockheld()`, or some variation of `is_lock_held()`, `lock_is_held()`.

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+   return TRUE;
+}
+
+void
+platform_batch_rwlock_claim(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   uint64 wait = 1;
+   while (!platform_batch_rwlock_try_claim(lock, lock_idx)) {
+      platform_sleep(wait);
+      wait = wait > 2048 ? wait : 2 * wait;
```
gapisback, line 289: After seeing multiple instances of this line in this file, it occurred to me that it may be useful to package this in an inline function, e.g., `wait = platform_lock_waitfor(wait);`

That way, if we ever change our minds on how the `wait` value is managed, we need to change it in just one place.

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+   }
+}
+
+void
+platform_batch_rwlock_unclaim(platform_batch_rwlock *lock, uint64 lock_idx)
+{
+   __sync_lock_release(&lock->write_lock[lock_idx].claim);
```
gapisback, line 296: Can we assert that before unclaiming that a `claim` was held; i.e. `platform_batch_rwlock_claimheld()`?

src/platform_linux/platform.c
```diff
@@ -202,6 +202,140 @@ platform_spinlock_destroy(platform_spinlock *lock)
...
+ * These are generic distributed reader/writer locks with an intermediate claim
+ * state. These offer similar semantics to the locks in the clockcache, but in
+ * these locks read locks cannot be held with write locks.
+ *
+ * These locks are allocated in batches of PLATFORM_CACHELINE_SIZE / 2, so that
+ * the write lock and claim bits, as well as the distributed read counters, can
+ * be colocated across the batch.
+ *-----------------------------------------------------------------------------
+ */
+
```
gapisback, line 218: Just a stylistic preference / request: Can you delete this blank line between a function prolog and the start of the function. 

This way, we don't have to worry about dangling fn-/block-comment blocks in future (across code merges).

Issue #474, opened 2022-11-07
Large inserts stress tests with 4 inserting-threads, and background threads, runs into Assertion from trunk_get_new_bundle() : (hdr->end_bundle != hdr->start_bundle)". [3363] No available bundles in trunk node. 


gapisback:
This issue was unearthed while trying to stabilize newly developed `large_inserts_bugs_stress_test` with support for background threads enabled. Background threads support is being stabilized and enabled through test under in-flight issue #469 .

This bug can be reproduced using that feature enablement, testing support and this new test case from the `agurajada/trunk_get_new_bundle-Assert-w-bg-threads` branch.

```
$ build/debug/bin/unit/large_inserts_bugs_stress_test --num-inserts 20000000 --num-threads 4 --num-bg-threads 2 --num-memtable-bg-threads 2 test_seq_key_seq_values_inserts_threaded
Running 1 CTests, suite name 'large_inserts_bugs_stress', test case 'test_seq_key_seq_values_inserts_threaded'.
[...]
Assertion failed at src/trunk.c:2003:trunk_get_new_bundle_no(): "(hdr->end_bundle != hdr->start_bundle)". [3363] No available bundles in trunk node. page disk_addr=524288, end_bundle=7, start_bundle=7
Aborted (core dumped)
```

The problem reproduces albeit a bit sporadically. It is certainly triggered by the use of new background thread-related testing config parameters: i.e., `--num-bg-threads 2 --num-memtable-bg-threads 2`.

All test cases in this new test run cleanly for large #s of inserts per thread w/o the use of background threads. I have exercised this test case for `--num-inserts 20000000` (20 million inserts / thread) and it runs cleanly.

----
From code inspection, this assertion is occurring from L3363, as reported by the enhanced assert message:

```
3325 static void
3326 trunk_memtable_incorporate(trunk_handle  *spl,
[...]
3357    /*
3358     * X. Get a new branch in a bundle for the memtable
3359     */
3360    trunk_compacted_memtable *cmt =
3361       trunk_get_compacted_memtable(spl, generation);
3362    trunk_compact_bundle_req *req = cmt->req;
3363    req->bundle_no                = trunk_get_new_bundle(spl, root);
```

Given that the assert is in the code path of `trunk_memtable_incorporate()`, it's a likely possibility that this is induced by the use of the `--num-memtable-bg-threads 2` argument, which creates 2 threads of type TASK_TYPE_MEMTABLE. 

[...]

Issue #310, opened 2022-03-03
driver_test splinter_test --delete runs into Assertion failed src/trunk.c:1832:trunk_get_new_bundle(): "(hdr->end_bundle != hdr->start_bundle)


gapisback:
Reproduces on /main:

```
Fusion-LocalVM:[22] $ ./bin/driver_test splinter_test --delete --max-async-inflight 20 --num-insert-threads 20 --num-lookup-threads 10
./bin/driver_test: splinterdb_build_version a2768f5d
Dispatch test splinter_test
fingerprint_size: 27
Bumped up IO queue size to 416
Bumped up IO queue size to 416
Running splinter_test with 1 caches
splinter_test: splinter deletion test started with 1                tables
inserting  20% complete for table 0Assertion failed at src/trunk.c:1832:trunk_get_new_bundle(): "(hdr->end_bundle != hdr->start_bundle)". page disk_addr=393216, end_bundle=3, start_bundle=3
Aborted (core dumped)
```

Also ran into the same assertion when running `driver_test splinter_test --seq-perf`

```
Fusion-LocalVM:[30] $ ./bin/driver_test splinter_test --seq-perf --max-async-inflight 20 --num-insert-threads 20 --num-lookup-threads 10
./bin/driver_test: splinterdb_build_version a2768f5d-dirty
Dispatch test splinter_test
fingerprint_size: 27
Bumped up IO queue size to 416
Bumped up IO queue size to 416
Running splinter_test with 1 caches
splinter_test: splinter performance test started with 1                tables
inserting  16% complete for table 0Assertion failed at src/trunk.c:1832:trunk_get_new_bundle(): "(hdr->end_bundle != hdr->start_bundle)". page disk_addr=393216, end_bundle=10, start_bundle=10
Aborted (core dumped)
```

Also ran into the same issue when running `splinter_test --perf` with higher disk capacity `--db-capacity-gib 60`. At lower capacity, we run into Out-of-space errors.

```
Fusion-LocalVM:[49] $ ./bin/driver_test splinter_test --perf --max-async-inflight 10 --num-insert-threads 20 --num-lookup-threads 10 --num-range-lookup-threads 5 --db-capacity-gib 60
./bin/driver_test: splinterdb_build_version a2768f5d-dirty
Dispatch test splinter_test
fingerprint_size: 27
Running splinter_test with 1 caches
splinter_test: splinter performance test started with 1                tables
[...]

PR #497, opened 2022-12-04
Robj/task system config

rtjohnso:
This cleans up several things about the task system:
- Allow using background and foreground threads at the same time
- Create a separate config for whether to do foreground work
-  Introduce a `task_system_config` and all associated initialization, defaults, etc in tests
- Generalize the `--num-bg-threads` command-line option in `driver_test splinter_test` to be available for all of `driver_test`.


src/task.h
```diff
@@ -149,13 +154,11 @@ task_deregister_thread(task_system *ts,
...
-                   task_system       **system,
-                   bool                use_stats,
-                   bool                use_bg_threads,
-                   uint8               num_bg_threads[NUM_TASK_TYPES],
-                   uint64              scratch_size);
+task_system_create(platform_heap_id          hid,
+                   platform_io_handle       *ioh,
+                   task_system             **system,
+                   const task_system_config *cfg,
+                   uint64                    scratch_size);
```
rtjohnso, line 161: Maybe make scratch_size part of config?

src/task.h
```diff
@@ -26,9 +26,10 @@ typedef struct task {
 typedef struct {
    timestamp max_runtime_ns;
    void     *max_runtime_func;
-   uint64    total_latency_ns;
-   uint64    total_tasks;
    uint64    max_latency_ns;
+   uint64    total_latency_ns;
```
rtjohnso, line 30: rename latency to queue_wait_time

rtjohnso: for both

src/task.h
ajhconway: Should we scrap the scratches? It's bending over backwards to reduce `malloc`s, but it's awkward and also since we do it for foreground threads, it means that threads that only read still consume a bunch of memory for no reason.

src/task.c
```diff
@@ -366,93 +368,110 @@ task_thread_create(const char            *name,
    return STATUS_OK;
 }
 
-/* Worker function for the background task pool. */
-static void
-task_worker_thread(void *arg)
+/* Caller must hold lock on the group. */
+static inline task *
```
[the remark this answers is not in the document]
ajhconway: I don't see why this needs inlining

src/task.c
```diff
@@ -610,91 +606,45 @@ task_enqueue(task_system *ts,
...
 static inline platform_status
-task_group_perform_one(task_group *group)
+task_group_perform_one(task_group *group, uint64 min_backlog)
 {
-   platform_assert(!task_system_use_bg_threads(group->ts));
    platform_status rc;
    task           *assigned_task = NULL;
-   task_queue     *tq            = &group->tq;
-   if (group->current_outstanding_tasks == 0) {
+   if (group->current_outstanding_tasks <= min_backlog) {
```
ajhconway, line 628: Is timed out the best status to return here?

src/task.c
```diff
@@ -462,61 +481,47 @@ task_group_stop_and_wait_for_threads(task_group *group)
...
          goto out;
       }
    }
    return STATUS_OK;
 
 out:
    debug_assert(!SUCCESS(rc));
-   if (use_bg_threads) {
-      platform_condvar_destroy(&group->bg.cv);
+   if (num_bg_threads) {
```
rtjohnso, line 523: We always init condvar so we should always destroy it.  i.e. delete the if.

src/task.h
ajhconway: +1

src/task.c
```diff
@@ -610,91 +606,45 @@ task_enqueue(task_system *ts,
...
-         while (group->current_outstanding_tasks != 0) {
-            platform_sleep(USEC_TO_NSEC(100000)); // 100 msec.
-         }
-      }
-   }
+   platform_status rc;
+   do {
+      rc = task_perform_one(ts);
+      debug_assert(SUCCESS(rc) || STATUS_IS_EQ(rc, STATUS_TIMEDOUT));
+   } while (STATUS_IS_NE(rc, STATUS_TIMEDOUT));
```
rtjohnso, line 613: Add comment explaining that STATUS_TIMEDOUT is being repurposed here to indicate that there are no tasks to run.

tests/config.c
```diff
@@ -51,14 +54,17 @@ config_set_defaults(master_config *cfg)
       .allocator_capacity       = GiB_TO_B(TEST_CONFIG_DEFAULT_DISK_SIZE_GB),
       .cache_capacity           = GiB_TO_B(TEST_CONFIG_DEFAULT_CACHE_SIZE_GB),
       .btree_rough_count_height = 1,
-      .filter_remainder_size    = 6,
+      .filter_remainder_size    = 9,
```
ajhconway, line 57: This is just going to trigger the autoconfig which sets it to 4. 9 is actually going to larger in practice because filters very rarely cover an entire node.

src/task.c
```diff
@@ -610,91 +606,45 @@ task_enqueue(task_system *ts,
 void
```
rtjohnso, line 606: Add comment (in task.h) explaining that this is useful at shutdown time.

ajhconway (approved):
Very happy with these changes.

The main thing I want to mention is that you can't actually increase the fingerprint size in the filters. It'll actually get autoconfig'd down to 4 anyway.

src/task.c
```diff
@@ -707,17 +657,48 @@ task_group_perform_one(task_group *group)
...
+   return rc;
+}
+
+platform_status
+task_perform_one_if_needed(task_system *ts)
+{
+   platform_status rc = STATUS_OK;
+   for (task_type type = TASK_TYPE_FIRST; type != NUM_TASK_TYPES; type++) {
+      rc = task_group_perform_one(&ts->group[type],
+                                  ts->group[type].bg.num_threads);
```
rtjohnso, line 676: Comment explaining what this policy means

src/task.c
```diff
@@ -707,17 +657,48 @@ task_group_perform_one(task_group *group)
 platform_status
```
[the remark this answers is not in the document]
rtjohnso: Maybe explain why we have separate memtable and normal task groups. (in header)

src/task.c
```diff
@@ -707,17 +657,48 @@ task_group_perform_one(task_group *group)
...
    }
    return rc;
 }
 
+platform_status
+task_system_config_init(task_system_config *task_cfg,
+                        bool                use_stats,
+                        const uint64        num_bg_threads[NUM_TASK_TYPES])
+{
+   if ((num_bg_threads[TASK_TYPE_NORMAL] == 0)
```
rtjohnso, line 689: Maybe relax this constraint if they have foreground threads?

src/task.h
rtjohnso: Moved scratch into config for now, but it will go away in a subsequent PR.

src/task.c
rtjohnso: The only other plausible one I see is NOT_FOUND.  TIMEDOUT is what was used before, and I could see it generalizing in the future if we add time limits to some of these functions.

rtjohnso:
This version makes several additional changes:

- During thread creation, threadid's and scratch space are allocated by the thread performing the creation, not by the new thread.  This allows us to report errors (e.g. too many threads) to the the function asking to create a child thread rather than silently failing.
- The above also fixes a race condition in one of the tests that assumed thread ids were allocated in order of thread_create calls, which might not have been the case when each child thread allocated its own tid.
- It introduces a new `task_perform_until_quiescent` API that runs tasks until all tasks have been completed, not just until the task queue is empty.  This solves some race conditions with splinter shutdown.
- This also cleans up tid assignment and the use of tid == 0 to indicate that a thread is not registered.
- Additional cleanups and improved error handling.

src/platform_linux/platform.c
```diff
@@ -7,7 +7,7 @@
 
 #include <sys/mman.h>
 
-__thread threadid xxxtid;
+__thread threadid xxxtid = INVALID_TID;
```
gapisback, line 10: Thanks for doing this initialization. It will now simplify the way I was intending to fix the issue raised in #499

src/task.c
```diff
@@ -747,25 +686,116 @@ task_group_perform_one(task_group *group)
...
+platform_status
+task_system_config_init(task_system_config *task_cfg,
+                        bool                use_stats,
+                        const uint64        num_bg_threads[NUM_TASK_TYPES],
+                        uint64              scratch_size)
+{
+   if ((num_bg_threads[TASK_TYPE_NORMAL] == 0)
+       != (num_bg_threads[TASK_TYPE_MEMTABLE] == 0))
+   {
+      return STATUS_BAD_PARAM;
```
gapisback, line 779: This refactoring to check `task_system_config_init()` separately here loses the error message that the other fn, `task_config_valid()` calls on L 801.

I think it will be helpful to move that error message here so that we don't get a silent failure due to an illegal configuration.

gapisback: Also, `task_config_valid()` checks for:

```
 810    if ((normal_bg_threads + memtable_bg_threads) >= MAX_THREADS) {
 811       platform_error_log("Total number of background threads configured"
```

Don't you want to move that check here? 

If the expectation is that anywhere in Splinter main code or in the test code, before we call `task_system_create()` that the code would [should] have called your new function `task_system_config_init()`, then we could do something like this:

- Have `task_system_config_init()` call `task_config_valid()` to do all the required checks. The callee will do the required error reporting.
- then once validation succeeds, continue with `task_system_config_init()`.

src/splinterdb.c
```diff
@@ -188,6 +189,15 @@ splinterdb_init_config(const splinterdb_config *kvs_cfg, // IN
 
    shard_log_config_init(&kvs->log_cfg, &kvs->cache_cfg.super, kvs->data_cfg);
 
+   uint64 num_bg_threads[NUM_TASK_TYPES] = {0};
+   num_bg_threads[TASK_TYPE_MEMTABLE]    = kvs_cfg->num_memtable_bg_threads;
+   num_bg_threads[TASK_TYPE_NORMAL]      = kvs_cfg->num_normal_bg_threads;
+   rc                                    = task_system_config_init(
```
gapisback, line 195: Minor request: You may want to leave a blank line before this line, so the call to `task_system_config_init()` looks like a function call, and does not fall into the grips of clang-format indentation. (Right now, the code is a bit unreadable.)

src/task.c
```diff
@@ -50,100 +40,71 @@ task_init_tid_bitmask(uint64 *tid_bitmask)
...
- * defined as a thread's hook and gets run as part of the thread's
- * registration. This function initializes each thread's thread-ID, which
- * is really an index into the array of thread-specific data structures.
+/*
+ * Allocate a threadid
  */
-static void
-task_init_threadid(task_system *ts)
+static threadid
+allocate_threadid(task_system *ts)
```
gapisback, line 68: Request: Most other functions in this file have the `task_` prefix. Can you consider naming this as `task_allocate_threadid()` and its sibling as `task_deallocate_threadid()`?

src/task.c
```diff
@@ -50,100 +40,71 @@ task_init_tid_bitmask(uint64 *tid_bitmask)
...
 {
-   threadid  tid         = INVALID_TID;
-   uint64   *tid_bitmask = task_system_get_tid_bitmask(ts);
-   threadid *max_tid     = task_system_get_max_tid(ts);
+   threadid  tid        = INVALID_TID;
+   uint64   *tid_bitset = task_system_get_tid_bitmask(ts);
```
gapisback, line 71: Any specific reason why the variable was renamed from `tid_bitmask` to `tid_bitset`?

I think the term "bitmask" is fairly standard and is consistent with the name of the fn itself; i.e. `task_system_get_tid_bitmask()`.

I think `bitset` could be mis-interpreted as "bit is set" (some form of boolean) or bit number that is set etc.

Can you consider retaining the `bitmask` term, and also rename the newly added variables to `old_bitmask` and `new_bitmask`?

src/task.c
```diff
@@ -50,100 +40,71 @@ task_init_tid_bitmask(uint64 *tid_bitmask)
...
+   return *task_system_get_tid_bitmask(ts);
+}
 
- * The first thing a newly started thread that wants to use Splinter is expected
- * to do is to "register itself" with the task system. This function is
- * defined as a thread's hook and gets run as part of the thread's
- * registration. This function initializes each thread's thread-ID, which
- * is really an index into the array of thread-specific data structures.
+/*
+ * Allocate a threadid
```
gapisback, line 65: You may want to elaborate this a bit more to mention that this function will return `INVALID_TID` to signal a condition when all possible configured threads are already allocated (in-use).

src/task.c
```diff
@@ -50,100 +40,71 @@ task_init_tid_bitmask(uint64 *tid_bitmask)
...
-task_init_threadid(task_system *ts)
+static threadid
+allocate_threadid(task_system *ts)
 {
-   threadid  tid         = INVALID_TID;
-   uint64   *tid_bitmask = task_system_get_tid_bitmask(ts);
-   threadid *max_tid     = task_system_get_max_tid(ts);
+   threadid  tid        = INVALID_TID;
+   uint64   *tid_bitset = task_system_get_tid_bitmask(ts);
+   threadid *max_tid    = task_system_get_max_tid(ts);
```
gapisback, line 72: As this variable is not used till much later, for readability, consider moving this line to just before L95 below.

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+task_register_thread(task_system *ts,
+                     uint64       scratch_size,
+                     const char  *file,
+                     const int    lineno,
+                     const char  *func)
+{
+   threadid thread_tid;
+
+   debug_code(thread_tid = platform_get_tid());
+   debug_assert(
```
gapisback, line 352: Can we not make this a `platform_assert()`, so that any integrations using release binary will also get this check?

Thread registration is a one-time event in the life of a thread so, I think, the overhead of an additional assertion check (which this code adds in debug-builds) is very minor.

While building task-system unit-tests, and also during other integration, I found this business of registration needs to be done carefully. It's better to tighten this check for release builds, too,  as a safeguard.

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+      thread_tid);
+
+   thread_tid = allocate_threadid(ts);
+   if (thread_tid == INVALID_TID) {
+      return STATUS_NO_SPACE;
+   }
+   platform_set_tid(thread_tid);
+
+   debug_assert(ts->thread_scratch[thread_tid] == NULL,
+                "Scratch space should not yet exist.");
```
gapisback, line 367: The older version of this message was raised under `platform_assert()`. For the same reasons given a few lines earlier, can't we always do this assertion check even for release builds?

Also, the older message printed out the calling function's info, too: file, lineno, func. I found that information useful while debugging. Can we extend this message to print those values too?

Also, it will certainly be helpful to print the offending `thread_tid` which ran into this situation.

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+      file,
+      lineno,
+      func,
+      thread_tid);
+
+   thread_tid = allocate_threadid(ts);
+   if (thread_tid == INVALID_TID) {
+      return STATUS_NO_SPACE;
+   }
+   platform_set_tid(thread_tid);
```
gapisback, line 364: Suggestion for small cleanup: If you could move this line to just before the call to `task_run_thread_hooks()` on L379, then you could remove the call to `platform_set_tid(INVALID_TID);` on L372 below.

All this business of `set_tid()` to some valid value and then to back it out in case of an error could be simplified.

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+ */
+void
+task_deregister_thread(task_system *ts,
+                       const char  *file,
+                       const int    lineno,
+                       const char  *func)
+{
+   threadid tid = platform_get_tid();
+
+   debug_assert(tid != INVALID_TID,
```
gapisback, line 399: Can this be changed to `platform_assert()`? I think, as requested for the checks in registration, these events are not that frequent, so it's better to tighten these checks for release builds, too.

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+                "Error! Attempt to deregister unregistered thread.\n");
```
gapisback, line 400: The message printed in the old code was also reporting the caller's info; i.e. file, lineno, func. As this function `task_deregister_this_thread()` is called from more than one place, I think it's useful to report that call-site info too.

Any reason why you chose to not continue with the old code's message printing style?

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+ * Background task management
+ ****************************************/
+
+/* Caller must hold lock on the group. */
+static task *
+task_group_get_one_task(task_group *group)
+{
+   task_queue *tq            = &group->tq;
+   task       *assigned_task = NULL;
+   if (group->current_waiting_tasks != 0) {
```
gapisback, line 422: Would you consider changing this to check `if group->current_waiting_tasks == 0) { return assigned_task; }`

Then, the rest of the main body of this function could be tabbed over one stop to the left, w/o having to be nested under an `if()` which has no matching `else{}`.

src/task.h
```diff
@@ -25,9 +25,11 @@ typedef struct task {
 typedef struct {
    timestamp max_runtime_ns;
    void     *max_runtime_func;
-   uint64    total_latency_ns;
-   uint64    total_tasks;
-   uint64    max_latency_ns;
+   uint64    max_queue_wait_time_ns;
```
gapisback, line 28: As long as you are renaming these variables, I suggest you consider renaming these as:

- `queue_wait_time_hwm_ns`
- `outstanding_tasks_hwm`

The `hwm` is a commonly understood acronym for High-Water Mark and is usually used to denote such fields that track the 'max-value of something".

src/task.c
```diff
@@ -395,6 +325,154 @@ task_thread_create(const char            *name,
...
+   deallocate_threadid(ts, tid); // allow thread id to be re-used
+}
+
+/****************************************
+ * Background task management
+ ****************************************/
+
+/* Caller must hold lock on the group. */
+static task *
+task_group_get_one_task(task_group *group)
```
gapisback, line 418: As the return is simply one `task *`, suggest renaming this simply to `task_group_get_task()` or if you prefer, call it `task_group_get_next_task()`. [ I prefer the 1st option. ]

src/task.c
```diff
@@ -471,124 +517,73 @@ task_worker_thread(void *arg)
...
+   platform_condvar_lock(&group->cv);
 
    platform_assert(group->tq.head == NULL);
    platform_assert(group->tq.tail == NULL);
-   platform_assert(group->current_outstanding_tasks == 0);
-
-   if (!use_bg_threads) {
-      return;
-   }
+   platform_assert(group->current_waiting_tasks == 0);
```
gapisback, line 524: Please print the `group->current_waiting_tasks` when the assert fails.

src/task.c
```diff
@@ -471,124 +517,73 @@ task_worker_thread(void *arg)
...
 }
 
-/*
- * task_lock_task_queue(), task_unlock_task_queue():
- *
- * Primitives to lock / unlock task queues depending on whether we
- * are dealing with foreground or background threads.
- */
-static inline platform_status
-task_lock_task_queue(task_group *group)
```
gapisback, line 571: I understand that you rework makes these two lock / unlock functions unnecessary. 

However, would you consider retaining these shell inline functions as wrappers to `platform_mutex_lock/unlock()` interfaces, so that one can set a debugger breakpoint in these wrappers?

src/task.c
```diff
@@ -747,25 +686,116 @@ task_group_perform_one(task_group *group)
...
+   }
+   return rc;
+}
+
+platform_status
+task_perform_one_if_needed(task_system *ts)
+{
+   platform_status rc = STATUS_OK;
+   for (task_type type = TASK_TYPE_FIRST; type != NUM_TASK_TYPES; type++) {
+      /* Perform a task only if there are more waiting tasks than bg
```
gapisback, line 705: Maybe move this hidden common as prolog comment block, formatted like it's done for `task_perform_one()`.

src/task.c
```diff
@@ -632,109 +627,53 @@ task_enqueue(task_system *ts,
...
-   if (group->current_outstanding_tasks == 0) {
+   if (group->current_waiting_tasks <= min_backlog) {
       return STATUS_TIMEDOUT;
    }
 
-   platform_thread_cleanup_push(task_queue_unlock, group);
-   rc = platform_mutex_lock(&group->fg.mutex);
+   platform_thread_cleanup_push((void (*)(void *))platform_condvar_unlock,
+                                &group->cv);
+   rc = platform_condvar_lock(&group->cv);
```
gapisback, line 657: If this fails, then on L659    we do: `   goto out;`

Then, in the `out:` block, on L 669  we may attempt to do: `platform_condvar_unlock(&group->cv);`
That may also complain.

It's difficult to repro this situation, but from code-walk through there may be a latent problem. 

I think some subtle rearrangement of the placement of code under `out: ` label is needed.

I could not immediately think of what rework to suggest, as it involved handling properly `assigned_task` and `__sync_fetch_and_add()` calls.

src/task.c
```diff
@@ -799,58 +829,46 @@ task_config_valid(uint64 num_bg_threads[NUM_TASK_TYPES])
...
-                   uint64              num_bg_threads[NUM_TASK_TYPES],
-                   uint64              scratch_size)
-{
-   platform_status rc = task_config_valid(num_bg_threads);
+task_system_create(platform_heap_id          hid,
+                   platform_io_handle       *ioh,
+                   task_system             **system,
+                   const task_system_config *cfg)
+{
+   platform_status rc = task_config_valid(cfg);
```
gapisback, line 837: Not sure how best to resolve this duplication of cfg-validation checks but here's my thought:

Even if you add a call to `task_config_valid()` in `task_system_config_init()`, I think it may still be ok to retain this call on L837. It's a small function doing a quick check.

It will cover the odd case if some test / other code were to call `task_system_create()` with a `cfg` which would fail the checks in this validator function.

src/task.c
```diff
@@ -632,109 +627,53 @@ task_enqueue(task_system *ts,
...
-{
-   task_group *group = (task_group *)arg;
-   platform_assert(!task_system_use_bg_threads(group->ts));
-   platform_mutex_unlock(&group->fg.mutex);
-}
-
-static inline platform_status
-task_group_perform_one(task_group *group)
+static platform_status
+task_group_perform_one(task_group *group, uint64 min_backlog)
```
gapisback, line 647: Would be helpful to add a prolog documenting the behaviour controlled by `min_backlog` argument.

src/task.c
gapisback: Agreed; some comments will be helpful. I might put them in the function's prolog.

I do have a question whether this heuristic will address some of the issues we've seen when running large-inserts stress test. For example, under issue #474 (and perhaps, similar configuration problem is also the RCA for the issue #310), we have a situation where trunk # bundles is inadequate.

Does it likely that such problems may be resolved with foreground threads picking up some work based on configured # of bg-threads?

I guess this policy may need to be tweaked a bit to get more such stress tests running stably.

I am wondering (just asking aloud ...) whether we should make fg-threads be more aggressive by changing the `min_backlog` arg to `ts->group[type].bg.num_threads / 2`? 

I know it's all a bit arbitrary. Just wanted to hear your thoughts on how we could best refine this heuristic.

src/task.h
```diff
@@ -172,12 +198,45 @@ task_enqueue(task_system *ts,
...
+ * Performs one task if there is one waiting.  Otherwise returns
+ * immediately.  Returns STATUS_TIMEDOUT to indicate that there was no
+ * task to run.
+ */
 platform_status
 task_perform_one(task_system *ts);
 
+/*
+ * Like task_perform_one, except it only runs a task if the system
+ * appears to be falling behind.
```
gapisback, line 211: Do you want to add: `" ... falling behind based on internal heuristics."` ?

tests/config.c
```diff
@@ -25,6 +25,9 @@
 #define TEST_CONFIG_DEFAULT_FANOUT                8
 #define TEST_CONFIG_DEFAULT_MAX_BRANCHES_PER_NODE 24
 
+#define TEST_CONFIG_DEFAULT_NUM_BACKGROUND_THREADS          (0)
```
gapisback, line 28: I think this and the next one can come out.

I think in a previous check-in to this file, I've added `TEST_CONFIG_DEFAULT_NUM_NORMAL_BG_THREADS` and `TEST_CONFIG_DEFAULT_NUM_MEMTABLE_BG_THREADS`, but forgot to take these two out.

tests/config.c
```diff
@@ -115,6 +119,12 @@ config_usage()
    platform_error_log("\t--cache-capacity-mib (%d)\n",
                       (int)(TEST_CONFIG_DEFAULT_CACHE_SIZE_GB * KiB));
    platform_error_log("\t--cache-debug-log\n");
+   platform_error_log("\t--perform-bg-tasks\n");
+   platform_error_log("\t--no-perform-bg-tasks\n");
+   platform_error_log("\t--num-background-threads (%d)\n",
```
gapisback, line 124: I think these 4 lines can come out as these are not the supported args: `--num-background-threads` and `--num-background-memtable-threads`.

The ones I finalized in a previous check-in are on lines 137 & 140: `--num-normal-bg-threads` and `--num-memtable-bg-threads`.

The choice of these `--args` is to keep their names consistent with the fields in `master_config{}` that store these config params.

tests/config.c
```diff
@@ -344,6 +366,13 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
                                (MAX_PAGES_PER_EXTENT * cfg[cfg_idx].page_size));
             return STATUS_BAD_PARAM;
          }
+         if ((cfg[cfg_idx].num_normal_bg_threads == 0)
```
gapisback, line 369: This check appears below on L384. I had checked that in as part of the previous rework to this file to support bg-threads.

tests/functional/splinter_test.c
```diff
@@ -2383,9 +2383,6 @@ usage(const char *argv0)
       "\t%s --cache-per-table\n"
       "\t%s --parallel-perf --max-async-inflight [num] --num-pthreads [num] "
       "--lookup-positive-percent [num] --seed [num]\n"
-      "\t%s --num-normal-bg-threads (number of normal background threads)\n"
```
gapisback, line 2386: Any reason why these lines were deleted? I thought these two params are supported when invoking this test-case.

I do see one case invoking this combination in `test.sh` on L555. So, they are valid outputs in `--help`:

```
555     run_with_timing "Functionality test, default key size, with background threads" \
556         "$BINDIR"/driver_test splinter_test --functionality 1000000 100 \
557                                             --num-normal-bg-threads 4 --num-memtable-bg-threads 2 \
558                                             --seed "$SEED"
```

tests/unit/btree_test_common.c
```diff
@@ -49,6 +49,20 @@ init_clockcache_config_from_master_config(clockcache_config *cache_cfg,
...
+int
+init_task_config_from_master_config(task_system_config  *task_cfg,
+                                    const master_config *master_cfg,
+                                    uint64               scratch_size)
+{
+   platform_status rc;
+   uint64          num_bg_threads[NUM_TASK_TYPES] = {0};
+   num_bg_threads[TASK_TYPE_NORMAL]   = master_cfg->num_normal_bg_threads;
+   num_bg_threads[TASK_TYPE_MEMTABLE] = master_cfg->num_memtable_bg_threads;
+   rc                                 = task_system_config_init(
```
gapisback, line 61: Minor request: If you can insert a blank line before this one, this fn-call will look like a function call.

tests/unit/task_system_test.c
```diff
@@ -478,18 +477,16 @@ create_task_system_without_bg_threads(void *datap)
...
-                           TRUE, // Use statistics,
-                           data->num_bg_threads,
-                           trunk_get_scratch_size());
+   // no background threads by default.
+   uint64 num_bg_threads[NUM_TASK_TYPES] = {0};
+   rc = task_system_config_init(&data->task_cfg,
+                                TRUE, // use stats
+                                num_bg_threads,
+                                trunk_get_scratch_size());
+   platform_assert(SUCCESS(rc));
```
gapisback, line 488: It may be better to change this to `ASSERT_TRUE()` so that if this `rc` is not SUCCESS, the test will fail, rather than asserting.

(I have tried to only use `platform_assert()` to check return status from stuff not directly related to this test itself; e.g. on L 127 in this file.)

tests/unit/task_system_test.c
```diff
@@ -503,20 +500,21 @@ create_task_system_with_bg_threads(void  *datap,
...
-   data->num_bg_threads[TASK_TYPE_MEMTABLE] = num_memtable_bg_threads;
-   data->num_bg_threads[TASK_TYPE_NORMAL]   = num_normal_bg_threads;
+   uint64 num_bg_threads[NUM_TASK_TYPES] = {0};
+   num_bg_threads[TASK_TYPE_MEMTABLE]    = num_memtable_bg_threads;
+   num_bg_threads[TASK_TYPE_NORMAL]      = num_normal_bg_threads;
+   rc = task_system_config_init(&data->task_cfg,
+                                TRUE, // use stats
+                                num_bg_threads,
+                                trunk_get_scratch_size());
+   platform_assert(SUCCESS(rc));
```
gapisback, line 515: Better to idiomatically use `ASSERT_TRUE(SUCCESS(rc))` here.

gapisback:
@rtjohnso - I am done with this review. I found it a bit difficult to review as the diff / diff-blocks were quite a few.

Before applying any corrections or changes to address review feedback, it will be helpful if you could, perhaps, squash all 15 commits into one commit. Then, apply your changes, so I can see the delta-diffs in just one new commit.

Hopefully, I'll get one more short for a quick pass on the delta-diffs after you've made all changes. Let me know when your next round of updates are done.

Thanks, for all this work!

src/task.c
rtjohnso: I agree the policy is a bit arbitrary.  However, I don't think this is the place to solve liveness bugs in trunk.c.

src/task.c
rtjohnso: I actually think static functions should not be required to have the module name as a prefix (and in fact, I think they probably should not have the module name as a prefix), but I'm ok either way.  Changed for now.

src/task.c
rtjohnso: A bitmask is a value that is used to select a subset of bits in an integer, via and, or, etc.
A bitset represents a set.

I think tid_bitmask is really a set.  But since it reverses the usual meaning of 0 and 1, I'm ok changing these back to bitmask.

src/task.c
rtjohnso: Per our discussion today I think we should get rid of it.  But restoring it for now.

src/task.h
rtjohnso: Never heard of that.  Already discussed this rename in previous convos.

src/task.c
rtjohnso: Good catch.

It turns out there are no cancellation points in the critical section, so we don't need to push a thread cleanup function at all.

Cleaned up based on that.

src/task.c
```diff
@@ -747,25 +686,116 @@ task_group_perform_one(task_group *group)
...
+   }
+
+   return result;
+}
+
+void
+task_perform_until_quiescent(task_system *ts)
+{
+   while (!task_system_is_quiescent(ts)) {
+      task_perform_all(ts);
```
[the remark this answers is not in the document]
rtjohnso: Cleaned up a bit along these lines.

tests/config.c
rtjohnso: Agreed, except that --perform-bg-tasks controls whether _foreground_ threads perform bg tasks.  It is in use.

tests/functional/splinter_test.c
rtjohnso: These have been replaced by the more general config.c system for specifying background threads.

rtjohnso:
The change today responds to all of Aditya's comments and also tweaks the way foreground threads do background work so that there is no longer a constraint on how background threads for different task groups are configured in the task system.

src/task.c
```diff
@@ -640,72 +661,69 @@ task_enqueue(task_system *ts,
...
-   assigned_task = task_group_get_one_task(group);
+   assigned_task = task_group_get_next_task(group);
 
-out:
+   /* It is important to update the current_executing_tasks while
+      holding the lock. The reason is that, if we release the lock
+      before updating current_executing_tasks, then another thread
+      might observe that both the number of enqueued tasks and the
+      number of executing tasks are both 0, and hence that the system
+      is quiescent, even though it is not. */
```
gapisback, line 697: Minor, but .. could you please format the comment block to conform the style we follow elsewhere for multi-line comment blocks? Thanks!

src/task.c
```diff
@@ -752,39 +770,28 @@ task_system_is_quiescent(task_system *ts)
...
    }
 
    return result;
 }
 
-void
+platform_status
 task_perform_until_quiescent(task_system *ts)
 {
+   int wait = 1;
```
gapisback, line 783: Thanks for adding some sort of defensive looping mechanism here.

A suggestion: You may want to declare this as `uint64` as that's what `platform_sleep()` requires. And if possible, for readability, rename this as `wait_ns`.

src/task.c
```diff
@@ -640,72 +661,69 @@ task_enqueue(task_system *ts,
       }
    }
    platform_condvar_signal(&group->cv);
-   return platform_condvar_unlock(&group->cv);
+   return task_group_unlock(group);
 }
 
+/*
+ * Run a task if the number of waiting tasks is at least min_backlog.
```
gapisback, line 668: The comment is no longer consistent with the introduction of the new `queue_scale_percent` argument.

I found it a bit difficult to ferret out the semantic of this variable. 

Can you please add some comments here explaining what this %age variable means? And give some guidance on what it should be set to, if ever someone wishes to change it from its current default of 100% (`TEST_CONFIG_DEFAULT_QUEUE_SCALE_PERCENT`)?

tests/config.c
```diff
@@ -46,6 +43,7 @@
 #define TEST_CONFIG_DEFAULT_NUM_NORMAL_BG_THREADS   0
 #define TEST_CONFIG_DEFAULT_NUM_MEMTABLE_BG_THREADS 0
 
+#define TEST_CONFIG_DEFAULT_QUEUE_SCALE_PERCENT (100)
```
gapisback, line 46: It will be helpful to document here what the expected value for this newly added `--queue-scale-percent` config param should be.

I am assuming the expected value is an integer within the range of [0 - 100]. If so, it will be helpful to document the expected range of valid values.

rtjohnso: Any number is valid.  No other parameter is documented.

tests/config.c
```diff
@@ -249,18 +243,7 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
...
-               cfg[cfg_idx].perform_bg_tasks = TRUE;
-            }
-         }
-         config_has_option("no-perform-bg-tasks")
-         {
-            for (uint8 cfg_idx = 0; cfg_idx < num_config; cfg_idx++) {
-               cfg[cfg_idx].perform_bg_tasks = FALSE;
-            }
-         }
+         config_set_uint64("queue-scale-percent", cfg, queue_scale_percent) {}
```
gapisback, line 246: I don't see any validation checks further below to ensure that bogus values do not creep through.

I think, it will be useful to have some checks in the validation checks block from L326 to check for unsupportable or bogus values.

src/task.c
```diff
@@ -640,72 +661,69 @@ task_enqueue(task_system *ts,
...
    }
 
    return rc;
 }
 
 /*
- * task_perform_one() - Checks if there is a task to run and runs it.
- * Returns ETIMEDOUT if it didn't run any task;
+ * Perform a task only if there are more waiting tasks than bg
+ * threads.
```
gapisback, line 721: Request to clarify this comment: "_if there are more waiting tasks than bg threads._" ... is accurate if the queue_scale_percent is at its default, 100%.

But the actual behaviour implemented by `task_group_perform_one()` to decide "if_needed" semantic depends on this new param added here.

Sorry to be nit picky ... but it will help correct the comment and somehow cast the behaviour of "if_needed" in terms of this new argument, so that the lay reader is not mislead.

src/task.c
```diff
@@ -819,6 +816,21 @@ task_config_valid(const task_system_config *cfg)
    return STATUS_OK;
 }
 
+platform_status
```
gapisback, line 821: There is one remaining call to `task_system_config_init()`, which now can return an `rc`, that still needs to be checked in the caller. See the call in `functional/test.h` in `test_config_init()`. Minor cleanup pending ...

src/task.c
```diff
@@ -819,6 +816,21 @@ task_config_valid(const task_system_config *cfg)
...
+                        const uint64        num_bg_threads[NUM_TASK_TYPES],
+                        uint64              scratch_size)
+{
+   task_cfg->use_stats    = use_stats;
+   task_cfg->scratch_size = scratch_size;
+
+   memcpy(task_cfg->num_background_threads,
+          num_bg_threads,
+          NUM_TASK_TYPES * sizeof(num_bg_threads[0]));
+   return task_config_valid(task_cfg);
```
gapisback, line 833: Wouldn't it be better to see if `task_config_valid()` is true, and only then do the `memcpy()`? In other words, why not return immediately up front on L827 if `task_config_valid()` returns something other than `STATUS_OK`?

That way, we won't be polluting the output `task_cfg`'s contents in case of invalid configurations.

tests/config.c
rtjohnso: Any value is allowed.

src/task.h
```diff
@@ -199,19 +196,35 @@ task_enqueue(task_system *ts,
...
+ * Possibly performs one background task if there is one waiting,
+ * based on the specified queue_scale_percent.  Otherwise returns
+ * immediately.  Returns STATUS_TIMEDOUT to indicate that it did not
+ * run any task.  STATUS_OK indicates that it did run a task.
+ * queue_scale_percent specifies how big the queue must be, relative
+ * to the number of background threads for that task group, for us to
+ * perform a task in that group.  So, for example,
+ * - A queue_scale_percent of 0 means always perform a task if one is
+     waiting.
+ * - A queue_scale_percent of 100 means perform a task if there are as
```
gapisback, line 208: minor: Should this line read: "... there are at least as many waiting tasks ..." ?

tests/functional/splinter_test.c
```diff
@@ -2852,8 +2852,10 @@ splinter_test(int argc, char *argv[])
          platform_assert(SUCCESS(rc));
          break;
       case parallel_perf:
-         platform_assert(max_async_inflight == 0
-                         || task_system_use_bg_threads(ts));
+         platform_assert(
+            max_async_inflight == 0
+            || (0 < task_cfg.num_background_threads[TASK_TYPE_MEMTABLE]
+                && 0 < task_cfg.num_background_threads[TASK_TYPE_NORMAL]));
```
gapisback, line 2858: I am a bit confused with this check.

I am not quite following what the new rules is. Previous to this PR, we needed both bg-thread config params to be 0 or both to be non-zero.

And I sort of understood that with this newest round you are relaxing that check. Aren't you now allowing only one  of the 2 params to be set?

If so, should the check on L2858 be `||` ?

rtjohnso: I actually don't know what is required by the async code, so I just conservatively required both to be non-zero, which is what was being required before anyway.

gapisback (approved):
HI, @rtjohnso - I've taken a quick look at the latest round of diffs you pushed based on my earlier feedback from last night.

I scribbled some minor cleanup comments / suggestion in this last pass, and will leave it up to you to address / incorporate as appropriate.

Thank you very much for patiently plodding through all comments and incorporating changes.

Overall, this whole fg-/bg-threads stuff is a lot better now, so am looking forward to running more tests once this integrates.

include/splinterdb/splinterdb.h
```diff
@@ -50,6 +50,23 @@ typedef struct {
    bool        cache_use_stats;
    const char *cache_logfile;
 
+   // task system
+   // Background threads configuration: Both have to be non-zero in order for
```
gapisback, line 54: @rtjohnso -- heads-up: Seems like you've removed this restriction that both have to be non-zero (or zero).

This comment block could come out now, otherwise, it will be inaccurate.

PR #502, opened 2022-12-20
Document cache.h interfaces based on clockcache comments.

jonhnet:
I tried to document all the public methods on the cache.h polymorphic interface. Please see XXXs for places where I don't even have a guess. But also please check that the other text I wrote is actually correct.

### DIFF
--- a/src/cache.h
+++ b/src/cache.h
@@ -506,10 +506,6 @@ cache_mark_dirty(cache *cc, page_handle *page)
  *
  * This is a performance optimization, used when the caller knows this page
  * will be needed again very soon.
- *
- * XXX I note that clockcache_evict_all has an !ignore_pinned_pages mode.
- * Do users of this interface need to know how that interacts?
- *
  *----------------------------------------------------------------------
  */
 static inline void
@@ -602,8 +598,11 @@ cache_flush(cache *cc)
  * cache_evict
  *
  * Evicts all the pages.
- * Asserts that there are no pins, read locks, claims or write locks.
+ * Asserts that there are no pins (if ignore_pinned_pages is false), read
+ * locks, claims or write locks.
  * Always returns 0.
+ *
+ * XXX Does ignore_pinned_pages ignore the pages or the pinnedness of the pages?
  *
  * Test facility.
  * This method is only used for testing, specifically in cache_test.
--- a/src/cache.h
+++ b/src/cache.h
@@ -269,10 +269,12 @@ cache_alloc(cache *cc, uint64 addr, page_type type)
  * This call may block on I/O (to complete writebacks initiated before
  * this call).
  *
- * This function is used to maintain an invariant that the cache only
- * contains only contains pages allocated in the RC allocator. Once an
- * extent is freed, this function is used to evict all of its pages from
- * the cache.
+ * Once an extent is freed, this function is used to evict all of its pages
+ * from the cache.  This function is used to maintain an invariant that the
+ * cache only contains only contains pages allocated in the RC allocator; if
+ * this function were not called, and this extent were to be reallocated,
+ * subsequent calls to cache_alloc could create duplicate entries in the cache
+ * for the same backing page.
  *----------------------------------------------------------------------
  */
 static inline void
@@ -354,7 +356,8 @@ cache_get_async(cache *cc, uint64 addr, page_type type, cache_async_ctxt *ctxt)
  *----------------------------------------------------------------------
  * cache_async_done
  *
- * XXX @aconway How is this used?
+ * Perform callbacks on the thread that made the async call after an async
+ * operation completes.
  *----------------------------------------------------------------------
  */
 static inline void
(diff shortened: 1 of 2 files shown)


src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
    return cc->ops->page_unpin(cc, page);
 }
 
+/*
+ *-----------------------------------------------------------------------------
+ * cache_page_sync
+ *
+ * Asynchronously writes the page back to disk. Currently there is no way to
+ * check when the writeback has completed,
+ * XXX thus this call is only useful to prepare a page for eviction?
```
ajhconway: This is used to sync log pages opportunistically. The current API doesn't inform the user when this happens. It's not the ideal API.

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
+/*
+ *-----------------------------------------------------------------------------
+ * cache_extent_sync
+ *
+ * Asynchronously syncs the extent beginning at addr.
+ *
+ * *pages_outstanding is immediately incremented by the number of pages
+ * issued for writeback (the non-clean pages of the extent); as writebacks
+ * complete, *pages_outstanding is decremented. The caller may use the
+ * XXX atomically-read? Because it's an aligned uint64?
```
ajhconway: I think it's assumed that it's an aligned uint64 so on x86 it should be safe to read.

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
+ * value of *pages_outstanding to determine when the extent writeback is
+ * complete.
+ *
+ * All pages in the extent must be clean or cleanable.
+ * XXX What other states might there be, and what would happen if you
```
ajhconway: Pages cannot be write locked for example.

If they were, I think those pages would not be sync'd and I'm not 100% sure what happens.

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
+ * called this with a page in such a state?
+ * XXX should we copy this remark from flush?
+ * XXX Asserts that there are no pins, read locks, claims or write locks.
```
ajhconway: No, I think it's fine if there are pins, read locks or claims, since those do not block writeback.

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
 cache_evict(cache *cc, bool ignore_pinned_pages)
 {
    return cc->ops->evict(cc, ignore_pinned_pages);
 }
 
+/*
+ *-----------------------------------------------------------------------------
+ * cache_cleanup
+ *
+ * XXX ?
```
ajhconway: Bad naming.

This function performs cache callbacks. Generally this isn't required to be called externally, but there are some cases where it must be. For example, prior to shutdown, calling this ensures that all callbacks are called.

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
 {
    return cc->ops->assert_ungot(cc, addr);
 }
 
+/*
+ *-----------------------------------------------------------------------------
+ * cache_assert_free
+ *
+ * Debugging facility.
+ * Asserts that no thread has a write lock on the page at addr.
```
ajhconway, line 653: Needs a rename.

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
+ */
 static inline void
 cache_reset_stats(cache *cc)
 {
    return cc->ops->reset_stats(cc);
 }
 
+/*
+ *-----------------------------------------------------------------------------
+ * cache_reset_stats
```
ajhconway, line 706: cache_io_stats

src/cache.h
```diff
@@ -270,211 +330,541 @@ cache_ctxt_init(cache            *cc,
...
 cache_present(cache *cc, page_handle *page)
 {
    return cc->ops->cache_present(cc, page);
 }
 
+/*
+ *-----------------------------------------------------------------------------
+ * cache_enable_sync_get
+ *
+ * XXX No idea.
```
ajhconway: This is a safety function. There is a known deadlock with async gets, so they are (were?) disabled by default. This function must be called to enable their use or perhaps an assertion is triggered.

ajhconway (commented):
Thank you for doing this!

I tried to fill in the XXXs. I didn't know 100% all the answers.

I corrected a couple other things too.

src/cache.h
jonhnet: Wait, this function enables *sync* gets, not async gets?

jonhnet: Is the known deadlock between a sync get and an async get, or between any two async gets? Or something else?

jonhnet:
@aconway: thanks for all the clarifications. All of these comments make and pass the tests. :v) Please take another peek.

jonhnet:
@ajhconway gentle ping on this collection of comments. Thanks!

ajhconway (approved):
Ship it

PR #503, opened 2022-12-20
use pragma once in preference to matched ifndef blocks

jonhnet:
Switch from matched-ifndef blocks to the "nonstandard but widely supported" #pragma once. This has the nice benefit that there's no possibility of name collision, no need to rename the ifndef if you rename a file later, and just generally less boilerplate clutter to sift through.

### DIFF
--- a/include/splinterdb/data.h
+++ b/include/splinterdb/data.h
@@ -17,8 +17,7 @@
  *
  */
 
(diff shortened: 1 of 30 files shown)


rtjohnso (approved):
LGTM

ajhconway (approved):
Looks great

PR #504, opened 2022-12-21
Jonh move extent funcs to allocator

jonhnet:
This change reduces the width of the polymorphic cache interface by moving things that only depend on the allocator to the allocator.


ajhconway (approved):
LGTM

I'd prefer pulling the TODOs into issues

Issue #505, opened 2022-12-21
Enhance platform_error_log() to also report source code location and pid / tid values as part of the error message.


gapisback:
To make debugging easier, it will be useful to extend the message printed by `platform_error_log()` to also report:

- Source code-location info: file / line / function name
- OS-pid / thread-ID info. (This is helpful to understand thread's identity when it encounters an error.)

These changes are inline what has been done already for the messages printed by `platform_assert()`, initially added under SHA 3674df6 (issue #102) and was subsequently enhanced under issue #462 to report OS pid/tid info.

Prefixing source-code location was extensively used in an on-going dev prototype effort and was found to be useful to quickly spot where the error is coming from.

This issue is being registered to productize that prototype code and integrate it to `/main`.

A sample output with this kind of change will look like so, when `platform_error_log()` is called from a new unit test case:

```
TEST 4/5 misc:test_platform_error_log_w_args
Verbose mode on.  This test exercises error-reporting logic, so on success it will print a message that appears to be an error.
tests/unit/misc_test.c:198:ctest_misc_test_platform_error_log_w_args_run(): OS-pid=3405337, OS-tid=3405337, Thread-ID=0, This is a test message with arguments: sizeof(pid_t)=4, 'String arg to test msg.'
```

This bit `tests/unit/misc_test.c:198:ctest_misc_test_platform_error_log_w_args_run(): OS-pid=3405337, OS-tid=3405337, Thread-ID=0` is extra, that will be prepended by this suggested change to the error message text.

gapisback:
We discussed this proposal with core dev team ... sometime in Jan 2023 (I think; could n't find the slack thread).

There was no agreement in adding this change. 

People were not convinced about the usefulness of this extra info coming out, by default, for every error message.
People thought that it won't be very difficult to map the error text to the line of code where it's raised, so making this extra file / line # available by default (along with OS/tid/pid) may needlessly clutter the error message stream.

The basic concern was that if we do something in this area, we should look into building a proper error reporting machinery, with error-levels and so on. And may be put this extra-info as an optional opt-in, perhaps with a `--verbose-error-info` or `--extended-error-info` (my preference) cmdline argument.

No further work to be done on this issue.

PR #506, opened 2022-12-21
(#505) Enhance platform_error_log() to prepend source code & OS-pid/tid-info

gapisback:
This commit enhances platform_error_log() to prepend source-code location triplet (file, line, function-name) and OS-pid/tid and thread-ID to the message being printed. This extra info makes it easier to quickly find out where-from the code-base a particular error is raised and affecting which OS process / Splinter thread.

---

NOTE: This was a diagnostic enhancement I had used during on-going prototype integration code. I found it very useful to have this extra-info printed for an error.

Rather than applying these few extra fields to every call to `platform_error_log()` that wants to gain this new functionality, I think it will be better to do this generically. That way, in all instances of an error message we will get this extra diagnostic info reliably.

### DIFF
--- a/src/platform_linux/platform.c
+++ b/src/platform_linux/platform.c
@@ -372,7 +372,7 @@ platform_assert_msg(platform_log_handle *log_handle,
                     const char          *message,
                     va_list              varargs)
 {
-   static char assert_msg_fmt[] = "OS-pid=%d, OS-tid=%lu, Thread-ID=%lu, "
+   static char assert_msg_fmt[] = "OS-pid=%d, OS-tid=%d, Thread-ID=%lu, "
                                   "Assertion failed at %s:%d:%s(): \"%s\". ";
    platform_log(log_handle,
                 assert_msg_fmt,
--- a/src/platform_linux/platform_inline.h
+++ b/src/platform_linux/platform_inline.h
@@ -4,6 +4,7 @@
 #ifndef PLATFORM_LINUX_INLINE_H
 #define PLATFORM_LINUX_INLINE_H
 
+#include <unistd.h>
 #include <laio.h>
 #include <string.h> // for memcpy, strerror
 #include <time.h>   // for nanosecond sleep api.
@@ -288,9 +289,20 @@ platform_log_stream_to_string(platform_stream_handle *stream)
       platform_log(Platform_default_log_handle, __VA_ARGS__);                  \
    } while (0)
 
-#define platform_error_log(...)                                                \
+/*
+ * For improved debugging-ability, prepend the user-supplied message string,
+ * 'msgfmt' with source-code location info.
+ */
+#define platform_error_log(msgfmt, ...)                                        \
    do {                                                                        \
-      platform_log(Platform_error_log_handle, __VA_ARGS__);                    \
+      platform_log(Platform_error_log_handle,                                  \
+                   "%s:%d:%s(): OS-pid=%d, OS-tid=%d, Thread-ID=%lu, " msgfmt, \
+                   __FILE__,                                                   \
+                   __LINE__,                                                   \
+                   __FUNCTION__,                                               \
+                   getpid(),                                                   \
+                   gettid(),                                                   \
+                   platform_get_tid() __VA_OPT__(, ) __VA_ARGS__);             \
    } while (0)
 
 #define platform_log_stream(stream, ...)                                       \
--- a/tests/unit/misc_test.c
+++ b/tests/unit/misc_test.c
@@ -67,12 +67,14 @@ CTEST_SETUP(misc)
    // So lets send those messages to /dev/null unless VERBOSE=1.
    if (Ctest_verbose) {
       data->log_output = stdout;
+      platform_set_log_streams(stdout, stderr);
       CTEST_LOG_INFO("\nVerbose mode on.  This test exercises error-reporting "
                      "logic, so on success it will print a message "
                      "that appears to be an error.\n");
    } else {
-      data->log_output = fopen("/dev/null", "w");
+      FILE *dev_null = data->log_output = fopen("/dev/null", "w");
       ASSERT_NOT_NULL(data->log_output);
+      platform_set_log_streams(dev_null, dev_null);
    }
 }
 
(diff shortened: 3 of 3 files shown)


src/platform_linux/platform.c
```diff
@@ -372,7 +372,7 @@ platform_assert_msg(platform_log_handle *log_handle,
                     const char          *message,
                     va_list              varargs)
 {
-   static char assert_msg_fmt[] = "OS-pid=%d, OS-tid=%lu, Thread-ID=%lu, "
+   static char assert_msg_fmt[] = "OS-pid=%d, OS-tid=%d, Thread-ID=%lu, "
```
gapisback, line 375: Compilation on L299 (in the definition of `platform_error_log()`) fails if this is specified as `%lu`. I found that `sizeof(pid_t)` is `int`, so `%d` is the right specifier.

This old code succeeded compilation as this print-format specifier is hidden behind a collection of nested macros so isn't visible actually till run-time.

src/platform_linux/platform_inline.h
```diff
@@ -288,9 +289,20 @@ platform_log_stream_to_string(platform_stream_handle *stream)
...
-#define platform_error_log(...)                                                \
+/*
+ * For improved debugging-ability, prepend the user-supplied message string,
+ * 'msgfmt' with source-code location info.
+ */
+#define platform_error_log(msgfmt, ...)                                        \
    do {                                                                        \
-      platform_log(Platform_error_log_handle, __VA_ARGS__);                    \
+      platform_log(Platform_error_log_handle,                                  \
+                   "%s:%d:%s(): OS-pid=%d, OS-tid=%d, Thread-ID=%lu, " msgfmt, \
```
gapisback, line 299: Extra fields prepended to the error message ... as described in the issue and the commit message.

tests/unit/misc_test.c
```diff
@@ -183,6 +185,22 @@ CTEST2(misc, test_ctest_assert_prints_user_msg_with_params)
...
+ * Test case to exercise platform_error_log() w/ and w/o arguments to msg.
+ */
+CTEST2(misc, test_platform_error_log_wo_args)
+{
+   platform_error_log("This is a test message without arguments\n");
+}
+
+CTEST2(misc, test_platform_error_log_w_args)
+{
+   platform_error_log("This is a test message with arguments: "
```
gapisback, line 198: Couple of simple unit tests to show that this new printing stuff will work correctly.

Here is the output using `VERBOSE=1` when running these new test cases:

```
sdb-fdb-build:[41] $ VERBOSE=1 build/release/bin/unit/misc_test test_platform_error_log
Running 1 CTests, suite name 'misc', test case 'test_platform_error_log'.
TEST 1/2 misc:test_platform_error_log_wo_args
Verbose mode on.  This test exercises error-reporting logic, so on success it will print a message that appears to be an error.
tests/unit/misc_test.c:193:ctest_misc_test_platform_error_log_wo_args_run(): OS-pid=3406407, OS-tid=3406407, Thread-ID=0, This is a test message without arguments
[OK]
TEST 2/2 misc:test_platform_error_log_w_args
Verbose mode on.  This test exercises error-reporting logic, so on success it will print a message that appears to be an error.
tests/unit/misc_test.c:198:ctest_misc_test_platform_error_log_w_args_run(): OS-pid=3406407, OS-tid=3406407, Thread-ID=0, This is a test message with arguments: sizeof(pid_t)=4, 'String arg to test msg.'
[OK]
RESULTS: 2 tests (2 ok, 0 failed, 0 skipped) ran in 0 ms
```

tests/unit/misc_test.c
```diff
@@ -67,12 +67,14 @@ CTEST_SETUP(misc)
...
       data->log_output = stdout;
+      platform_set_log_streams(stdout, stderr);
       CTEST_LOG_INFO("\nVerbose mode on.  This test exercises error-reporting "
                      "logic, so on success it will print a message "
                      "that appears to be an error.\n");
    } else {
-      data->log_output = fopen("/dev/null", "w");
+      FILE *dev_null = data->log_output = fopen("/dev/null", "w");
       ASSERT_NOT_NULL(data->log_output);
+      platform_set_log_streams(dev_null, dev_null);
```
gapisback, line 77: Slightly unrelated change, but seems like the right thing to do here.

ajhconway (changes requested):
I understand that you could really use this functionality for your integration work, but I think that it doesn't make sense in general. What I think should happen for this PR is that you should implement this as an off-by-default flag, so that you can use it, but we don't have to sort out our logging infrastructure, which is a low priority right now.

The problem is that our error logging covers too cases. The first case covers bugs, unexpected values, and misconfigurations. These should really be assertions essentially and I think this is what you want to cover mostly.

The second are for additionally information for system errors, which we don't handle gracefully anyway, but for which printing call-site information isn't very useful.

The right solution is to have two different logging levels for these or maybe just to go through and turn lots of `platform_error_log` calls into assertions. I don't think this is high enough priority to do that now though.

gapisback:
Hi, @ajhconway : Let me think about your suggestion [here](https://github.com/vmware/splinterdb/pull/506#pullrequestreview-1238125051).

I would have liked this to be the default option. Opting-in (for now) is still an option, leaving open the possibility to make this output the default, once we get more experience with the usefulness of this kind of behaviour in future.

Let me also talk to @rosenhouse when he returns next week. We had doodled around some ideas of, perhaps, making the newly added `VERBOSE=1` env-var, that is used to spit out messages from unit-tests, a `--cmd-line` flag.

I am wondering if it's a good time to reconcile both efforts and may be add cmd-line flags like `--verbose-output` (for the env-var) and (something like) `--extended-error-info` for this new behaviour.

(The term Extended Error Info generally refers to this kind of extra info printed with error messages, listing stuff like `hostname`, `ip-addresses` and so on. Here's [an example](https://admx.help/?Category=Windows_10_2016&Policy=Microsoft.Policies.RemoteProcedureCalls::RpcExtendedErrorInformation).)

@rtjohnso may have some thoughts / opinions on the suggested name of `--extended-error-info`.

If this feature has to be made opt-in, I would like to find a way to make it so that it can only be enabled for a specific thread, so that one can toggle on / off this extra behaviour only for certain processes / threads that may be running into a specific error situation that one is debugging. Let me think about that ...

gapisback:
@rtjohnso : Consider cmd line option of `--error-level` to future-proof this.

gapisback:
Based on review feedback (see note from Alex) and internal discussion w/ core team during PR reviews, the recommendation is to *not* pursue with this change. 

See the notes in issue #505, explaining why the decision was made to not pursue this change-set.

Issue #507, opened 2022-12-21
Rework platform_buffer_create() to take platform_heap_id hid as arg. Fix (minor) latent bug in this routine.


gapisback:
`platform_buffer_create()` currently takes `platform_heap_handle UNUSED_PARAM(heap_handle)` as the arg to specify the "heap"-handle from which a new `mmap()`-ed buffer will be created.

Currently, in Splinter, `heap_id` is NULL, and we fall-back to using `malloc()`. The heap is used to allocte `buffer_handle *bh` structure.

Right now, things work ok. In a future dev effort, we are looking to support alternate memory handles, say, a shared-memory segment from which stuff like `bh` can be allocated. The `platform_buffer_create()` uses `platform_get_heap_id()` to access the heap-ID to allocate from (and that's anyway NULL right now).

This issue is being raised to make couple of small changes:

1. Change the interface of `platform_buffer_create()` to take `heap_id` as its argument. In future, this will make it easier to drop-in support for other sources of heap-memory given by the heap-ID.
2. There is a minor bug in the current code. This function could access a NULL `bh` if ever `TYPED_MALLOC()` were to fail. This code chunk will seg-fault:

```
 60 platform_buffer_create(size_t               length,
 61                        platform_heap_handle UNUSED_PARAM(heap_handle),
 62                        platform_module_id   UNUSED_PARAM(module_id))
 63 {
 64    buffer_handle *bh = TYPED_MALLOC(platform_get_heap_id(), bh);
 65
 66    if (bh != NULL) {
[...]
 91    bh->length = length;
 92    return bh;
```

PR #508, opened 2022-12-22
(#507) Change platform_buffer_create()/destroy() interfaces to platform_buffer_init() / deinit() methods.

gapisback:
This commit reworks the interfaces of `platform_buffer_create()` and `platform_buffer_destroy()` to now become `platform_buffer_init()` and `platform_buffer_deinit()` interfaces.

This removes a dependency on `platform_heap_id` or `platform_heap_handle` arg. 

This change does not functionally change anything in these methods. Also, this (indirectly) fixes a minor bug in the create function to deal with the (very rare) case of failing to allocate memory from the heap for `buffer_handle *bh` . Added small unit-test, platform_apis_test, to exercise these changes.

---
NOTE: This is some common infra-code built in a prototype dev branch for supporting different memory handles.

I'm pulling this piece out on its own for integration to /main, so that this ground work can be laid for integrating future dev-efforts (w/ little less pain).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -443,6 +443,9 @@ $(BINDIR)/$(UNITDIR)/task_system_test: $(UTIL_SYS)                              
                                        $(OBJDIR)/$(FUNCTIONAL_TESTSDIR)/test_async.o \
                                        $(LIBDIR)/libsplinterdb.so
 
+$(BINDIR)/$(UNITDIR)/platform_apis_test: $(UTIL_SYS)       \
+                                         $(PLATFORM_SYS)
+
 ########################################
 # Convenience mini unit-test targets
 unit/util_test:                    $(BINDIR)/$(UNITDIR)/util_test
--- a/src/clockcache.c
+++ b/src/clockcache.c
@@ -46,7 +46,7 @@
  *
  *      clockcache_log, etc. are used to write an output of cache operations to
  *      a log file for debugging purposes. If CC_LOG is set, then all output is
- *      written, if ADDR_TRACING is set, then only operations which affect
+ *      written. If ADDR_TRACING is set, then only operations which affect
  *      entries with either entry_number TRACE_ENTRY or address TRACE_ADDR are
  *      written.
  *
@@ -1761,14 +1761,13 @@ clockcache_config_init(clockcache_config *cache_cfg,
 }
 
 platform_status
-clockcache_init(clockcache          *cc,   // OUT
-                clockcache_config   *cfg,  // IN
-                io_handle           *io,   // IN
-                allocator           *al,   // IN
-                char                *name, // IN
-                platform_heap_handle hh,   // IN
-                platform_heap_id     hid,  // IN
-                platform_module_id   mid)    // IN
+clockcache_init(clockcache        *cc,   // OUT
+                clockcache_config *cfg,  // IN
+                io_handle         *io,   // IN
+                allocator         *al,   // IN
+                char              *name, // IN
+                platform_heap_id   hid,  // IN
+                platform_module_id mid)  // IN
 {
    int      i;
    threadid thr_i;
@@ -1803,10 +1802,9 @@ clockcache_init(clockcache          *cc,   // OUT
    clockcache_log(
       0, 0, "init: capacity %lu name %s\n", cc->cfg->capacity, name);
 
-   cc->al          = al;
-   cc->io          = io;
-   cc->heap_handle = hh;
-   cc->heap_id     = hid;
+   cc->al      = al;
+   cc->io      = io;
+   cc->heap_id = hid;
 
    /* lookup maps addrs to entries, entry contains the entries themselves */
    cc->lookup =
@@ -1824,12 +1822,14 @@ clockcache_init(clockcache          *cc,   // OUT
       goto alloc_error;
    }
 
+   platform_status rc = STATUS_NO_MEMORY;
+
    /* data must be aligned because of O_DIRECT */
-   cc->bh = platform_buffer_create(cc->cfg->capacity, cc->heap_handle, mid);
-   if (!cc->bh) {
+   rc = platform_buffer_init(&cc->bh, cc->cfg->capacity);
+   if (!SUCCESS(rc)) {
       goto alloc_error;
    }
-   cc->data = platform_buffer_getaddr(cc->bh);
+   cc->data = platform_buffer_getaddr(&cc->bh);
 
    /* Set up the entries */
    for (i = 0; i < cc->cfg->page_capacity; i++) {
@@ -1841,14 +1841,19 @@ clockcache_init(clockcache          *cc,   // OUT
 
    /* Entry per-thread ref counts */
    size_t refcount_size = cc->cfg->page_capacity * CC_RC_WIDTH * sizeof(uint8);
-   cc->rc_bh = platform_buffer_create(refcount_size, cc->heap_handle, mid);
-   if (!cc->rc_bh) {
+
+   rc = platform_buffer_init(&cc->rc_bh, refcount_size);
+   if (!SUCCESS(rc)) {
       goto alloc_error;
    }
-   cc->refcount = platform_buffer_getaddr(cc->rc_bh);
+   cc->refcount = platform_buffer_getaddr(&cc->rc_bh);
+
    /* Separate ref counts for pins */
    cc->pincount =
       TYPED_ARRAY_ZALLOC(cc->heap_id, cc->pincount, cc->cfg->page_capacity);
+   if (!cc->pincount) {
+      goto alloc_error;
+   }
 
    /* The hands and associated page */
    cc->free_hand  = 0;
@@ -1872,15 +1877,17 @@ alloc_error:
    return STATUS_NO_MEMORY;
 }
 
+/*
+ * De-init the resources allocated to initialize a clockcache.
+ * This function may be called to deal with error situations, or a failed
+ * clockcache_init(). So check for non-NULL handles before trying to release
+ * resources.
+ */
 void
 clockcache_deinit(clockcache *cc) // IN/OUT
 {
    platform_assert(cc != NULL);
 
-   /*
-    * Check for non-null cause this is also used to clean up a failed
-    * clockcache_init
-    */
    if (cc->logfile) {
       clockcache_log(0, 0, "deinit %s\n", "");
 #if defined(CC_LOG) || defined(ADDR_TRACING)
@@ -1888,19 +1895,34 @@ clockcache_deinit(clockcache *cc) // IN/OUT
 #endif
    }
 
-   if (cc->rc_bh) {
-      platform_buffer_destroy(cc->rc_bh);
+   if (cc->lookup) {
+      platform_free(cc->heap_id, cc->lookup);
+   }
+   if (cc->entry) {
+      platform_free(cc->heap_id, cc->entry);
    }
 
-   platform_free(cc->heap_id, cc->entry);
-   platform_free(cc->heap_id, cc->lookup);
-   if (cc->bh) {
-      platform_buffer_destroy(cc->bh);
+   debug_only platform_status rc = STATUS_TEST_FAILED;
+   if (cc->data) {
+      rc = platform_buffer_deinit(&cc->bh);
+
+      // We expect above to succeed. Anyway, we are in the process of
+      // dismantling the clockcache, hence, for now, can't do much by way
+      // of reporting errors further upstream.
+      debug_assert(SUCCESS(rc), "rc=%s", platform_status_to_string(rc));
+      cc->data = NULL;
    }
-   cc->data = NULL;
-   platform_free_volatile(cc->heap_id, cc->batch_busy);
+   if (cc->refcount) {
+      rc = platform_buffer_deinit(&cc->rc_bh);
+      debug_assert(SUCCESS(rc), "rc=%s", platform_status_to_string(rc));
+      cc->refcount = NULL;
+   }
+
    if (cc->pincount) {
       platform_free_volatile(cc->heap_id, cc->pincount);
+   }
+   if (cc->batch_busy) {
+      platform_free_volatile(cc->heap_id, cc->batch_busy);
    }
 }
 
--- a/src/clockcache.h
+++ b/src/clockcache.h
@@ -116,14 +116,13 @@ struct clockcache {
 
    uint32              *lookup;
    clockcache_entry    *entry;
-   buffer_handle       *bh;   // actual memory for pages
+   buffer_handle        bh;   // actual memory for pages
    char                *data; // convenience pointer for bh
    platform_log_handle *logfile;
-   platform_heap_handle heap_handle;
    platform_heap_id     heap_id;
 
    // Distributed locks (the write bit is in the status uint32 of the entry)
-   buffer_handle  *rc_bh;
+   buffer_handle   rc_bh;
    volatile uint8 *refcount;
    volatile uint8 *pincount;
 
@@ -157,14 +156,13 @@ clockcache_config_init(clockcache_config *cache_config,
                        uint64             use_stats);
 
 platform_status
-clockcache_init(clockcache          *cc,   // OUT
-                clockcache_config   *cfg,  // IN
-                io_handle           *io,   // IN
-                allocator           *al,   // IN
-                char                *name, // IN
-                platform_heap_handle hh,   // IN
-                platform_heap_id     hid,  // IN
-                platform_module_id   mid);   // IN
+clockcache_init(clockcache        *cc,   // OUT
+                clockcache_config *cfg,  // IN
+                io_handle         *io,   // IN
+                allocator         *al,   // IN
+                char              *name, // IN
+                platform_heap_id   hid,  // IN
+                platform_module_id mid); // IN
 
 void
 clockcache_deinit(clockcache *cc); // IN
--- a/src/platform_linux/platform.c
+++ b/src/platform_linux/platform.c
@@ -56,45 +56,49 @@ void
 platform_heap_destroy(platform_heap_handle UNUSED_PARAM(*heap_handle))
 {}
 
-buffer_handle *
-platform_buffer_create(size_t               length,
-                       platform_heap_handle UNUSED_PARAM(heap_handle),
-                       platform_module_id   UNUSED_PARAM(module_id))
+/*
+ * platform_buffer_init() - Initialize an input buffer_handle, bh.
+ *
+ * Certain modules, e.g. the buffer cache, need a very large buffer which
+ * may not be serviceable by the heap. Create the requested buffer using
+ * mmap() and initialize the input 'bh' to track this memory allocation.
+ */
+platform_status
+platform_buffer_init(buffer_handle *bh, size_t length)
 {
-   buffer_handle *bh = TYPED_MALLOC(platform_get_heap_id(), bh);
+   platform_status rc = STATUS_NO_MEMORY;
 
-   if (bh != NULL) {
-      int prot  = PROT_READ | PROT_WRITE;
-      int flags = MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE;
-      if (platform_use_hugetlb) {
-         flags |= MAP_HUGETLB;
-      }
-
-      bh->addr = mmap(NULL, length, prot, flags, -1, 0);
-      if (bh->addr == MAP_FAILED) {
-         platform_error_log(
-            "mmap (%lu) failed with error: %s\n", length, strerror(errno));
-         goto error;
-      }
-
-      if (platform_use_mlock) {
-         int rc = mlock(bh->addr, length);
-         if (rc != 0) {
-            platform_error_log(
-               "mlock (%lu) failed with error: %s\n", length, strerror(errno));
-            munmap(bh->addr, length);
-            goto error;
-         }
-      }
+   int prot  = PROT_READ | PROT_WRITE;
+   int flags = MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE;
+   if (platform_use_hugetlb) {
+      flags |= MAP_HUGETLB;
    }
 
+   bh->addr = mmap(NULL, length, prot, flags, -1, 0);
+   if (bh->addr == MAP_FAILED) {
+      platform_error_log(
+         "mmap (%lu bytes) failed with error: %s\n", length, strerror(errno));
+      goto error;
+   }
+
+   if (platform_use_mlock) {
+      int mlock_rv = mlock(bh->addr, length);
+      if (mlock_rv != 0) {
+         platform_error_log("mlock (%lu bytes) failed with error: %s\n",
+                            length,
+                            strerror(errno));
+         munmap(bh->addr, length);
+         rc = CONST_STATUS(errno);
+         goto error;
+      }
+   }
    bh->length = length;
-   return bh;
+   return STATUS_OK;
 
 error:
-   platform_free(platform_get_heap_id(), bh);
-   bh = NULL;
-   return bh;
+   // Reset, in case mmap() or mlock() failed.
+   bh->addr = NULL;
+   return rc;
 }
 
 void *
@@ -103,8 +107,14 @@ platform_buffer_getaddr(const buffer_handle *bh)
    return bh->addr;
 }
 
+/*
+ * platform_buffer_deinit() - Deinit the buffer handle, which involves
+ * unmapping the memory for the large buffer created using mmap().
+ * This is expected to be called on a 'bh' that has been successfully
+ * initialized, thru a prior platform_buffer_init() call.
+ */
 platform_status
-platform_buffer_destroy(buffer_handle *bh)
+platform_buffer_deinit(buffer_handle *bh)
 {
    int ret;
    ret = munmap(bh->addr, bh->length);
@@ -112,8 +122,8 @@ platform_buffer_destroy(buffer_handle *bh)
       return CONST_STATUS(errno);
    }
 
-   platform_free(platform_get_heap_id(), bh);
-
+   bh->addr   = NULL;
+   bh->length = 0;
    return STATUS_OK;
 }
 
--- a/src/platform_linux/platform.h
+++ b/src/platform_linux/platform.h
@@ -626,16 +626,14 @@ platform_heap_create(platform_module_id    module_id,
 void
 platform_heap_destroy(platform_heap_handle *heap_handle);
 
-buffer_handle *
-platform_buffer_create(size_t               length,
-                       platform_heap_handle heap_handle,
-                       platform_module_id   module_id);
+platform_status
+platform_buffer_init(buffer_handle *bh, size_t length);
 
 void *
 platform_buffer_getaddr(const buffer_handle *bh);
 
 platform_status
-platform_buffer_destroy(buffer_handle *bh);
+platform_buffer_deinit(buffer_handle *bh);
 
 platform_status
 platform_mutex_init(platform_mutex    *mu,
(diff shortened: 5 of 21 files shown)


Makefile
```diff
@@ -428,6 +428,9 @@ $(BINDIR)/$(UNITDIR)/task_system_test: $(UTIL_SYS)
                                        $(OBJDIR)/$(FUNCTIONAL_TESTSDIR)/test_async.o \
                                        $(LIBDIR)/libsplinterdb.so
 
+$(BINDIR)/$(UNITDIR)/platform_apis_test: $(UTIL_SYS)       \
+                                         $(PLATFORM_SYS)
+
```
gapisback, line 433: Minor new unit-test to exercise the changed APIs.

src/platform_linux/platform.c
```diff
@@ -56,38 +56,46 @@ void
...
+ * Certain modules, e.g. the buffer cache, need a very large buffer which
+ * may not be serviceable by the heap. Create the requested buffer using
+ * mmap() and return a handle to it.
+ */
 buffer_handle *
-platform_buffer_create(size_t               length,
-                       platform_heap_handle UNUSED_PARAM(heap_handle),
-                       platform_module_id   UNUSED_PARAM(module_id))
+platform_buffer_create(size_t             length,
+                       platform_heap_id   heap_id,
```
gapisback, line 68: This arg sets the stage to allow diff heap-ID handles in future.

src/platform_linux/platform.c
```diff
@@ -56,38 +56,46 @@ void
...
+                       platform_module_id UNUSED_PARAM(module_id))
 {
-   buffer_handle *bh = TYPED_MALLOC(platform_get_heap_id(), bh);
+   buffer_handle *bh = TYPED_MALLOC(heap_id, bh);
+   if (bh == NULL) {
+      return bh;
```
gapisback, line 73: Minor bug-fix; otherwise old L99 will seg-fault.

src/platform_linux/platform.c
```diff
@@ -103,16 +111,21 @@ platform_buffer_getaddr(const buffer_handle *bh)
...
-   int ret;
+   buffer_handle *bh = *bh_in;
+   int            ret;
    ret = munmap(bh->addr, bh->length);
    if (ret) {
       return CONST_STATUS(errno);
    }
 
-   platform_free(platform_get_heap_id(), bh);
+   platform_free(heap_id, *bh_in);
```
gapisback, line 128: Interface changed (for the same reasons as for the `create()` method) so that we can use the right `heap_id` while freeing memory. Otherwise, default `platform_get_heap_id()` may be NULL, and we may end up not freeing the memory with the proper heap-ID.

src/platform_linux/platform.c
```diff
@@ -56,38 +56,46 @@ void
...
+      return bh;
+   }
 
-   if (bh != NULL) {
-      int prot  = PROT_READ | PROT_WRITE;
-      int flags = MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE;
-      if (platform_use_hugetlb) {
-         flags |= MAP_HUGETLB;
-      }
+   int prot  = PROT_READ | PROT_WRITE;
```
gapisback, line 76: Blocks of code are indented to the right now as we've lost L 66 `if()` condition. Function will return early on L73 if `bh == NULL`.

src/platform_linux/platform.c
```diff
@@ -103,16 +111,21 @@ platform_buffer_getaddr(const buffer_handle *bh)
    return bh->addr;
 }
 
+/*
+ * platform_buffer_destroy() - Destroy the large buffer created using mmap()
+ * Free all associated memory. buffer_handle *bh_in is NULL'ed out on exit.
+ */
 platform_status
-platform_buffer_destroy(buffer_handle *bh)
+platform_buffer_destroy(platform_heap_id heap_id, buffer_handle **bh_in)
```
gapisback, line 119: Changed the interface to receive address of `buffer_handle *`, so that it can be NULL'ed out on exit.
(This is a safer way to avoid callers accidentally accessing `bh_in` after this call.

src/rc_allocator.c
```diff
@@ -388,7 +388,7 @@ rc_allocator_init(rc_allocator        *al,
 void
 rc_allocator_deinit(rc_allocator *al)
 {
-   platform_buffer_destroy(al->bh);
+   platform_buffer_destroy(al->heap_id, &al->bh);
```
gapisback, line 391: example of this usage. `al->bh` will be NULL'ed out upon return.

src/platform_linux/platform.h
```diff
@@ -627,15 +627,15 @@ void
...
+platform_buffer_create(size_t             length,
+                       platform_heap_id   heap_id,
+                       platform_module_id module_id);
 
 void *
 platform_buffer_getaddr(const buffer_handle *bh);
 
 platform_status
-platform_buffer_destroy(buffer_handle *bh);
+platform_buffer_destroy(platform_heap_id heap_id, buffer_handle **bh_in);
```
rtjohnso, line 638: Can you just add a field to the buffer handle so it remembers its own heapid so you don't need to pass the heapid to the destroy function?

rtjohnso (changes requested):
I would like to propose a different approach:

Change these from `create` and `destroy` functions to `init` and `deinit` functions.

Then change all users to declare `buffer_handle`s instead of `buffer_handle *` fields in their structures.

I believe this will solve the problem you are trying to solve.

I discussed this with Alex and he also thinks this would actually be better than the existing code.

src/platform_linux/platform.h
rtjohnso: This comment is obviated by my high level proposal.  I think the high-level proposal will eliminate the need for a heap parameter altogether.

gapisback: Like you observed, elsewhere, this new parameter `heap_id` became unnecessary for these interfaces. I didn't have to add it to the `buffer_handle` struct either, which keeps this change simpler.

gapisback:
HI @rtjohnso - Based on your [suggestion to rework](https://github.com/vmware/splinterdb/pull/508#pullrequestreview-1239454089) this approach, I made the required changes. Uploaded a [new commit, here.](https://github.com/vmware/splinterdb/pull/508/commits/b7a2b22fd926748fe3aba4f92512a68c8e246aa6) 

(I will squash the two commits and will clean up the commit messages once you approve this work, and before I integrate this to `/main`.)

It has certainly simplified the interface and code and solves the dependency on the `heap_id` argument entirely. This works well to pre-stage this change to absorb use of different memory streams in the future. So, thanks!

src/platform_linux/platform.c
```diff
@@ -56,45 +56,46 @@ void
...
-      if (platform_use_mlock) {
-         int rc = mlock(bh->addr, length);
-         if (rc != 0) {
-            platform_error_log(
-               "mlock (%lu) failed with error: %s\n", length, strerror(errno));
-            munmap(bh->addr, length);
-            goto error;
-         }
+   if (platform_use_mlock) {
+      int rc = mlock(bh->addr, length);
```
rtjohnso, line 85: Not you fault but...

I just noticed that this `rc` shadows the `platform_status rc` declared on line 69.

Also, when the `mlock` fails, we return `STATUS_NO_MEMORY`.

Could you change the name of this `rc` so it doesn't shadow the previous, and set a better error code for our return in this case?  Maybe `STATUS_NO_PERMISSION`?

rtjohnso: Oh or maybe `CONST_STATUS(errno)`?

src/clockcache.c
```diff
@@ -1841,11 +1843,12 @@ clockcache_init(clockcache          *cc,   // OUT
 
    /* Entry per-thread ref counts */
    size_t refcount_size = cc->cfg->page_capacity * CC_RC_WIDTH * sizeof(uint8);
-   cc->rc_bh = platform_buffer_create(refcount_size, cc->heap_handle, mid);
-   if (!cc->rc_bh) {
+
+   rc = platform_buffer_init(&cc->rc_bh, refcount_size);
+   if (!SUCCESS(rc)) {
```
rtjohnso, line 1848: The error handling needs to be tightened up.

If either `platform_buffer_init` fails, then we `goto alloc_error`, which does `clockcache_deinit`.  And `clockcache_deinit` unconditionally calls `platform_buffer_deinit` on both buffers.  And `platform_buffer_deinit` unconditionally calls `munmap` on the `buffer_handle`'s `addr`.  But, depending on where we fail, none, one, or both of the buffers may have not been inited, so calling `munmap` could cause an error.

There are two ways to fix this, I think:

1. Modify `clockcache_deinit` to check whether the corresponding pointer is `NULL` before it calls `platform_buffer_deinit`.  So, it would check `cc->data` before calling `platform_buffer_deinit(&cc->bh)` and it would check `cc->refcount` before calling `platform_buffer_deinit(&cc->rc_bh)`.
2. Modify `clockcache_init` to have different error handling targets and do the cleanup on its own, i.e. have `goto alloc_failure1;`, `goto alloc_failure2;`, etc.

My preference is Option 1, but either is fine.

src/clockcache.c
```diff
@@ -1888,15 +1891,12 @@ clockcache_deinit(clockcache *cc) // IN/OUT
 #endif
    }
 
-   if (cc->rc_bh) {
-      platform_buffer_destroy(cc->rc_bh);
-   }
+   platform_buffer_deinit(&cc->rc_bh);
 
    platform_free(cc->heap_id, cc->entry);
```
rtjohnso, line 1896: Along similar lines, I think all of these need to be wrapped in NULL tests.

src/rc_allocator.c
```diff
@@ -441,9 +440,14 @@ rc_allocator_mount(rc_allocator        *al,
 
    uint32 buffer_size = cfg->extent_capacity * sizeof(uint8);
    buffer_size        = ROUNDUP(buffer_size, cfg->io_cfg->page_size);
-   al->bh = platform_buffer_create(buffer_size, al->heap_handle, mid);
-   platform_assert(al->bh != NULL);
-   al->ref_count = platform_buffer_getaddr(al->bh);
+   status             = platform_buffer_init(&al->bh, buffer_size);
+   if (!SUCCESS(status)) {
```
rtjohnso, line 444: Nice!  Love getting rid of assertions.

src/rc_allocator.c
```diff
@@ -441,9 +440,14 @@ rc_allocator_mount(rc_allocator        *al,
...
+      platform_mutex_destroy(&al->lock);
+      platform_free(al->heap_id, al->meta_page);
```
rtjohnso, line 446: Although technically, if you want to undo allocations in the reverse order in which they were done, I believe the `platform_mutex_destroy` should come after the `platform_free`.

tests/functional/test_functionality.c
```diff
@@ -18,9 +18,10 @@
 #include "poison.h"
 
 void
-destroy_test_splinter_shadow_array(test_splinter_shadow_array *sharr)
+destroy_test_splinter_shadow_array(platform_heap_id            hid,
```
rtjohnso, line 21: I think you can get rid of the `hid` argument.

tests/functional/test_splinter_shadow.c
```diff
@@ -100,11 +100,12 @@ test_splinter_shadow_create(test_splinter_shadow_tree **tree,
                             platform_heap_id            hid,
```
rtjohnso, line 100: I wish `test_splinter_shadow_create` were also an `...init`.

But that is definitely outside the scope of this PR, so not your problem!

tests/functional/test_splinter_shadow.c
```diff
@@ -323,25 +319,25 @@ test_splinter_shadow_iterate_tree(AvlTreeLinks               *root,
  *       Will allocate memory for the key and value arrays.
  *-----------------------------------------------------------------------------
  */
-
 platform_status
 test_splinter_build_shadow_array(test_splinter_shadow_tree  *tree,
                                  test_splinter_shadow_array *shadow_array,
-                                 platform_heap_handle        hh)
+                                 platform_heap_handle        hid)
```
rtjohnso, line 325: Looks like `hid` is unused now.

tests/functional/test_splinter_shadow.c
rtjohnso: Also, `hh` is now unused, so delete it.

src/clockcache.h
```diff
@@ -116,14 +116,14 @@ struct clockcache {
 
    uint32              *lookup;
    clockcache_entry    *entry;
-   buffer_handle       *bh;   // actual memory for pages
+   buffer_handle        bh;   // actual memory for pages
    char                *data; // convenience pointer for bh
    platform_log_handle *logfile;
    platform_heap_handle heap_handle;
```
rtjohnso, line 122: The `heap_handle` field appears to be dead after these changes.

src/rc_allocator.h
```diff
@@ -58,7 +58,7 @@ typedef struct rc_allocator_stats {
 typedef struct rc_allocator {
    allocator               super;
    allocator_config       *cfg;
-   buffer_handle          *bh;
+   buffer_handle           bh;
```
rtjohnso, line 61: The `heap_handle` field of this structure is now dead, I think.

rtjohnso (changes requested):
Awesome.

There are two main things I'm requesting:

- Clean up the error handling in `clockcache_init` (see comments).
- Audit the code you've changed to see if any additional `heap_id` or `heap_handle` parameters or structure fields can be eliminated.  I think I flagged them all in comments, but it would be good to double-check.

Thank you!

src/clockcache.c
gapisback: @rtjohnso - There are quite a few callers to `platform_buffer_deinit()`, and technically, one has to comb through all instances to make sure that we are not calling this wrongly.

A 3rd approach would be to push the checks to inside of `platform_buffer_deinit()`.

It will try to do the `munmap()` only if `bh->addr` is non-NULL. This will solve the error handling issue from `clockcache_deinit()` and also protect all other instances where `platform_buffer_deinit()` is being called.

What are your thoughts on this approach (before I implement it).

src/clockcache.c
gapisback: That would not be needed if we pushed the NULL-check to inside `platform_buffer_deinit()`. No ?

rtjohnso:
Before the buffer_handle is inited, addr may be garbage, so you cannot
safely deinit based on the value of addr.

Best,
Rob

src/clockcache.c
gapisback: (And, ooops, ... sorry, I introduced this error handling issue when implementing the change to introduce `init() / deinit()` interface, without looking too hard at error handling. Sorry!)

src/clockcache.h
gapisback: Yes; you are right. Took it out.

src/platform_linux/platform.c
gapisback: Renamed `rc` on L85 to `mlock_rv` and on error, will set `rc` to `CONST_STATUS(errno)`.

src/rc_allocator.c
gapisback: Yes, right.  Changed it accordingly.

src/clockcache.c
gapisback: @rtjohnso : Re: your comment that "_Before the buffer_handle is inited, addr may be garbage, so you cannot safely deinit based on the value of addr._"

In the two places we are talking about, both `rc_allocator_init()` and `clockcache_init()` do a `ZERO_CONTENTS(al)` and `ZERO_CONTENTS(cc)`.

So, certainly for the error path going through `clockcache_deinit(), we should be sure that `cc->bh->addr` will be NULL even before `init()` happened.

Unless you can spot something else I am overlooking, seems like pushing the NULL-check to inside `platform_buffer_deinit()` should work.

rtjohnso:
That's pretty un-idiomatic of how `_init` functions usually work and makes
some weird usage rules.

What if I want to use a buffer_handle somewhere else?  What if I want one
on the stack?  Now I have to know that I need to zero it before I can init
it?  Very error prone.

gapisback:
@rtjohnso - I implemented much of the cleanup to rip out `heap_handle *` args from many places, but chasing other not-needed `hh` and `hid` will take me some time ( and a big screen ) ... Will follow-up tomorrow.

gapisback:
This part, Rob - "_What if I want one on the stack?  Now I have to know that I need to zero it before I can initit?  Very error prone._" .... I'm not so sure, it's that error prone.

I have been trained and have gotten used to a conventional way of coding where every on-stack structure is always `MEMZERO()`-ed; i.e. in our code base `ZERO_CONTENTS()`. For precisely the reason that you want to make sure that all fields are zero'ed out before anything -- including init -- is done with the structure.

I think requiring that on-stack variables, structures are always zero'ed out before use is a stronger discipline to have / enforce.

I can implement it in the ways you are suggesting, but I think then there is more work to chase down all other places where `platform_buffer_init()` is called and ensure that error handling is done right.

We should probably try to talk this through to resolve the right approach to adopt.

src/clockcache.c
```diff
@@ -1873,33 +1877,41 @@ clockcache_init(clockcache        *cc,   // OUT
...
-    */
    if (cc->logfile) {
       clockcache_log(0, 0, "deinit %s\n", "");
 #if defined(CC_LOG) || defined(ADDR_TRACING)
       platform_close_log_file(cc->logfile);
 #endif
    }
 
-   platform_buffer_deinit(&cc->rc_bh);
+   if (cc->lookup) {
```
gapisback, line 1898: I did not bother to arrange these backout codes in `reverse` order of how they are done in `init()` method as each backout fragment is protected under an `if()` check. 

So, although, not in "pure backout form", this ordering should work.

src/clockcache.c
```diff
@@ -1873,33 +1877,41 @@ clockcache_init(clockcache        *cc,   // OUT
...
+      platform_free(cc->heap_id, cc->lookup);
+   }
+   if (cc->entry) {
+      platform_free(cc->heap_id, cc->entry);
+   }
 
-   platform_free(cc->heap_id, cc->entry);
-   platform_free(cc->heap_id, cc->lookup);
    platform_buffer_deinit(&cc->bh);
+   platform_buffer_deinit(&cc->rc_bh);
```
gapisback, line 1906: I did not bracket this with a corresponding `if (!NULL)` check on the member field as I am relying on the the `bh` struct being zero'ed out on declaration. 

Current code always conforms to this ... Let's talk this point through ...

gapisback:
Hi, @rtjohnso & @ajhconway -- I am in the middle of amending the code to correctly handle backout error issues, and to purge  `heap_handle *` args, and fn-params that are no longer needed.

The error handling corrections are partially implemented; see [this commit](https://github.com/vmware/splinterdb/pull/508/commits/43efdcf0fecb6da8b2c0f5718713acfcab60e9c9). 

I paused as this needs some unit-testing and I have provided a sample of such a unit-test here. We need better `errno`-type error reporting back from individual sub-systems [ here, clockcache ], so the unit-test can verify that the backout was triggered at the expected error location. See, e.g., the logic implemented in new test case `test_clockcache_init_fails_to_allocate_lookup_array()`.

I think there are few open items with this work that are best discussed in-person. Let's try to cover these items on Tue's PR meeting.

src/rc_allocator.h
gapisback: Took it out. And purged unnecessary `heap_handle *` variables and args in other code where `rc_allocator_init()` was being called.

tests/functional/test_splinter_shadow.c
gapisback: Took it out. I did some more cleanup to percolate up the stack, cleaning up fn-args like `hh', `hid` that are no longer needed. Some might have slipped thru ... I'll make another pass once this PR checks-in.

tests/functional/test_functionality.c
gapisback: Done. Deleted.

src/clockcache.c
gapisback: I misunderstood your comment. Yes, you are right; the calls to `platform_free()` will now only be done if the ptr-variable is non-NULL. Fixed.

src/rc_allocator.c
```diff
@@ -351,7 +351,7 @@ rc_allocator_init(rc_allocator      *al,
       return rc;
    }
    // To ensure alignment always allocate in multiples of page size.
-   uint32 buffer_size = cfg->extent_capacity * sizeof(uint8);
+   uint64 buffer_size = cfg->extent_capacity * sizeof(uint8);
```
gapisback, line 354: Potential bugfix. `buffer_size` declared as uint32 could overflow, to 0, for very large `extent_capacity`. (Discovered this while trying to build negative test cases in newly added, incomplete, `clockcache_test.c` unit test.

src/platform_linux/platform.c
```diff
@@ -82,12 +82,13 @@ platform_buffer_init(buffer_handle *bh, size_t length)
    }
 
    if (platform_use_mlock) {
-      int rc = mlock(bh->addr, length);
-      if (rc != 0) {
+      int mlock_rv = mlock(bh->addr, length);
```
gapisback, line 85: Rename existing `rc` which shadows the  new instance added at the top of this function (in an earlier commit of this PR cycle).

src/platform_linux/platform.c
```diff
@@ -82,12 +82,13 @@ platform_buffer_init(buffer_handle *bh, size_t length)
...
+      if (mlock_rv != 0) {
          platform_error_log("mlock (%lu bytes) failed with error: %s\n",
                             length,
                             strerror(errno));
          munmap(bh->addr, length);
+         rc = CONST_STATUS(errno);
```
gapisback, line 91: As suggested by Rob ... this is is better.

tests/unit/btree_stress_test.c
```diff
@@ -174,6 +172,7 @@ CTEST_TEARDOWN(btree_stress)
    clockcache_deinit(&data->cc);
    rc_allocator_deinit(&data->al);
    task_system_destroy(data->hid, &data->ts);
+   platform_heap_destroy(&data->hh);
```
gapisback, line 175: Missing in initial code; not an issue right now ... (Not sure why ASAN-builds are not tripping up ...)

src/platform_linux/platform.c
```diff
@@ -112,13 +114,16 @@ platform_buffer_getaddr(const buffer_handle *bh)
 platform_status
 platform_buffer_deinit(buffer_handle *bh)
 {
-   int ret;
-   ret = munmap(bh->addr, bh->length);
-   if (ret) {
-      return CONST_STATUS(errno);
-   }
+   if (bh->addr) {
```
gapisback, line 117: Discuss with @rtjohnso -- is this acceptable to you now?

src/platform_linux/platform.c
```diff
@@ -112,13 +114,16 @@ platform_buffer_getaddr(const buffer_handle *bh)
...
+      int ret;
+      ret = munmap(bh->addr, bh->length);
+      if (ret) {
+         return CONST_STATUS(errno);
+      }
 
-   bh->addr = NULL;
+      bh->addr = NULL;
+   }
+   bh->length = 0;
```
gapisback, line 126: This and L124 are added so that we can validate the state of resulting output `bh` in unit-tests.

tests/functional/test_functionality.c
```diff
@@ -632,19 +630,18 @@ cmp_ptrs(const void *a, const void *b)
...
 platform_status
-test_functionality(allocator           *al,
-                   io_handle           *io,
-                   cache               *cc[],
-                   trunk_config        *cfg,
-                   uint64               seed,
-                   uint64               num_inserts,
-                   uint64               correctness_check_frequency,
-                   task_system         *state,
-                   platform_heap_handle hh,
```
gapisback, line 642: Deleting this arg caused clang-format to throw a fit.

tests/unit/platform_apis_test.c
```diff
@@ -24,6 +26,22 @@ CTEST_DATA(platform_api)
...
+   if (Ctest_verbose) {
+      platform_set_log_streams(stdout, stderr);
+      CTEST_LOG_INFO("\nVerbose mode on.  This test exercises an error case, "
+                     "so on sucess it "
+                     "will print a message that appears to be an error.\n");
+   } else {
+      FILE *dev_null = fopen("/dev/null", "w");
+      ASSERT_NOT_NULL(dev_null);
+      platform_set_log_streams(dev_null, dev_null);
+   }
```
gapisback, line 43: This block will be simplified with just call to one function when PR #514 checks-in to `/main`.

tests/unit/platform_apis_test.c
```diff
@@ -45,13 +63,39 @@ CTEST2(platform_api, test_platform_buffer_init)
...
+{
+   platform_status rc = STATUS_NO_MEMORY;
+
+   buffer_handle bh;
+   ZERO_CONTENTS(&bh);
+
+   size_t length = (1024 * KiB * GiB);
+
+   // This is expected to fail as length is very large
+   rc = platform_buffer_init(&bh, length);
```
gapisback, line 93: From code inspection and manual testing, the length set on L90 is sufficiently big-enough to cause `mmap()` to fail w/ an error.

Should this approach be tightened further? (I think this is sufficient.)

src/clockcache.c
gapisback: Discussed this in PR meeting. Rework this as suggested by @robj. `clockcache_deinit()` will only call `platform_buffer_deinit()` only if corresponding pointers are non-NULL.

Remove the fix as applied in the latter function.

gapisback:
@rtjohnso - As per our discussion in the PR review meeting, I reworked the flow in `clockcache_deinit()` and adjusted the logic accordingly in `platform_buffer_deinit()`.

This [latest commit](https://github.com/vmware/splinterdb/pull/508/commits/f422a168ec175e4fd80619b47ebded653d3df327) (squashing 2 prev related-commits) should address the last remaining issue.

src/clockcache.c
```diff
@@ -1847,9 +1847,13 @@ clockcache_init(clockcache        *cc,   // OUT
       goto alloc_error;
    }
    cc->refcount = platform_buffer_getaddr(&cc->rc_bh);
+
    /* Separate ref counts for pins */
    cc->pincount =
       TYPED_ARRAY_ZALLOC(cc->heap_id, cc->pincount, cc->cfg->page_capacity);
+   if (!cc->pincount) {
+      goto alloc_error;
```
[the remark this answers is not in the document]
rtjohnso: I think we decided to abandon that level of detail in the testing, right?

rtjohnso (approved):
Thank you!  That cleaned up a lot of stuff.

gapisback:
@rtjohnso - About your very last comment, I'm only adding the extra check that in case we are unable to allocate memory to `cc->pincount()` via `TYPED_ARRAY_ZALLOC()`, that we bail, and go to error-backout target. (This is not test code; this is mainline library code.)

This was also (already) being done in `/main` here, further below, and also in other places in this `clockcache_init()` function which is allocation memory using `TYPED_ARRAY_ZALLOC()` or variants.

```
1860    cc->batch_busy =
1861       TYPED_ARRAY_ZALLOC(cc->heap_id,
1862                          cc->batch_busy,
1863                          cc->cfg->page_capacity / CC_ENTRIES_PER_BATCH);
1864    if (!cc->batch_busy) {
1865       goto alloc_error;
1866    }
```

src/clockcache.c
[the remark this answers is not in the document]
gapisback: yes, you are right. I withdrew the newly developed `clockcache_test` unit-test before merging this to /main.

Issue #509, opened 2022-12-22
Assertion failures trip in clockcache.c due to off-by-1 issues handling MAX_THREADS.


gapisback:
Couple of spots in `clockcache.c` trip with assertions when running max # of configured threads, 63. These issues were discovered while running integration tests with large # of configured threads. 

These are easily reproduced using `driver_test splinter_test --perf --num-insert-threads 63`.

Here is the 1st instance:

```
sdb-fdb-build:[8] $ ./build/debug/bin/driver_test splinter_test --perf --num-insert-threads 63 --tree-size-gib 1
./build/debug/bin/driver_test: splinterdb_build_version f070c377
Dispatch test splinter_test
Fingerprint size 29 too large, max value size is 5, setting to 27
filter-index-size: 256 is too small, setting to 512
Bumped up IO queue size to 4032
Bumped up IO queue size to 4032
Running splinter_test with 1 caches
splinter_test: SplinterDB performance test started with 1 tables
splinter_perf_inserts() starting num_insert_threads=63, num_threads=63, num_inserts=13594624 (~13 million) ...
Thread 55 inserting   0% complete for table 0 ... OS-pid=3418189, OS-tid=3418245, Thread-ID=63, Assertion failed at src/clockcache.c:1669:clockcache_get_free_page(): "(tid < (64) - 1)". Invalid tid=63

Aborted (core dumped)
```

The extra message reporting `Invalid tid=63` is added to /main to see what's going wrong.

Once having fixed this instance, we next fail here:

```
sdb-fdb-build:[9] $ ./build/debug/bin/driver_test splinter_test --perf --num-insert-threads 63 --tree-size-gib 1
./build/debug/bin/driver_test: splinterdb_build_version f070c377
Dispatch test splinter_test
Fingerprint size 29 too large, max value size is 5, setting to 27
filter-index-size: 256 is too small, setting to 512
Bumped up IO queue size to 4032
Bumped up IO queue size to 4032
Running splinter_test with 1 caches
splinter_test: SplinterDB performance test started with 1 tables
splinter_perf_inserts() starting num_insert_threads=63, num_threads=63, num_inserts=13594624 (~13 million) ...
Thread 19 inserting   0% complete for table 0 ...
Thread 60 inserting   0% complete for table 0 ...
[...]

PR #510, opened 2022-12-22
(#509) Fix off-by-1 issues re. MAX_THREADS in clockcache.c

gapisback:
This commit fixes off-by-1 issues checking MAX_THREADS value in assertions in `clockcache.c` . These problems are easily reproduced with `driver_test splinter_test ` with max # of threads. A new test execution with --num-insert-threads 63 is added to nightly runs, in test.sh .

----

The issue #509 fully describes the repro and assertions seen.

These problems were discovered during an in-flight integration while exercising `MAX_THREADS` threads. These fixes were applied to the dev-branch and allowed testing to continue.

Integrating these to /main will help improve overall stability.

### DIFF
--- a/src/clockcache.c
+++ b/src/clockcache.c
@@ -816,13 +816,25 @@ clockcache_inc_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
 static inline void
 clockcache_dec_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
 {
+   debug_only threadid input_counter_no = counter_no;
+
    counter_no %= CC_RC_WIDTH;
    uint64 rc_number = clockcache_get_ref_internal(cc, entry_number);
-   debug_assert(rc_number < cc->cfg->page_capacity);
+   debug_assert((rc_number < cc->cfg->page_capacity),
+                "Entry number, %lu, is out of allocator "
+                "page capacity range, %u.\n",
+                rc_number,
+                cc->cfg->page_capacity);
 
    debug_only uint16 refcount = __sync_fetch_and_sub(
       &cc->refcount[counter_no * cc->cfg->page_capacity + rc_number], 1);
-   debug_assert(refcount != 0);
+   debug_assert((refcount != 0),
+                "Invalid refcount, %u, after decrement."
+                " input counter_no=%lu, rc_number=%lu, counter_no=%lu\n",
+                refcount,
+                input_counter_no,
+                rc_number,
+                counter_no);
 }
 
 static inline uint8
@@ -868,7 +880,7 @@ clockcache_assert_no_refs(clockcache *cc)
 {
    threadid        i;
    volatile uint32 j;
-   for (i = 0; i < (MAX_THREADS - 1); i++) {
+   for (i = 0; i < MAX_THREADS; i++) {
       for (j = 0; j < cc->cfg->page_capacity; j++) {
          if (clockcache_get_ref(cc, j, i) != 0) {
             clockcache_get_ref(cc, j, i);
@@ -1304,7 +1316,7 @@ clockcache_batch_start_writeback(clockcache *cc, uint64 batch, bool is_urgent)
 
    clockcache_entry *entry, *next_entry;
 
-   debug_assert(tid < MAX_THREADS - 1);
+   debug_assert((tid < MAX_THREADS), "Invalid tid=%lu\n", tid);
    debug_assert(cc != NULL);
    debug_assert(batch < cc->cfg->page_capacity / CC_ENTRIES_PER_BATCH);
 
@@ -1594,7 +1606,7 @@ clockcache_get_free_page(clockcache *cc,
    clockcache_entry *entry;
    timestamp         wait_start;
 
-   debug_assert(tid < MAX_THREADS);
+   debug_assert((tid < MAX_THREADS), "Invalid tid=%lu\n", tid);
    if (cc->per_thread[tid].free_hand == CC_UNMAPPED_ENTRY) {
       clockcache_move_hand(cc, FALSE);
    }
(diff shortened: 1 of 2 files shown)


src/clockcache.c
```diff
@@ -863,16 +863,41 @@ clockcache_inc_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
    debug_assert(refcount != MAX_READ_REFCOUNT);
 }
 
+/*
+ * Caller-macro to figure out who's calling this dec_ref interface.
+ *
+ * clockcache_dec_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
+ */
+#define clockcache_dec_ref(cc, entry_number, counter_no)                       \
```
gapisback, line 871: This change isn't directly related to this fix, per se, but if you invoke:

`driver_test splinter_test --perf --num-insert-threads 63 --num-lookup-threads 63 ...`

We will run into an assertion on L893, complaining about ref-count going to 0 after a dec-ref.

I need to log a diff issue to trace this, and so this instrumentation is added to help debug that separate issue.

src/clockcache.c
```diff
@@ -918,7 +943,7 @@ clockcache_assert_no_refs(clockcache *cc)
 {
    threadid        i;
    volatile uint32 j;
-   for (i = 0; i < (MAX_THREADS - 1); i++) {
+   for (i = 0; i < MAX_THREADS; i++) {
```
gapisback, line 946: 1st instance of an off-by-1 check.

src/clockcache.c
```diff
@@ -1354,7 +1379,7 @@ clockcache_batch_start_writeback(clockcache *cc, uint64 batch, bool is_urgent)
 
    clockcache_entry *entry, *next_entry;
 
-   debug_assert(tid < MAX_THREADS - 1);
+   debug_assert((tid < MAX_THREADS), "Invalid tid=%lu\n", tid);
```
gapisback, line 1382: 2nd instance of an off-by-1 check.

src/task.c
```diff
@@ -91,10 +91,11 @@ task_init_threadid(task_system *ts)
    }
 
 out:
+   // The main program counts for thread-0; hence limit is "-1"
    platform_assert((tid != INVALID_TID),
                    "Cannot create a new thread as the limit on"
                    " concurrent threads, %d, will be exceeded.\n",
-                   MAX_THREADS);
+                   (MAX_THREADS - 1));
```
gapisback, line 98: Slightly unrelated fix. 

Without this fix, we will be reporting an error like so, when  `--num-insert-threads 64` is used:

```
Assertion failed at src/task.c:95:task_init_threadid(): "(tid != INVALID_TID)". Cannot create a new thread as the limit on concurrent threads, 64, will be exceeded.
```

The real-limit (externally) is 63 (user + bg-threads)

test.sh
```diff
@@ -270,6 +270,21 @@ function nightly_sync_perf_tests() {
...
+    # Exercise a case with max # of insert-threads which tripped an assertion
+    # This isn't really a 'perf' test but a regression / stability test exec
+    nins_t=63
+    nrange_lookup_t=0
+    test_descr="${nins_t} insert threads"
+    dbname="splinter_test.max_threads.db"
+
+    run_with_timing "Performance with max-threads ${test_descr}" \
+            "$BINDIR"/driver_test splinter_test --perf \
+                                                --num-insert-threads ${nins_t} \
```
gapisback, line 283: Adding this case to nightly runs. Even though they run on VMs w/  8 CPUs, we don't particularly care how long this will take to run.

Also, use of `--tree-size-gib 1` reduces # of rows to 1M , which allows the test to complete in a reasonable amount of time.

src/clockcache.c
```diff
@@ -808,16 +808,41 @@ clockcache_inc_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
...
+
    counter_no %= CC_RC_WIDTH;
    uint64 rc_number = clockcache_get_ref_internal(cc, entry_number);
-   debug_assert(rc_number < cc->cfg->page_capacity);
+   debug_assert((rc_number < cc->cfg->page_capacity),
+                "[%d] Entry number, %lu, is out of allocator "
+                "capacity range, %u.\n",
+                line,
+                rc_number,
+                cc->cfg->page_capacity);
```
ajhconway, line 834: I like these more verbose assertions, but please take out the line numbers.

src/clockcache.c
ajhconway: The history of these is that we used to re-use the `(MAX_THREADS - 1)` counter for pins.

src/clockcache.c
ajhconway: I don't see the assertion you're talking about on L893 either in the before or after of this PR.

ajhconway (changes requested):
This is a good catch, but please remove the debugging code.

src/clockcache.c
gapisback: Reverted the debug code as suggested. New diffs to be uploaded soon.

I don't recall which assertion was tripping. I should have documented it better here when I ran into it.
I'll try to repro this later on ...

src/clockcache.c
```diff
@@ -808,16 +808,41 @@ clockcache_inc_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
...
+   clockcache_dec_refcount(cc, entry_number, counter_no, __LINE__)
+
 static inline void
-clockcache_dec_ref(clockcache *cc, uint32 entry_number, threadid counter_no)
+clockcache_dec_refcount(clockcache *cc,
+                        uint32      entry_number,
+                        threadid    counter_no,
+                        int         line)
 {
+   debug_only threadid inp_counter_no = counter_no;
```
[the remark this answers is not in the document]
gapisback: Done.

ajhconway (approved):

rtjohnso (approved):
LGTM

Issue #511, opened 2022-12-23
Add utility routines to print 'size' bytes value in human-readable string with unit-specifiers.


gapisback:
In various places, we store 'size' in units of bytes, which tend to be printed as big numbers, or rounded-off to print in `MiB`, `GiB` etc.

This issue is logged to add facilities to print such values in a user-output buffers with unit-specifiers. The print method should take care of using the appropriate unit-specifiers.

Examples: Print 1024 as `1 KiB`, Print 2048 as `2 MiB` and so on.
Print, e.g, 1280 bytes as `1.25 KiB`. Rather than printing big values, instead print something like `1.75 GiB`.

This convenience function was developed in a prototype dev-project and found to be useful for debugging and diagnostics.
Productize and integrate this to `/main`.

PR #512, opened 2022-12-23
(#511) Add fns to print 'size' as human-readable string w/ unit-specifiers.

gapisback:
This commit adds couple of utility functions to convert 'size' unit to a human-readable string with unit-specifiers, in an output buffer.

- size_to_str() - Convert 'size' to a string in an output buffer
- size_to_fmtstr() - Same as above, using user-specified format-string. Useful to generate output enclosed in '(%s)'.

Add utility bytes to Units conversion macros. Add test cases to exercise these interfaces. Apply these utility fns in couple of stats-printing methods, to display values in units of bytes as a string.

---
NOTE: This work was done in an in-flight integration development branch, to closely track memory usage (free / allocation metrics). The utility methods have been formalized, productized with new unit-tests. 

This commit applies the conversion functions to couple of existing print functions. These helper conversion functions can become useful in other places, too.

---- Updated (28.Dec.2021) ---- this work has now been rebased on top of another in-flight PR #521 . And the new changes to print with unit-specifiers has been applied to `btree_print_btree_pivot_stats()`. See the sample outputs shown in the diff block associated with this file.

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -3083,6 +3083,10 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
    platform_log(log_handle, "\n");
 }
 
+// Macro to deal with printing Pivot stats 'sz' bytes as uninitialized
+#define PIVOT_STATS_BYTES_AS_STR(sz)                                           \
+   (((sz) == BTREE_UNKNOWN_COUNTER) ? "BTREE_UNKNOWN_COUNTER" : size_str((sz)))
+
 static void
 btree_print_btree_pivot_stats(platform_log_handle *log_handle,
                               btree_pivot_stats   *pivot_stats)
@@ -3099,11 +3103,13 @@ btree_print_btree_pivot_stats(platform_log_handle *log_handle,
    // Indentation is dictated by outer caller
    platform_log(log_handle,
                 "   (num_kvs=%u\n"
-                "    key_bytes=%u\n"
-                "    message_bytes=%u)\n",
+                "    key_bytes=%u (%s)\n"
+                "    message_bytes=%u (%s))\n",
                 pivot_stats->num_kvs,
                 pivot_stats->key_bytes,
-                pivot_stats->message_bytes);
+                PIVOT_STATS_BYTES_AS_STR(pivot_stats->key_bytes),
+                pivot_stats->message_bytes,
+                PIVOT_STATS_BYTES_AS_STR(pivot_stats->message_bytes));
 }
 
 static void
@@ -3255,6 +3261,7 @@ btree_print_subtree(platform_log_handle *log_handle,
    node.addr = addr;
    btree_print_node(log_handle, cc, cfg, &node);
    if (!allocator_page_valid(cache_get_allocator(cc), node.addr)) {
+      platform_log(log_handle, "Unallocated BTree node addr=%lu\n", addr);
       return;
    }
    btree_node_get(cc, cfg, &node, PAGE_TYPE_BRANCH);
--- a/src/platform_linux/platform.h
+++ b/src/platform_linux/platform.h
@@ -88,14 +88,27 @@
 
 // Data unit constants
 #define KiB (1024UL)
-#define MiB (KiB * KiB)
-#define GiB (MiB * KiB)
+#define MiB (KiB * 1024)
+#define GiB (MiB * 1024)
+#define TiB (GiB * 1024)
 
+// Convert 'x' in unit-specifiers to bytes
 #define KiB_TO_B(x) ((x)*KiB)
 #define MiB_TO_B(x) ((x)*MiB)
 #define GiB_TO_B(x) ((x)*GiB)
+#define TiB_TO_B(x) ((x)*TiB)
+
+// Convert 'x' in bytes to 'int'-value with unit-specifiers
+#define B_TO_KiB(x) ((x) / KiB)
 #define B_TO_MiB(x) ((x) / MiB)
 #define B_TO_GiB(x) ((x) / GiB)
+#define B_TO_TiB(x) ((x) / TiB)
+
+// For x bytes, returns as int the fractional portion modulo a unit-specifier
+#define B_TO_KiB_FRACT(x) ((100 * ((x) % KiB)) / KiB)
+#define B_TO_MiB_FRACT(x) ((100 * ((x) % MiB)) / MiB)
+#define B_TO_GiB_FRACT(x) ((100 * ((x) % GiB)) / GiB)
+#define B_TO_TiB_FRACT(x) ((100 * ((x) % TiB)) / TiB)
 
 // Time unit constants
 #define THOUSAND (1000UL)
--- a/src/rc_allocator.c
+++ b/src/rc_allocator.c
@@ -777,50 +777,52 @@ rc_allocator_assert_noleaks(rc_allocator *al)
 void
 rc_allocator_print_stats(rc_allocator *al)
 {
-   int64 divider = GiB / al->cfg->io_cfg->extent_size;
+   // clang-format off
+   const char *dashes = "-------------------------------------------------------------------";
+   platform_default_log("|%s|\n", dashes);
+   platform_default_log("| Allocator Stats                                                   |\n");
+   platform_default_log("|%s|\n", dashes);
+   // clang-format on
+
+   uint64 extent_size = al->cfg->io_cfg->extent_size; // bytes
    platform_default_log(
-      "----------------------------------------------------------------\n");
-   platform_default_log(
-      "| Allocator Stats                                              |\n");
-   platform_default_log(
-      "|--------------------------------------------------------------|\n");
-   uint64 curr_gib = al->stats.curr_allocated / divider;
-   platform_default_log(
-      "| Currently Allocated: %12lu extents (%4luGiB)          |\n",
+      "| Currently Allocated: %12lu extents %-14s          |\n",
       al->stats.curr_allocated,
-      curr_gib);
-   uint64 max_gib = al->stats.max_allocated / divider;
+      size_fmtstr("(%s)", (al->stats.curr_allocated * extent_size)));
+
    platform_default_log(
-      "| Max Allocated:       %12lu extents (%4luGiB)          |\n",
+      "| Max Allocated:       %12lu extents %-14s          |\n",
       al->stats.max_allocated,
-      max_gib);
-   platform_default_log(
-      "|--------------------------------------------------------------|\n");
-   platform_default_log(
-      "| Page Type | Allocations | Deallocations | Footprint (bytes)  |\n");
-   platform_default_log(
-      "|--------------------------------------------------------------|\n");
+      size_fmtstr("(%s)", (al->stats.max_allocated * extent_size)));
+
+   // clang-format off
+   platform_default_log("|%s|\n", dashes);
+   platform_default_log("| Page Type  | Allocations | Deallocations |      Footprint         |\n");
+   platform_default_log("|            |      (Number of extents)    | # extents  (bytes)     |\n");
+   platform_default_log("|%s|\n", dashes);
+   // clang-format on
+
    int64 exp_allocated_count = 0;
    for (page_type type = PAGE_TYPE_FIRST; type < NUM_PAGE_TYPES; type++) {
-      const char *str           = page_type_str[type];
-      int64       allocs        = al->stats.extent_allocs[type];
-      int64       deallocs      = al->stats.extent_deallocs[type];
-      int64       footprint     = allocs - deallocs;
-      int64       footprint_gib = footprint / divider;
+      const char *str       = page_type_str[type];
+      int64       allocs    = al->stats.extent_allocs[type];
+      int64       deallocs  = al->stats.extent_deallocs[type];
+      int64       footprint = allocs - deallocs;
 
       exp_allocated_count += footprint;
 
-      platform_default_log("| %-10s | %11ld | %13ld | %8ld (%4ldGiB) |\n",
+      platform_default_log("| %-10s | %11ld | %13ld | %8ld %14s|\n",
                            str,
                            allocs,
                            deallocs,
                            footprint,
-                           footprint_gib);
+                           size_fmtstr("(%s)", (footprint * extent_size)));
    }
+   platform_default_log("|%s|\n", dashes);
    platform_default_log(
-      "----------------------------------------------------------------\n");
-   platform_default_log("Expected allocation count from footprint = %ld\n",
-                        exp_allocated_count);
+      "Expected count of extents in-use from footprint = %ld extents (%s)\n",
+      exp_allocated_count,
+      size_str(exp_allocated_count * extent_size));
 }
 
 /*
@@ -846,17 +848,20 @@ rc_allocator_print_allocated(rc_allocator *al)
    platform_default_log("   Index  ExtentAddr  Count\n");
 
    // # of extents with non-zero referenced page-count found
-   uint64 found = 0;
+   uint64 nextents_found = 0;
+   uint64 extent_size    = al->cfg->io_cfg->extent_size;
 
    for (i = 0; i < al->cfg->extent_capacity; i++) {
       ref = al->ref_count[i];
       if (ref != 0) {
-         found++;
-         uint64 ext_addr = (i * al->cfg->io_cfg->extent_size);
+         nextents_found++;
+         uint64 ext_addr = (i * extent_size);
          platform_default_log("%8lu %12lu     %u\n", i, ext_addr, ref);
       }
    }
-   platform_default_log("%sFound %lu extents with allocated pages.\n",
+
+   platform_default_log("%sFound %lu extents (%s) with allocated pages.\n",
                         (print_curly ? "}\n" : ""),
-                        found);
+                        nextents_found,
+                        size_str(nextents_found * extent_size));
 }
--- a/src/rc_allocator.h
+++ b/src/rc_allocator.h
@@ -44,8 +44,8 @@ _Static_assert(offsetof(rc_allocator_meta_page, splinters) == 0,
  *----------------------------------------------------------------------
  */
 typedef struct rc_allocator_stats {
-   int64 curr_allocated;
-   int64 max_allocated;
+   int64 curr_allocated; // # of extents allocated
+   int64 max_allocated;  // # of extents allocated high-water mark
    int64 extent_allocs[NUM_PAGE_TYPES];
    int64 extent_deallocs[NUM_PAGE_TYPES];
 } rc_allocator_stats;
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -7905,8 +7905,11 @@ trunk_print_space_use(platform_log_handle *log_handle, trunk_handle *spl)
                 "Space used by level: trunk_tree_height=%d\n",
                 trunk_tree_height(spl));
    for (uint16 i = 0; i <= trunk_tree_height(spl); i++) {
-      platform_log(
-         log_handle, "%u: %8lu MiB\n", i, B_TO_MiB(bytes_used_by_level[i]));
+      platform_log(log_handle,
+                   "%u: %lu bytes (%s)\n",
+                   i,
+                   bytes_used_by_level[i],
+                   size_str(bytes_used_by_level[i]));
    }
    platform_log(log_handle, "\n");
 }
@@ -8851,7 +8854,7 @@ trunk_node_print_branches(trunk_handle *spl, uint64 addr, void *arg)
          spl, &node, branch_no, &num_tuples_in_branch, &kv_bytes_in_branch);
       uint64 kib_in_branch = 0;
       // trunk_branch_extent_count(spl, &node, branch_no);
-      kib_in_branch *= trunk_extent_size(&spl->cfg) / 1024;
+      kib_in_branch *= B_TO_KiB(trunk_extent_size(&spl->cfg));
       fraction space_amp =
          init_fraction(kib_in_branch * 1024, kv_bytes_in_branch);
       platform_log(
@@ -8861,7 +8864,7 @@ trunk_node_print_branches(trunk_handle *spl, uint64 addr, void *arg)
          branch_no,
          addr,
          num_tuples_in_branch,
-         kv_bytes_in_branch / 1024,
+         B_TO_KiB(kv_bytes_in_branch),
          kib_in_branch,
          FRACTION_ARGS(space_amp));
    }
--- a/src/util.c
+++ b/src/util.c
@@ -364,3 +364,64 @@ debug_hex_dump_slice(platform_log_handle *plh, uint64 grouping, slice data)
 {
    debug_hex_dump(plh, grouping, slice_length(data), slice_data(data));
 }
+
+/*
+ * Format a size value with unit-specifiers, in an output buffer.
+ * Returns 'outbuf', as ptr to size-value snprintf()'ed as a string.
+ */
+char *
+size_to_str(char *outbuf, size_t outbuflen, size_t size)
+{
+   debug_assert(outbuflen >= SIZE_TO_STR_LEN, "outbuflen=%lu.\n", outbuflen);
+   size_t unit_val  = 0;
+   size_t frac_val  = 0;
+   bool   is_approx = FALSE;
+
+   char *units = NULL;
+   if (size >= TiB) {
+      unit_val  = B_TO_TiB(size);
+      frac_val  = B_TO_TiB_FRACT(size);
+      is_approx = (size > TiB_TO_B(unit_val));
+      units     = "TiB";
+
+   } else if (size >= GiB) {
+      unit_val  = B_TO_GiB(size);
+      frac_val  = B_TO_GiB_FRACT(size);
+      is_approx = (size > GiB_TO_B(unit_val));
+      units     = "GiB";
+
+   } else if (size >= MiB) {
+      unit_val  = B_TO_MiB(size);
+      frac_val  = B_TO_MiB_FRACT(size);
+      is_approx = (size > MiB_TO_B(unit_val));
+      units     = "MiB";
+
+   } else if (size >= KiB) {
+      unit_val  = B_TO_KiB(size);
+      frac_val  = B_TO_KiB_FRACT(size);
+      is_approx = (size > KiB_TO_B(unit_val));
+      units     = "KiB";
+   } else {
+      unit_val = size;
+      units    = "bytes";
+   }
+
+   if (frac_val || is_approx) {
+      snprintf(outbuf, outbuflen, "~%ld.%ld %s", unit_val, frac_val, units);
+   } else {
+      snprintf(outbuf, outbuflen, "%ld %s", unit_val, units);
+   }
+   return outbuf;
+}
+
+/*
+ * Sibling of size_to_str(), but uses user-provided print format specifier.
+ * 'fmtstr' is expected to have just one '%s', and whatever other text user
+ * wishes to print with the output string.
+ */
+char *
+size_to_fmtstr(char *outbuf, size_t outbuflen, const char *fmtstr, size_t size)
+{
+   snprintf(outbuf, outbuflen, fmtstr, size_str(size));
+   return outbuf;
+}
--- a/src/util.h
+++ b/src/util.h
@@ -382,4 +382,40 @@ debug_hex_dump_slice(platform_log_handle *, uint64 grouping, slice data);
     : ((intval) < 1000) ? "3d"                                                 \
                         : "4d")
 
+// Length of output buffer to snprintf()-into size as string w/ unit specifier
+#define SIZE_TO_STR_LEN 20
+
+// Format a size value with unit-specifiers, in an output buffer.
+char *
+size_to_str(char *outbuf, size_t outbuflen, size_t size);
+
+char *
+size_to_fmtstr(char *outbuf, size_t outbuflen, const char *fmtstr, size_t size);
+
+/*
+ * Convenience caller macros to convert 'sz' bytes to return a string,
+ * formatting the input size as human-readable value with unit-specifiers.
+ */
+// char *size_str(size_t sz)
+#define size_str(sz)                                                           \
+   (({                                                                         \
+       struct {                                                                \
+          char buffer[SIZE_TO_STR_LEN];                                        \
+       } onstack_chartmp;                                                      \
+       size_to_str(                                                            \
+          onstack_chartmp.buffer, sizeof(onstack_chartmp.buffer), sz);         \
+       onstack_chartmp;                                                        \
+    }).buffer)
+
+// char *size_fmtstr(const char *fmtstr, size_t sz)
+#define size_fmtstr(fmtstr, sz)                                                \
+   (({                                                                         \
+       struct {                                                                \
+          char buffer[SIZE_TO_STR_LEN];                                        \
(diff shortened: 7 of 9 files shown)


src/platform_linux/platform.h
```diff
@@ -88,14 +88,37 @@
 
 // Data unit constants
 #define KiB (1024UL)
-#define MiB (KiB * KiB)
-#define GiB (MiB * KiB)
+#define MiB (KiB * 1024)
+#define GiB (MiB * 1024)
+#define TiB (GiB * 1024)
```
gapisback, line 93: These diffs extend the existing conversion macros to be complete.

src/platform_linux/platform.h
```diff
@@ -88,14 +88,37 @@
...
+#define TiB_TO_B(x) ((x)*TiB)
+
+// Convert 'x' in bytes to 'int'-value with unit-specifiers
+#define B_TO_KiB(x) ((x) / KiB)
 #define B_TO_MiB(x) ((x) / MiB)
 #define B_TO_GiB(x) ((x) / GiB)
+#define B_TO_TiB(x) ((x) / TiB)
+
+// Return, as int, the fractional portion modulo a KiB for given x bytes.
+#define B_TO_KiB_FRACT(x)                                                      \
```
gapisback, line 108: This family of macros are used to generate the decimal portion of `bytes` when converted to a size-specifier unit.

src/rc_allocator.c
```diff
@@ -769,50 +769,61 @@ rc_allocator_assert_noleaks(rc_allocator *al)
...
+   platform_default_log("|%s|\n", dashes);
+   // clang-format on
+
+   char   bytes_to_str[SIZE_TO_STR_LEN];
+   size_t nbytes = (al->stats.curr_allocated * al->cfg->io_cfg->extent_size);
+   size_to_fmtstr(bytes_to_str, sizeof(bytes_to_str), "(%s)", nbytes);
+
    platform_default_log(
-      "| Currently Allocated: %12lu extents (%4luGiB)          |\n",
+      "| Currently Allocated: %12lu extents %-14s          |\n",
```
gapisback, line 784: This change prints a more accurate value, rather than just printing as n-`GiBs`.

src/trunk.c
```diff
@@ -7906,8 +7906,13 @@ trunk_print_space_use(platform_log_handle *log_handle, trunk_handle *spl)
...
    for (uint16 i = 0; i <= trunk_tree_height(spl); i++) {
-      platform_log(
-         log_handle, "%u: %8lu MiB\n", i, B_TO_MiB(bytes_used_by_level[i]));
+      char size_str[SIZE_TO_STR_LEN];
+      size_to_str(size_str, sizeof(size_str), bytes_used_by_level[i]);
+      platform_log(log_handle,
+                   "%u: %lu bytes (%s)\n",
+                   i,
+                   bytes_used_by_level[i],
+                   size_str);
```
gapisback, line 7915: A typical example of how to use this conversion function.

src/trunk.c
```diff
@@ -8852,7 +8857,7 @@ trunk_node_print_branches(trunk_handle *spl, uint64 addr, void *arg)
          spl, &node, branch_no, &num_tuples_in_branch, &kv_bytes_in_branch);
       uint64 kib_in_branch = 0;
       // trunk_branch_extent_count(spl, &node, branch_no);
-      kib_in_branch *= trunk_extent_size(&spl->cfg) / 1024;
+      kib_in_branch *= B_TO_KiB(trunk_extent_size(&spl->cfg));
```
gapisback, line 8860: Use standardized conversion macros.

src/util.c
```diff
@@ -364,3 +364,64 @@ debug_hex_dump_slice(platform_log_handle *plh, uint64 grouping, slice data)
 {
    debug_hex_dump(plh, grouping, slice_length(data), slice_data(data));
 }
+
+/*
+ * Format a size value with unit-specifiers, in an output buffer.
+ */
+void
+size_to_str(char *outbuf, size_t outbuflen, size_t size)
```
gapisback, line 372: Core utility function.

tests/unit/misc_test.c
```diff
@@ -183,13 +190,84 @@ CTEST2(misc, test_ctest_assert_prints_user_msg_with_params)
    platform_close_log_stream(&stream, data->log_output);
 }
 
+/*
+ * Test case to verify conversion of size value to a string with
+ * unit-specifiers. Do quick cross-check of the lower-level conversion macros
+ * involved.
+ */
+CTEST2(misc, test_size_to_str)
```
gapisback, line 198: new Unit test case.

src/rc_allocator.c
```diff
@@ -769,50 +769,61 @@ rc_allocator_assert_noleaks(rc_allocator *al)
...
       al->stats.curr_allocated,
-      curr_gib);
-   uint64 max_gib = al->stats.max_allocated / divider;
+      bytes_to_str);
+
+   nbytes = (al->stats.max_allocated * al->cfg->io_cfg->extent_size);
+   size_to_fmtstr(bytes_to_str, sizeof(bytes_to_str), "(%s)", nbytes);
    platform_default_log(
-      "| Max Allocated:       %12lu extents (%4luGiB)          |\n",
+      "| Max Allocated:       %12lu extents %-14s          |\n",
```
gapisback, line 791: Example output from `./build/release/bin/unit/splinter_test test_splinter_print_diags` is as below, for this line and for L784 above:

```
| Currently Allocated:        10495 extents (~1.28 GiB)             |
| Max Allocated:              11193 extents (~1.36 GiB)             |
```

Now you get `~1.28 GiB`, whereas previously you'd get `1 GiB`. New output is slightly more accurate.

src/rc_allocator.c
```diff
@@ -769,50 +769,61 @@ rc_allocator_assert_noleaks(rc_allocator *al)
...
-   platform_default_log(
-      "| Page Type | Allocations | Deallocations | Footprint (bytes)  |\n");
-   platform_default_log(
-      "|--------------------------------------------------------------|\n");
+      bytes_to_str);
+
+   // clang-format off
+   platform_default_log("|%s|\n", dashes);
+   platform_default_log("| Page Type  | Allocations | Deallocations |      Footprint         |\n");
+   platform_default_log("|            |      (Number of extents)    | # extents  (bytes)     |\n");
```
gapisback, line 798: Clarified in banner line that both `Allocations | Deallocations` are in units of `(Number of extents)`.

And that `Footprint` is reported in two units: ` # extents  (bytes)`

src/rc_allocator.c
```diff
@@ -769,50 +769,61 @@ rc_allocator_assert_noleaks(rc_allocator *al)
...
       exp_allocated_count += footprint;
 
-      platform_default_log("| %-10s | %11ld | %13ld | %8ld (%4ldGiB) |\n",
+      platform_default_log("| %-10s | %11ld | %13ld | %8ld %-14s|\n",
                            str,
                            allocs,
                            deallocs,
                            footprint,
-                           footprint_gib);
+                           bytes_to_str);
```
gapisback, line 818: The new output for these rows when running `./build/release/bin/unit/splinter_test test_splinter_print_diags` is shown below:

```
| Page Type  | Allocations | Deallocations |      Footprint         |
|            |      (Number of extents)    | # extents  (bytes)     |
|-------------------------------------------------------------------|
| trunk      |          10 |             0 |       10 (~1.25 MiB)   |
| branch     |       26506 |         16602 |     9904 (~1.20 GiB)   |
| memtable   |       18827 |         18432 |      395 (~49.37 MiB)  |
| filter     |        2375 |          2193 |      182 (~22.75 MiB)  |
| log        |           0 |             0 |        0 (0 bytes)     |
| superblock |           3 |             0 |        3 (384 KiB)     |
| misc       |           0 |             0 |        0 (0 bytes)     |
| lock       |           1 |             0 |        1 (128 KiB)     |
```

We get better sizes reported, showing in `MiB` and `GiB` where applicable.

Previously some of these were being reported as 0, which is slightly misleading.

src/rc_allocator.c
```diff
@@ -769,50 +769,61 @@ rc_allocator_assert_noleaks(rc_allocator *al)
...
+   platform_default_log("|%s|\n", dashes);
+   nbytes = (exp_allocated_count * al->cfg->io_cfg->extent_size);
+   size_to_str(bytes_to_str, sizeof(bytes_to_str), nbytes);
    platform_default_log(
-      "----------------------------------------------------------------\n");
-   platform_default_log("Expected allocation count from footprint = %ld\n",
-                        exp_allocated_count);
+      "Expected total allocation count from footprint = %ld extents %s\n",
+      exp_allocated_count,
+      bytes_to_str);
```
gapisback, line 826: New output:

`Expected total allocation count from footprint = 10495 extents ~1.28 GiB`

src/rc_allocator.c
```diff
@@ -848,7 +859,12 @@ rc_allocator_print_allocated(rc_allocator *al)
...
-   platform_default_log("%sFound %lu extents with allocated pages.\n",
+
+   char   bytes_to_str[SIZE_TO_STR_LEN];
+   uint64 nbytes = (found * al->cfg->io_cfg->extent_size);
+   size_to_str(bytes_to_str, sizeof(bytes_to_str), nbytes);
+   platform_default_log("%sFound %lu extents (%s) with allocated pages.\n",
                         (print_curly ? "}\n" : ""),
-                        found);
+                        found,
+                        bytes_to_str);
```
gapisback, line 869: New output from this change:

`Found 10495 extents (~1.28 GiB) with allocated pages.`

src/trunk.c
gapisback, line 7915: New output from this change (for same test case run):

```
Space used by level: trunk_tree_height=1
0: 1123287040 bytes (~1.4 GiB)
1: 172621824 bytes (~164.62 MiB)
```

src/btree.c
```diff
@@ -3083,13 +3083,66 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
...
+btree_print_btree_pivot_stats(platform_log_handle *log_handle,
+                              btree_pivot_stats   *pivot_stats)
+{
+   // If nothing has been updated, yet, print a msg accordingly.
+   if ((pivot_stats->num_kvs == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->key_bytes == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->message_bytes == BTREE_UNKNOWN_COUNTER))
+   {
+      platform_log(log_handle, "   (Pivot Stats=BTREE_UNKNOWN_COUNTER)\n");
+      return;
```
gapisback, line 3096: This little chunk will come ahead of this PR under another in-flight change-set for PR #521 . (Ignore this diff for this review.)

src/btree.c
```diff
@@ -3083,13 +3083,66 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
...
+
+   size_t nmbytes = pivot_stats->message_bytes;
+   char   nmbytes_str[SIZE_TO_STR_LEN];
+   platform_log(log_handle,
+                "   (num_kvs=%u\n"
+                "    key_bytes=%lu (%s)\n"
+                "    message_bytes=%lu (%s))\n",
+                pivot_stats->num_kvs,
+                nkbytes,
+                size_to_str(nkbytes_str, sizeof(nkbytes_str), nkbytes),
```
gapisback, line 3111: The newly added interface to print with unit-specifiers is applied here to print BTree pivot stats fields. The new output will look like follows, as generated by `./build/release/bin/unit/btree_stress_test`, which does a `btree_pack()`, which will set these values:

```
BTree node at addr=1333133312
{
**  INDEX NODE
**  Header ptr: 0x7f55c36ce000
**  addr: 1333133312
**  next_addr: 0
[...]
-------------------
Offset Table num_entries=3
[0] 4010    [1] 3884    [2] 3787
-------------------
Array of 3 index entries:
[ 0]: key=0x00000d093a9a17000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
   child_addr=1333657600
   (num_kvs=377607
    key_bytes=21716772 (~20.71 MiB)
    message_bytes=260863299 (~248.77 MiB))
[ 1]: key=0x60aaf36fb8dd4d000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
   child_addr=1333661696
   (num_kvs=379751
    key_bytes=21839065 (~20.82 MiB)                        <<< You will now get the bytes + string with unit-specifiers.
    message_bytes=262304591 (~250.15 MiB))        <<<<
...
`

File: src/btree.c [hunk 15]
```diff
@@ -3108,29 +3161,31 @@ btree_print_index_node(platform_log_handle *log_handle,
...
+                       leaf_entry          *entry,
+                       uint64               entry_num)
+{
+   data_config *dcfg = cfg->data_cfg;
+   platform_log(log_handle,
+                "[%2lu]: %s -- %s\n",
+                entry_num,
+                key_string(dcfg, leaf_entry_key(entry)),
+                message_string(dcfg, leaf_entry_message(entry)));
+}
```
gapisback, line 3181: Ignore this refactoring of BTree print code into sub-functions. That will be separately integrated to `/main` before this change-set under PR #521 .

tests/unit/misc_test.c
```diff
@@ -183,13 +190,102 @@ CTEST2(misc, test_ctest_assert_prints_user_msg_with_params)
    platform_close_log_stream(&stream, data->log_output);
 }
 
+/*
+ * Simple test to verify lower-level conversion macros used by size_to_str()
+ */
+CTEST2(misc, test_bytes_to_fractional_value_macros)
```
gapisback, line 196: Just a brain-dead, `make-sure-I-am-not-making-any-silly-mistakes-test`-test case.

tests/unit/misc_test.c
```diff
@@ -183,13 +190,102 @@ CTEST2(misc, test_ctest_assert_prints_user_msg_with_params)
...
+      size_t size;
+      char  *expstr;
+   } size_to_expstr;
+
+   // Verify base-case, and validate ptr returned to output string buffer.
+   size_t size   = 0;
+   char  *expstr = "0 bytes";
+   char  *outstr = size_to_str(size_str, sizeof(size_str), size);
+   // Conversion fn should return input-str's start addr as outstr.
+   ASSERT_TRUE(outstr == size_str);
```
gapisback, line 231: Validates that the returned string `char *` is, indeed, the start addr of `outbuf`; i.e. `size_str[]`.

tests/unit/btree_stress_test.c
```diff
@@ -255,6 +265,12 @@ CTEST2(btree_stress, test_random_inserts_concurrent)
       (cache *)&data->cc, &data->dbtree_cfg, packed_root_addr, nkvs, data->hid);
    ASSERT_NOT_EQUAL(0, rc, "Invalid ranges in packed tree\n");
 
+   // Exercise print method to verify that it basically continues to work.
+   btree_print_tree(Platform_default_log_handle,
```
gapisback, line 269: This call here is a bit out-of-line with the nature of this "stress" test, whose original intent was to do some stress-activity to exercise BTree interfaces.

The reason this is here is because in this test we are invoking `pack_tests()` -> `btree_pack()`, and it is through this interface that the fields in`btree_pivot_stats{}` are updated.

This change-set enhances `btree_print_btree_pivot_stats()` which prints these fields to now also print them using unit-specifiers.

I needed a way to exercise this print code to cross-check that the right values are being printed.

For that purpose, this small change leverages the work done in this BTree stress test to also invoke the print methods as a way to do this validation.

gapisback: As mentioned elsewhere, the outputs from the updated print methods looks like:

```
BTree node at addr=1333133312
{
**  INDEX NODE
**  Header ptr: 0x7f55c36ce000
**  addr: 1333133312
**  next_addr: 0
**  next_extent_addr: 0
[...]
Array of 3 index entries:
[ 0]: key=0x00000d093a9a17000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
   child_addr=1333657600
   (num_kvs=377607
    key_bytes=21716772 (~20.71 MiB)
    message_bytes=260863299 (~248.77 MiB))
[ 1]: key=0x60aaf36fb8dd4d000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
   child_addr=1333661696
   (num_kvs=379751
    key_bytes=21839065 (~20.82 MiB)
    message_bytes=262304591 (~250.15 MiB))

[... And so on ...]

```

src/btree.c
```diff
@@ -3083,13 +3083,66 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
...
+                size_to_str(nkbytes_str, sizeof(nkbytes_str), nkbytes),
+                nmbytes,
+                size_to_str(nmbytes_str, sizeof(nmbytes_str), nmbytes));
+}
+
+static void
+btree_print_btree_pivot_data(platform_log_handle *log_handle,
+                             btree_pivot_data    *pivot_data)
+{
+
```
gapisback, line 3120: [Will] delete this line ...

src/platform_linux/platform.h
gapisback: @ajhconway - In some other reviews you had said that we don't want to print decimal values using `%f` format. 

I forget now .. whether we are also not supposed to do division by float values, like is done in the body of the macros below.

I am hoping that this form of division to generate the decimal portion of the bytes value converted to a unit-specifier would be acceptable.

The meat of this change depends entirely on this small functionality.

rtjohnso:
Great idea.  Here's a way to make it even easier to use.  See `message_str` in `src/data_config.h`:

```
#define message_string(cfg, msg)                                               \
   (({                                                                         \
       struct {                                                                \
          char buffer[128];                                                    \
       } b;                                                                    \
       data_message_to_string((cfg), (msg), b.buffer, 128);                    \
       b;                                                                      \
    }).buffer)
```

So I think you could define something like 
```
#define size_str(sz)                                               \
   (({                                                                         \
       struct {                                                                \
          char buffer[128];                                                    \
       } b;                                                                    \
       size_to_str(b.buffer, 128, sz);                    \
       b;                                                                      \
    }).buffer)
```

You can then use `size_str` in `platform_log` parameter lists, e.g.
```
platform_default_log("Here's a pretty-print of 1 megabyte: %s\n", size_str(1<<20));
```

gapisback:
@rtjohnso - I applied your suggested addition, modified `util.h` to add two new caller macros: `size_str()` and `size_fmtstr()`. 

One thing I noted is that if we call, say, `size_str()` multiple times in a function across different `platform_log()` statements to print different values, then we end-up declaring multiple copies of the on-stack buffer `b.buffer` that is hidden behind the body of these macros. This needlessly bloats the size of the `.o` generated.

So, I only used these new convenience macros where it made sense -- i.e. , we were printing just one value in a statement in a function. And, in many other places, left the original implementation as-is, re-using on-stack char output buffer.

I think one has to use these caller macros judiciously, otherwise, it will needlessly bloat the `.o`s size. Have added some cautionary comments to `util.h` indicating how / when to use these wrapper caller-macros

Amended the change-set accordingly, made couple of minor corrections, and have [pushed a new commit](https://github.com/vmware/splinterdb/pull/512/commits/40dbee3655ed681caeff3fedd78a17e04bc05d3f). You can review the rest of the diffs in this PR.

FYI; @ajhconway - you were concerned about printing / using float values -- Kindly do re-look at this PR (if you have the time) and let me know if there are major issues that will prevent us from going down this path.

rtjohnso:
I would let the compiler optimize the redundant buffer decls.  Its liveness
analysis should handle that.  So please just use the macros.

rtjohnso:
OK bummer.  I wrote a quick example and neither gcc or clang is actually able to recognize that these buffers are dead and re-use the stack space.

But I still think it is ok for your application.  The buffers you need are like 8 bytes.  They're no worse than declaring another int on the stack.

gapisback:
@rtjohnso : About "_The buffers you need are like 8 bytes._ " ... It's a little bit more than that.

To print something like '~250.15 MiB' we will need 12 bytes and to print '(~250.15 MiB)' we will need 14 bytes.
If the compiler were to cleanly optimize this stack space usage, I agree, we can wholesale use the new `size_str()` and `size_fmstr()` everywhere. (Whenever possible, I am a bit sensitive to and careful about excessive stack space usage.)

Given that this is not happening, I have selectively used these new interfaces in few places, such as `trunk_print_space_use()`, `rc_allocator_print_stats()`, and `btree_print_btree_pivot_stats()`.

In other places where multiple instances are printed in a single function, like in `rc_allocator_print_stats()`, I am re-using on-stack buffers for multiple calls to `size_to_str()` or `size_to_fmtstr()` calls.

I think this is a reasonable compromise balancing the ease-of-use of the `size_str()` macro approach v/s the reuse of on-stack buffers that the fn-call-equivalent provides (with a little bit of coding overhead).

If you still very strongly feel that I must only use `size_str()` everywhere, then let's talk this through. Otherwise, kindly go through the rest of the review packet and let me know if you have any other feedback.

src/rc_allocator.c
rtjohnso: Ditto

src/rc_allocator.c
rtjohnso: Can you put the bytes part in parentheses, as above?

rtjohnso: Also, I don't really understand what this line is telling me.  Can you improve it?

src/rc_allocator.h
```diff
@@ -44,8 +44,8 @@ _Static_assert(offsetof(rc_allocator_meta_page, splinters) == 0,
  *----------------------------------------------------------------------
  */
 typedef struct rc_allocator_stats {
-   int64 curr_allocated;
-   int64 max_allocated;
+   int64 curr_allocated; // # of extents allocated
+   int64 max_allocated;  // # of extents allocated hwm
```
rtjohnso, line 48: I don't think "hwm" is as widely known as you think it is.  "High-water mark" is fine if you want.  But do you really think "hwm" is clearer than "max"?

src/util.c
```diff
@@ -364,3 +364,65 @@ debug_hex_dump_slice(platform_log_handle *plh, uint64 grouping, slice data)
...
+   } else if (size >= KiB) {
+      unit_val  = B_TO_KiB(size);
+      frac_val  = B_TO_KiB_FRACT(size);
+      is_approx = (size > KiB_TO_B(unit_val));
+      units     = "KiB";
+   } else {
+      unit_val = size;
+      units    = "bytes";
+   }
+   if (frac_val || is_approx) {
```
rtjohnso, line 409: This could be much much simpler and safer:

```C
if (frac_val || is_approx) {
   snprintf(outbuf, outbuflen, "~%ld.%d %s", unit_val, frac_val, units);
} else {
   snprintf(outbuf, outbuflen, "%ld %s", unit_val, units);
}
```

src/util.c
```diff
@@ -364,3 +364,65 @@ debug_hex_dump_slice(platform_log_handle *plh, uint64 grouping, slice data)
...
+   snprintf(outbuf, outbuflen, fmtstr, unit_val, frac_val);
+   return outbuf;
+}
+
+/*
+ * Sibling of size_to_str(), but uses user-provided print format specifier.
+ * 'fmtstr' is expected to have just one '%s', and whatever other text user
+ * wishes to print with the output string.
+ */
+char *
```
rtjohnso, line 423: I don't see a reason for this function to exist.  First, I don't like the call sites.  They are working too hard for a minor change in the output.

Second, if we do want to keep the unnecessarily complicated code, then call sites should just do the snprintf directly.

src/util.h
```diff
@@ -382,4 +382,49 @@ debug_hex_dump_slice(platform_log_handle *, uint64 grouping, slice data);
     : ((intval) < 1000) ? "3d"                                                 \
                         : "4d")
 
+// Size of on-stack char message buffer (mostly used in diagnostics)
+#define DIAG_MSGBUF_SIZE 80
```
rtjohnso, line 386: Seems to be unused.

src/util.h
```diff
@@ -382,4 +382,49 @@ debug_hex_dump_slice(platform_log_handle *, uint64 grouping, slice data);
...
+
+// Convenience specifier for output buffer to convert size with unit specifier
+#define SIZE_TO_STR_LEN 20
+
+// Min size of buffer needed if size is enclosed in (), as : '(~3.50 GiB)'
+#define SIZE_TO_STR_MINLEN (SIZE_TO_STR_LEN - 2)
```
rtjohnso, line 392: This is used only in one  debug_assert in size_to_str.

Furthermore, I don't think that debug_assert even makes sense.  Why is there room for parens required in these buffers?  size_to_str does not put parens in its output buffers.

src/util.h
```diff
@@ -382,4 +382,49 @@ debug_hex_dump_slice(platform_log_handle *, uint64 grouping, slice data);
...
+/*
+ * Convenience caller macros to convert 'sz' bytes to return a string,
+ * formatting the input size as human-readable value with unit-specifiers.
+ *
+ * NOTE: These macros are best used when you have just one instance of a size
+ * to print in a function. If in places, e.g. diagnostic code, you have to
+ * print multiple size values across different statements, it's better to
+ * re-use on-stack output buffers.
+ */
+// char *size_str(size_t sz)
```
rtjohnso, line 410: One thing I noticed in my testing is that the name of the temporary, "b", is just too short and likely to collide with local variables.  Give it some long crazy name.

rtjohnso (changes requested):
Mostly good, but a several strongly preferred changes a few required ones.

src/btree.c
```diff
@@ -3256,6 +3259,7 @@ btree_print_subtree(platform_log_handle *log_handle,
    node.addr = addr;
    btree_print_node(log_handle, cc, cfg, &node);
    if (!allocator_page_valid(cache_get_allocator(cc), node.addr)) {
+      platform_log(log_handle, "Invalid BTree node addr=%lu\n", addr);
```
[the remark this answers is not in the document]
gapisback: Done.

src/platform_linux/platform.h
gapisback: Yes! good catch! (Sorry, I was sloppy -- forgot my arithmetic.) Fixed in latest.

src/rc_allocator.c
gapisback: @rtjohnso - I can't tell from your comment if this item is (in your words) "a strongly preferred changes" or "a required change".

The 2-step process does not bother me. I am not very keen to use multiple instances of `size_str()` in one function due to the potential stack over-use issue we discussed previously.

What remains is formatting and indentation. The 'problem', I think, is also that we have carefully constructed these tabular outputs (which are nice to read!), so I have to work a bit extra hard to get the alignment right.

Also, I don't think `"(       ~1.6 GiB)"` reads very well. It spoils the aesthetics of the carefully constructed aligned tabular outputs.

One small thing I could do for just the two lines reporting currently allocated and max allocated extents is to drop the enclosing `( )`, but that would make the output of this size in unit-specifiers field inconsistent with the rest of the output generated in this function. So, I'm not so keen on that either.

Let's talk off-line ... to see how best to resolve this.

src/rc_allocator.c
gapisback: Changed the message to the following to address both comments:

`"Expected count of extents in-use from footprint = %ld extents (%s)\n"`

The code above is counting footprint as diff of (# of allocated-extents - # of deallocated-extents).
That is called the 'footprint'; in my words, currently in-use extents.
That's why (seems like) this message says "Expected count" ... as we are running off of stats being gathered (I'm guessing) during extent allocation / deallocation.

src/rc_allocator.h
gapisback: I think so, yes, Rob.

In this area of metrics and stats-gathering, for things like this one which go up and down, the terms 'HWM' and 'LWM' (low-water mark) are fairly common.

They are usually preferred over 'max' and 'min', which could be interpreted as the "max-value this thing can go to" or the "min-value this thing can drop to"; i.e. hard min/max limits.

I would have changed the name of the variable itself to `hwm_allocated` but refrained as it would be considered unacceptable. But, I think it's better to go with hwm / lwm terminology for such monitoring related metrics variables.

src/util.c
gapisback: Done.

src/util.h
gapisback: Done. Deleted.

src/util.h
gapisback: You are right. This is left-over cruft from some versions of code I wrote earlier, and is now no longer needed. 
Deleted this `SIZE_TO_STR_MINLEN` define.

I reworked the assertion in `size_to_str()` to check that the `outbuflen >= SIZE_TO_STR_LEN`.

src/util.h
gapisback: Renamed this to `onstack_chartmp`.

(Also, I'd like to learn from you offline how you figured out that there could be likely name collision. I thought `{ }` scoping would prevent that.)

src/util.c
gapisback: The problem as I explained previously, Rob, is that we want to see the formatted string enclosed in `( )` and the outer tabular output has all these columnar alignments. That's why the calls in the call-sites are working a bit harder to conform to this aligned output.

Earlier, in my dev, I actually had this as `size_to_str2()` which would only support one output formatting; i.e. enclosing the output in `(%s)`. That would work as well.

I just thought it would be better to make this formatting a bit more general, which is why I changed this function to `size_to_fmtstr()`, receiving a user-provided format string, which in all cases (currently) is simply `(%s)`. 
It also makes the call self-documenting in the caller's context to see that the output will be enclosed in `()`.

I have thought about this quite a bit and have fuddled with ways to try to get the printing to work even while retaining the aesthetics of the aligned columnar output. I couldn't think of any better scheme.

Honestly, I don't think the existence of this sibling printing routine is a huge annoyance. In fact, the work of `size_to_fmtstr()` has been even improved by the introduction of `size_str()` macro-ization you suggested.

So, I think it's an improvement overall! (I'd like to retain this.)

rtjohnso:
Since you and I disagreed on the above issues, I discussed it with Alex as a tie-breaker.

He votes:
- use `size_str` and don't worry about stack sizes.
- keep the code you have for properly aligning the numbers in parentheses.

gapisback:
HI, @rtjohnso and @ajhconway - Based on this above feedback, I have reworked the fix. See [this commit](https://github.com/vmware/splinterdb/pull/512/commits/ec4cfa94c30de1e077b45d5dc4fc28be591edfab). Briefly in this version:

- Always use either `size_str()` or `size_fmtstr()` in all places where we want to print a `size` bytes as a formatted string. (I, do, agree that this has greatly improved the usability of these printing macro interfaces. Code is much more readable!)
- All on-stack char buffers declared in prior versions are now withdrawn in favor of using the hidden on-stack buffer allocated by these caller-macros. 
- Minor rework to macro-ize printing BTree pivot-stats when an uninitialized value is found.
- Minor improvements for readability

Please do take a final look ... to see if this version meets all your asks.

rtjohnso (approved):
Thank you!

rosenhouse (commented):
Hi @gapisback I know you were requesting a review from me, but I think I need to "pass" here.    Function-like macros and string-mangling are the parts of C I feel least confident in.  Furthermore, the rationale for this (stats printing) is a feature I haven't needed to use before.  And I don't have the bandwidth right now to really dig in and learn this stuff sufficiently to provide a meaningful review.  Please move forward without my review.   Sorry.

Issue #513, opened 2022-12-23
Refactor code that manages stdout / stderr file handles in unit tests to a common routine.


gapisback:
Under PR #494 (at SHA 83de5e0), we introduced `platform_set_log_streams(stdout, stderr)` as the mechanism to keep execution of unit-tests silent, and to redirect outputs to `/dev/null` for normal execution. Output log-stream handling is supposed to be done in the `CTEST_SETUP()` method.

This code chunk is now appearing in multiple unit-test source files. Refactor the code chunk into a common function that can be called from all unit-test files.

As part of this issue, also clean-up the db-name defined for unit-tests, in `tests/unit/unit_tests.h`: `#define TEST_DB_NAME "unit_tests_db"`

Change this to conform to r.e. defined in `.gitignore` to suppress listing it in `git status` output.

PR #514, opened 2022-12-23
(#513) Refactor into set_log_streams_for_tests() to manage unit-test outputs.

gapisback:
This commit refactors existing chunks of code that exists in different unit- test sources to manage output file handles to a common function, now defined in new file `unit/unit_tests_common.c` ; `set_log_streams_for_tests()`.

Tests that check error-raising behaviour will now need to call `set_log_streams_for_neg_tests()`, to manage output streams.

Minor correction to `TEST_DB_NAME`; change it to conform to the r.e. defined in .gitignore to suppress listing this in 'git status' output.

----
NOTE: I did not try to make this even more generic beyond unit-tests to have `functional/ tests` also call into these interfaces. That's why this new file `unit_tests_common.c` is in `tests/unit` sub-dir and not in top-level `tests/` dir.

I think, let's see how this settles down and if we ever find the need to make functional tests also get this capability, we can consider relocating this new file to `tests/` sub-dir.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,12 @@ FUNCTIONAL_TESTSRC := $(call rwildcard, $(FUNCTIONAL_TESTSDIR), *.c)
 # Symbol for all unit-test sources, from which we will build standalone
 # unit-test binaries.
 UNIT_TESTSRC := $(call rwildcard, $(UNIT_TESTSDIR), *.c)
+
+# Specify the r.e. so it will only pick-up sources that are common to multiple
+# unit-tests (and won't pick-up test-specific common files, e.g.
+# btree_test_common.c)
+COMMON_UNIT_TESTSRC := $(wildcard $(UNIT_TESTSDIR)/*tests_common.c)
+
 TESTSRC := $(COMMON_TESTSRC) $(FUNCTIONAL_TESTSRC) $(UNIT_TESTSRC)
 
 # Some unit-tests will be excluded from the list of dot-oh's that are linked into
@@ -226,6 +232,9 @@ OBJ := $(SRC:%.c=$(OBJDIR)/%.o)
 # Objects from test sources in tests/ that are shared by functional/ and unit/ tests
 COMMON_TESTOBJ= $(COMMON_TESTSRC:%.c=$(OBJDIR)/%.o)
 
+# Objects from unit-test sources in tests/unit that are shared by unit tests
+COMMON_UNIT_TESTOBJ= $(COMMON_UNIT_TESTSRC:%.c=$(OBJDIR)/%.o)
+
 # Objects from test sources in tests/functional/ sub-dir
 FUNCTIONAL_TESTOBJ= $(FUNCTIONAL_TESTSRC:%.c=$(OBJDIR)/%.o)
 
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -386,7 +395,7 @@ BTREE_SYS = $(OBJDIR)/$(SRCDIR)/btree.o           \
 # Note each test bin/unit/<x> also depends on obj/unit/<x>.o, as
 # defined above using unit_test_self_dependency.
 #
-$(BINDIR)/$(UNITDIR)/misc_test: $(UTIL_SYS)
+$(BINDIR)/$(UNITDIR)/misc_test: $(UTIL_SYS) $(COMMON_UNIT_TESTOBJ)
```
gapisback, line 398: I thought about including this new `.o` in `COMMON_TESTOBJ` symbol, but that would mean we are mixing up shared `.o's` from diff sub-dirs. 

The way it's done is a bit more work to keep adding this `COMMON_UNIT_TESTOBJ` dependency in several lines, but it's at least clear, and cleaner, to specify exactly which `unit_test-dot-oh` depends on which common object.

tests/unit/splinter_test.c
```diff
@@ -115,9 +115,8 @@ CTEST_DATA(splinter)
 // clang-format off
 CTEST_SETUP(splinter)
 {
-   if (Ctest_verbose) {
-      platform_set_log_streams(stdout, stderr);
-   }
+   set_log_streams_for_tests();
```
gapisback, line 118: For symmetry, I introduced this `set_` method, too. This is really not needed, but it makes it easier for new unit-test dev to just follow one `set_log_streams_for*()` pattern.

tests/unit/unit_tests.h
```diff
@@ -5,9 +5,18 @@
  *  Define things in here that are going to be needed across most unit tests.
  */
 
+#include "splinterdb/public_platform.h"
+
 /* Name of SplinterDB device created for unit-tests */
-#define TEST_DB_NAME "unit_tests_db"
+#define TEST_DB_NAME "splinterdb_unit_tests_db"
```
gapisback, line 11: Slightly unrelated bug-fix ... arising from one of @rosenhouse 's suggestion on an earlier PR.

Makefile
```diff
@@ -41,7 +41,7 @@ UNIT_TESTSRC := $(call rwildcard, $(UNIT_TESTSDIR), *.c)
 # Specify the r.e. so it will only pick-up sources that are common to multiple
 # unit-tests (and won't pick-up test-specific common files, e.g.
 # btree_test_common.c)
-COMMON_UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*tests_common.c")
+COMMON_UNIT_TESTSRC := $(wildcard $(UNIT_TESTSDIR)/*tests_common.c)
```
gapisback, line 44: Comes about due to changes introduced under earlier PR integrated at SHA 615050a to remove use of `shell` built-ins in favor of using `Makefile` built-in commands.

gapisback:
Small reminder @rosenhouse - to see if you can spend some time on this review. I just rebased off of /main and pushed an updated commit set. Thanks!

tests/unit/unit_tests_common.c
```diff
@@ -0,0 +1,68 @@
...
+ * generate lots of "error" messages. By default, that would go to stderr, which
+ * would pollute test output. This function sets up output file handles such
+ * that those expected error messages are only printed when the caller sets the
+ * VERBOSE env var to opt-in. By default, execution from unit-tests will be
+ * silent, redirecting output handles to /dev/null.
+ *
+ * Returns: Output file handles (out/error) if caller has requested for them.
+ */
+void
+set_log_streams_for_neg_tests(platform_log_handle **log_stdout,
```
rosenhouse, line 41: I recognize this function may be difficult to name well, but I found "neg" confusing and inconsistent with our general style of using full descriptive names for functions.  Perhaps instead
```
set_log_streams_for_error_reporting_tests
```
or
```
set_log_streams_for_error_case_tests
```
or
```
set_log_streams_for_expected_error_tests
```

Since this function is always called unconditionally, on its own line, with short arguments, I think the longer name is fine.

gapisback: I settled on a slightly diff version of your 3rd option, i.e.,  `set_log_streams_for_error_tests()`. 

Confirmed off-line w/ you on slack.

tests/unit/unit_tests_common.c
```diff
@@ -0,0 +1,68 @@
...
+   if (Ctest_verbose) {
+      platform_set_log_streams(stdout, stderr);
+      if (log_stdout) {
+         *log_stdout = stdout;
+      }
+      if (log_stderr) {
+         *log_stderr = stderr;
+      }
+      CTEST_LOG_INFO("\nVerbose mode on.  This test exercises an error case, "
+                     "so on sucess it "
```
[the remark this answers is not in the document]
gapisback: Done.

PR #519, opened 2022-12-24
Move abstract base implementation into allocator.c

### LINKED ISSUE #516: add allocator.c
allocator.h is an interface over rc_allocator. The config module has grown to include some implementations that are more than just inline accessors, making  allocator.h more of an abstract base class. They should move into a new allocator.c file.

jonhnet:
fixes #516.


ajhconway (approved):
LGTM

Issue #520, opened 2022-12-28
Refactor inline BTree index_entry{}, leaf_entry{} and related sub-structure print code chunks into separate functions.


gapisback:
Under this issue, modularize existing code to print BTree `index_entry{}`, `leaf_entry{}`, `btree_pivot_data{}` and `btree_pivot_stats{}` structures into separate functions. This item continues the work done under PR #280 to refactor inline BTree-printing code to sub-functions. 

This change does not introduce any significant change in functionality but will now allow calling these print methods from other running code and from the debugger.

PR #521, opened 2022-12-28
(#520) Refactor few BTree print code chunks into sub-functions.

gapisback:
This commit adds following stand-alone fns to print diff BTree sub-structs:

- btree_print_btree_pivot_stats()
- btree_print_btree_pivot_data()
- btree_print_index_entry()
- btree_print_leaf_entry()

Existing print code inline in other methods is pulled-out into above functions to facilitate code-reuse and the ability to call from the debugger. No changes are done to existing functionality of print methods.

----
NOTE: This modularization sets up the stage to enhance the printing of BTree pivot stats in `btree_print_btree_pivot_stats()` to also print the `bytes` value as a string with unit-specifiers. This is an in-flight concurrent change under PR #512, also under review at this time. 

I'd like to integrate this change-set first and then lay down the above change-set.

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -3084,12 +3084,57 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
 }
 
 static void
+btree_print_btree_pivot_stats(platform_log_handle *log_handle,
+                              btree_pivot_stats   *pivot_stats)
+{
+   // If nothing has been updated, yet, print a msg accordingly.
+   if ((pivot_stats->num_kvs == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->key_bytes == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->message_bytes == BTREE_UNKNOWN_COUNTER))
+   {
+      platform_log(log_handle, "   (Pivot Stats=BTREE_UNKNOWN_COUNTER)\n");
+      return;
+   }
+
+   // Indentation is dictated by outer caller
+   platform_log(log_handle,
+                "   (num_kvs=%u\n"
+                "    key_bytes=%u\n"
+                "    message_bytes=%u)\n",
+                pivot_stats->num_kvs,
+                pivot_stats->key_bytes,
+                pivot_stats->message_bytes);
+}
+
+static void
+btree_print_btree_pivot_data(platform_log_handle *log_handle,
+                             btree_pivot_data    *pivot_data)
+{
+   // Indentation is dictated by outer caller
+   platform_log(log_handle, "   child_addr=%lu\n", pivot_data->child_addr);
+   btree_print_btree_pivot_stats(log_handle, &pivot_data->stats);
+}
+
+static void
+btree_print_index_entry(platform_log_handle *log_handle,
+                        btree_config        *cfg,
+                        index_entry         *entry,
+                        uint64               entry_num)
+{
+   data_config *dcfg = cfg->data_cfg;
+   platform_log(log_handle,
+                "[%2lu]: key=%s\n",
+                entry_num,
+                key_string(dcfg, index_entry_key(entry)));
+   btree_print_btree_pivot_data(log_handle, &entry->pivot_data);
+}
+
+static void
 btree_print_index_node(platform_log_handle *log_handle,
                        btree_config        *cfg,
                        uint64               addr,
                        btree_hdr           *hdr)
 {
-   data_config *dcfg = cfg->data_cfg;
    platform_log(log_handle, "**  INDEX NODE \n");
    platform_log(log_handle, "**  Header ptr: %p\n", hdr);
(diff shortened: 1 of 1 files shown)


src/btree.c
```diff
@@ -3083,13 +3083,56 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
...
+static void
+btree_print_btree_pivot_stats(platform_log_handle *log_handle,
+                              btree_pivot_stats   *pivot_stats)
+{
+   // If nothing has been updated, yet, print a msg accordingly.
+   if ((pivot_stats->num_kvs == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->num_kvs == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->num_kvs == BTREE_UNKNOWN_COUNTER))
+   {
+      platform_log(log_handle, "   (Stats=BTREE_UNKNOWN_COUNTER)\n");
```
gapisback, line 3095: While exercising different print methods, I found that many times these BTree stats are in this "uninitialized state" till the time BTree-Pack is done. So, I added this check to capture this case and [to] report L3095.

(Btw, this change sets up the stage that the changes under PR #512 will bring-in to print the `key_bytes` and `message_bytes` as values with unit-specifiers. If I did that for this uninitialized state, we will get a very large "(2 GiB)" reported in the outputs, which will be misleading.)

gapisback: The output when exercising these print methods thru the 

`VERBOSE=1 build/release/bin/unit/splinter_test test_splinter_print_diags` 

test is as follows:

```
    23 Memtable root_addr=48: gen 5615779840 ref_count 3 state 1
    24 *******************
    25 BTree node at addr=5615779840
    26 {
    27 **  INDEX NODE
    28 **  Header ptr: 0x7fc56e5a9000
    29 **  addr: 5615779840
[...]

    40 Array of 4 index entries:
    41 [ 0]: key=0x00001c2c38bd531100000000000000000000000000000000
    42    child_addr=5616173056
    43    (Stats=BTREE_UNKNOWN_COUNTER)
...
```

This test case simply does a bunch of inserts, and BTree-pack is not explicitly being invoked. Hence, these pivot-stats values are still in their uninitialized state.

src/btree.c
```diff
@@ -3083,13 +3083,59 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
...
+static void
+btree_print_btree_pivot_stats(platform_log_handle *log_handle,
+                              btree_pivot_stats   *pivot_stats)
+{
+   // If nothing has been updated, yet, print a msg accordingly.
+   if ((pivot_stats->num_kvs == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->key_bytes == BTREE_UNKNOWN_COUNTER)
+       && (pivot_stats->message_bytes == BTREE_UNKNOWN_COUNTER))
+   {
+      platform_log(log_handle, "   (Pivot Stats=BTREE_UNKNOWN_COUNTER)\n");
```
gapisback, line 3095: The only real change in output, other than modularizing inline code into fns, is this line.

In a follow-on PR #512 , I am updating lines 3105 and 3106 below to print the # of bytes also with unit-specifier strings.

For the default case when BTree stats are not gathered, all values are 'uninitialized' to this large `BTREE_UNKNOWN_COUNTER` value (i.e., 2 billion something). I did not want to continually misreport these values as '2.xx GiB' as that's misleading.

Hence, this check above to report when all values are still at their default uninitialized state.

src/btree.c
```diff
@@ -3083,13 +3083,59 @@ btree_print_offset_table(platform_log_handle *log_handle, btree_hdr *hdr)
...
+      return;
+   }
+
+   // Indentation is dictated by outer caller
```
rtjohnso, line 3099: I suspect this indentation comment is dead?

src/btree.c
rtjohnso: Would it be better to check each counter separately?

rtjohnso (approved):
Feel free to ignore my minor comments.

src/btree.c
gapisback: It's a little bit messy to do that right now. Also, I think, in places where these stats are accumulated, say in `btree_accumulate_pivot_stats()`, we will update all 3.

From running tests, I found that either all 3 are uninitialized or they will have some valid value.

So did it this way. In an upcoming change, I am going to print the values as a string. At that time, I will revisit if I can check each value and print `"BTREE_UNKNOWN_COUNTER"`, as appropriate.

src/btree.c
gapisback: It is still needed, just to indicate why we have that leading `"   (num_kvs=%u\n"`. 

This comes about as this code was modularized out of the caller, where some indentation is needed.

Issue #342, opened 2022-03-11
splinter_test with invalid --fanout 100 raises misleading error: Maximum key size supported with current config: 18446744073709551578B.


gapisback:
Encountered this on /main [SHA f60f2df] during manual testing. Tried out some random value for `--fanout`, to see if we are defensively checking for legal values. We run into an assertion, so there is no real damage done.

But in the error below, we are mis-reporting some very BIG key-size. Seems like some arithmetic is off.

```
Fusion-LocalVM:[151] $ bin/driver_test splinter_test --functionality 1000000 100 --page-size 4096 --fanout 100
bin/driver_test: splinterdb_build_version f60f2df3
Dispatch test splinter_test
Assertion failed at src/trunk.c:8377:trunk_config_init(): "trunk_node_min_size < cache_config_page_size(cache_cfg)".
Trunk node does not fit in page size as configured.
hdr: 1102B
pivots: 9328B (max_pivot=106 x 88B)
branches 192B (max_branches=24 x 8B).
Maximum key size supported with current config: 18446744073709551578B. <<<<<<<

Aborted (core dumped)
```

Side-request. I am finding outputs like "88B" and "8B" hard to parse the bytes. They both look as all 8s.
Why not stick with "88 bytes" for readability everywhere?

Similar, **incorrect message is reported for this usage** of `--max-branches-per-node 2000`:

```
Fusion-LocalVM:[155] $ bin/driver_test splinter_test --functionality 1000000 100 --page-size 4096 --max-branches-per-node 2000
bin/driver_test: splinterdb_build_version f60f2df3
Dispatch test splinter_test
Assertion failed at src/trunk.c:8377:trunk_config_init(): "trunk_node_min_size < cache_config_page_size(cache_cfg)".
Trunk node does not fit in page size as configured.
hdr: 1102B
pivots: 1232B (max_pivot=14 x 88B)
branches 16000B (max_branches=2000 x 8B).
Maximum key size supported with current config: 1317624576693538408B.
```

Under this issue, also chase down error-checking for legal values of `--fanout` and similar parameters to avoid awkward bugs and un-helpful error messages.

PR #522, opened 2022-12-30
(#342) Fix checks in trunk_config_init() to report correct arg-values.

gapisback:
This commit does a minor cleanup in `trunk_config_init()` to correctly report the value for 'Maximum key size' reported when the checks fail for an invalid configuration. We were overflowing the value for an uint64 variable, leading to bogus values reported in the error. Changed this function and other helper functions in test-code that parse command-line arguments to return platform_status. Adjust the handling of this failed return-value in diff test code. Add new unit-test to verify that invalid config params correctly fail to init the trunk-config.

--- NOTE: While at it, I changed the code in `trunk_config_init()` from a `platform_assert()` to a clean error message. 
This should improve the usability of the library when, say, invoking `driver_test` with bogus configs that fail to init the trunk-configuration.

### DIFF
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -8923,7 +8923,7 @@ trunk_validate_data_config(const data_config *cfg)
  *       This function calls btree_config_init
  *-----------------------------------------------------------------------------
  */
-void
+platform_status
 trunk_config_init(trunk_config        *trunk_cfg,
                   cache_config        *cache_cfg,
                   data_config         *data_cfg,
@@ -8944,6 +8944,7 @@ trunk_config_init(trunk_config        *trunk_cfg,
 {
    trunk_validate_data_config(data_cfg);
 
+   platform_status rc = STATUS_BAD_PARAM;
    uint64          trunk_pivot_size;
    uint64          bytes_for_branches;
    routing_config *filter_cfg = &trunk_cfg->filter_cfg;
@@ -8965,7 +8966,7 @@ trunk_config_init(trunk_config        *trunk_cfg,
    // Inline what we would get from trunk_pivot_size(trunk_handle *).
    trunk_pivot_size = data_cfg->max_key_size + sizeof(trunk_pivot_data);
 
-   // Setting hard limit and over overprovisioning
+   // Setting hard limit and check configuration for over-provisioning
    trunk_cfg->max_pivot_keys = trunk_cfg->fanout + TRUNK_EXTRA_PIVOT_KEYS;
    uint64 header_bytes       = sizeof(trunk_hdr);
 
@@ -8973,31 +8974,44 @@ trunk_config_init(trunk_config        *trunk_cfg,
                          * (data_cfg->max_key_size + sizeof(trunk_pivot_data)));
    uint64 branch_bytes =
       trunk_cfg->max_branches_per_node * sizeof(trunk_branch);
-   uint64 trunk_node_min_size = header_bytes + pivot_bytes + branch_bytes;
-   uint64 available_pivot_bytes =
-      cache_config_page_size(cache_cfg) - header_bytes - branch_bytes;
+   uint64 trunk_node_min_size   = header_bytes + pivot_bytes + branch_bytes;
+   uint64 page_size             = cache_config_page_size(cache_cfg);
+   uint64 available_pivot_bytes = page_size - header_bytes - branch_bytes;
    uint64 available_bytes_per_pivot =
       available_pivot_bytes / trunk_cfg->max_pivot_keys;
-   uint64 available_bytes_per_pivot_key =
-      available_bytes_per_pivot - sizeof(trunk_pivot_data);
-   platform_assert(trunk_node_min_size < cache_config_page_size(cache_cfg),
-                   "\nTrunk node does not fit in page size as configured.\n"
-                   "node->hdr: %luB\n"
-                   "pivots: %luB (max_pivot=%lu x %luB)\n"
-                   "branches %luB (max_branches=%lu x %luB).\n"
-                   "Maximum key size supported with current config: %luB.\n",
-                   header_bytes,
-                   pivot_bytes,
-                   trunk_cfg->max_pivot_keys,
-                   data_cfg->max_key_size + sizeof(trunk_pivot_data),
-                   branch_bytes,
-                   max_branches_per_node,
-                   sizeof(trunk_branch),
-                   available_bytes_per_pivot_key);
+
+   // Deal with mis-configurations where we don't have available bytes per
+   // pivot key
+   uint64 available_bytes_per_pivot_key = 0;
+   if (available_bytes_per_pivot > sizeof(trunk_pivot_data)) {
+      available_bytes_per_pivot_key =
+         available_bytes_per_pivot - sizeof(trunk_pivot_data);
+   }
+
+   if (trunk_node_min_size >= page_size) {
+      platform_error_log("Trunk node min size=%lu bytes "
+                         "does not fit in page size=%lu bytes as configured.\n"
+                         "node->hdr: %lu bytes, "
+                         "pivots: %lu bytes (max_pivot=%lu x %lu bytes),\n"
+                         "branches %lu bytes (max_branches=%lu x %lu bytes).\n"
+                         "Maximum key size supported with current "
+                         "configuration: %lu bytes.\n",
+                         trunk_node_min_size,
+                         page_size,
+                         header_bytes,
+                         pivot_bytes,
+                         trunk_cfg->max_pivot_keys,
+                         trunk_pivot_size,
+                         branch_bytes,
+                         max_branches_per_node,
+                         sizeof(trunk_branch),
+                         available_bytes_per_pivot_key);
+      return rc;
+   }
 
    // Space left for branches past end of pivot array of [max_pivot_keys]
-   bytes_for_branches = (trunk_page_size(trunk_cfg) - trunk_hdr_size()
-                         - trunk_cfg->max_pivot_keys * trunk_pivot_size);
+   bytes_for_branches = (page_size - trunk_hdr_size()
+                         - (trunk_cfg->max_pivot_keys * trunk_pivot_size));
 
    // Internally determined hard-limit, which effectively depends on the
    // - configured page size and trunk header size
@@ -9086,6 +9100,8 @@ trunk_config_init(trunk_config        *trunk_cfg,
       filter_cfg->index_size *= 2;
       filter_cfg->log_index_size++;
    }
+   // When everything succeeds, return success.
+   return STATUS_OK;
 }
 
 size_t
--- a/tests/functional/test.h
+++ b/tests/functional/test.h
@@ -203,7 +203,7 @@ generator_average_message_size(test_message_generator *gen)
  * input master configuration, master_cfg. A few command-line config parameters
  * may have been used to setup master_cfg beyond its initial defaults.
  */
-static inline void
+static inline platform_status
 test_config_init(trunk_config           *splinter_cfg,  // OUT
                  data_config           **data_cfg,      // OUT
                  shard_log_config       *log_cfg,       // OUT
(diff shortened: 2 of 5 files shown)


src/trunk.c
```diff
@@ -8964,39 +8965,52 @@ trunk_config_init(trunk_config        *trunk_cfg,
...
-                   data_cfg->max_key_size + sizeof(trunk_pivot_data),
-                   branch_bytes,
-                   max_branches_per_node,
-                   sizeof(trunk_branch),
-                   available_bytes_per_pivot_key);
+
+   // Deal with mis-configurations where we don't have available bytes per
+   // pivot key
+   uint64 available_bytes_per_pivot_key = 0;
+   if (available_bytes_per_pivot > sizeof(trunk_pivot_data)) {
```
gapisback, line 8985: The error in mis-reported 'max key size' value creeps in here, due to an overflow of the `uint64` variable.

src/trunk.c
```diff
@@ -8964,39 +8965,52 @@ trunk_config_init(trunk_config        *trunk_cfg,
...
+                         page_size,
+                         header_bytes,
+                         pivot_bytes,
+                         trunk_cfg->max_pivot_keys,
+                         trunk_pivot_size,
+                         branch_bytes,
+                         max_branches_per_node,
+                         sizeof(trunk_branch),
+                         available_bytes_per_pivot_key);
+      return rc;
```
gapisback, line 9008: Change in interface.

src/trunk.c
```diff
@@ -8964,39 +8965,52 @@ trunk_config_init(trunk_config        *trunk_cfg,
...
+   // Deal with mis-configurations where we don't have available bytes per
+   // pivot key
+   uint64 available_bytes_per_pivot_key = 0;
+   if (available_bytes_per_pivot > sizeof(trunk_pivot_data)) {
+      available_bytes_per_pivot_key =
+         available_bytes_per_pivot - sizeof(trunk_pivot_data);
+   }
+
+   if (trunk_node_min_size >= page_size) {
+      platform_error_log("Trunk node min size=%lu bytes "
```
gapisback, line 8991: Raise an error rather than an assertion.

tests/functional/test.h
```diff
@@ -307,18 +313,21 @@ test_parse_args_n(trunk_config           *splinter_cfg,            // OUT
 
    rc = config_parse(master_cfg, num_config, argc, argv);
    if (!SUCCESS(rc)) {
-      return rc;
+      goto out;
```
gapisback, line 316: This fixes a minor bug in old code where we return w/o freeing memory allocated for `master_cfg`.

tests/unit/limitations_test.c
```diff
@@ -401,6 +404,55 @@ CTEST2(limitations, test_disk_size_not_integral_multiple_of_extents)
...
 }
 
+/*
+ * **************************************************************************
+ * Test that an invalid configuration that makes trunk node configuration
+ * impractical fails correctly with an error message. We try out few diff
+ * config params that go into error checks in trunk_config_init().
+ * **************************************************************************
+ */
+CTEST2(limitations, test_trunk_config_init_fails_for_invalid_configs)
```
gapisback, line 414: These are really not "limitations" but checks for invalid configs; i.e., negative test cases.

I don't have a very good place to dump these into, so for now ... this file should suffice.

src/trunk.c
```diff
@@ -8964,39 +8965,52 @@ trunk_config_init(trunk_config        *trunk_cfg,
...
+                         trunk_node_min_size,
+                         page_size,
+                         header_bytes,
+                         pivot_bytes,
+                         trunk_cfg->max_pivot_keys,
+                         trunk_pivot_size,
+                         branch_bytes,
+                         max_branches_per_node,
+                         sizeof(trunk_branch),
+                         available_bytes_per_pivot_key);
```
gapisback, line 9007: An example of this new error message when running `driver_test splinter_test` is here:

```
sdb-fdb-build:[65] $ ./build/debug/bin/driver_test splinter_test --functionality 1000000 100 --page-size 4096 --fanout 100
./build/debug/bin/driver_test: splinterdb_build_version d852a1db-dirty
Dispatch test splinter_test
Trunk node min size=13580 bytes does not fit in page size=4096 bytes as configured. node->hdr: 1198 bytes, pivots: 12190 bytes (max_pivot=106 x 115 bytes), branches 192 bytes (max_branches=24 x 8 bytes), Maximum key size supported with current configuration: 0 bytes.
Failed to parse config arguments. See 'driver_test splinter_test --help' for usage information: Invalid argument
```

tests/unit/limitations_test.c
```diff
@@ -401,6 +404,55 @@ CTEST2(limitations, test_disk_size_not_integral_multiple_of_extents)
...
+      TYPED_ARRAY_MALLOC(data->hid, data->splinter_cfg, num_tables);
+
+   data->cache_cfg = TYPED_ARRAY_MALLOC(data->hid, data->cache_cfg, num_tables);
+
+   char *unit_test_argv0[] = {"--key-size", "1000"};
+   int   unit_test_argc    = ARRAY_SIZE(unit_test_argv0);
+
+   char **unit_test_argv = unit_test_argv0;
+   rc = parse_cmdline_args(data, unit_test_argc, unit_test_argv);
+   ASSERT_FALSE(SUCCESS(rc));
```
gapisback, line 430: Will fail with following message when run with `VERBOSE=1`:

```
Trunk node min size=16664 bytes does not fit in page size=4096 bytes as configured. node->hdr: 1198 bytes, pivots: 15274 bytes (max_pivot=14 x 1091 bytes), branches 192 bytes (max_branches=24 x 8 bytes), Maximum key size supported with current configuration: 102 bytes.
```

tests/unit/limitations_test.c
```diff
@@ -401,6 +404,55 @@ CTEST2(limitations, test_disk_size_not_integral_multiple_of_extents)
...
+
+   char *unit_test_argv1[] = {"--page-size", "4096", "--fanout", "100"};
+   unit_test_argc          = ARRAY_SIZE(unit_test_argv1);
+
+   unit_test_argv = unit_test_argv1;
+   rc             = parse_cmdline_args(data, unit_test_argc, unit_test_argv);
+   ASSERT_FALSE(SUCCESS(rc));
```
gapisback, line 437: Will fail as follows:

```
Trunk node min size=13580 bytes does not fit in page size=4096 bytes as configured. node->hdr: 1198 bytes, pivots: 12190 bytes (max_pivot=106 x 115 bytes), branches 192 bytes (max_branches=24 x 8 bytes), Maximum key size supported with current configuration: 0 bytes.
```

tests/unit/limitations_test.c
```diff
@@ -401,6 +404,55 @@ CTEST2(limitations, test_disk_size_not_integral_multiple_of_extents)
...
+
+   char *unit_test_argv2[] = {"--max-branches-per-node", "200"};
+   unit_test_argc          = ARRAY_SIZE(unit_test_argv2);
+
+   unit_test_argv = unit_test_argv2;
+   rc             = parse_cmdline_args(data, unit_test_argc, unit_test_argv);
+   ASSERT_FALSE(SUCCESS(rc));
```
gapisback, line 444: Will result in this error message:

```
Trunk node min size=4408 bytes does not fit in page size=4096 bytes as configured. node->hdr: 1198 bytes, pivots: 1610 bytes (max_pivot=14 x 115 bytes), branches 1600 bytes (max_branches=200 x 8 bytes), Maximum key size supported with current configuration: 1 bytes.
```

src/trunk.c
```diff
@@ -8964,39 +8965,52 @@ trunk_config_init(trunk_config        *trunk_cfg,
    // Inline what we would get from trunk_pivot_size(trunk_handle *).
    trunk_pivot_size = data_cfg->max_key_size + sizeof(trunk_pivot_data);
 
-   // Setting hard limit and over overprovisioning
+   // Setting hard limit and check configuration for over-provisioning
```
ajhconway, line 8968: I don't understand this change. It looks like we're setting the overprovision, not checking for it.

src/trunk.c
```diff
@@ -8964,39 +8965,52 @@ trunk_config_init(trunk_config        *trunk_cfg,
...
 
    uint64 pivot_bytes = (trunk_cfg->max_pivot_keys
                          * (data_cfg->max_key_size + sizeof(trunk_pivot_data)));
    uint64 branch_bytes =
       trunk_cfg->max_branches_per_node * sizeof(trunk_branch);
-   uint64 trunk_node_min_size = header_bytes + pivot_bytes + branch_bytes;
-   uint64 available_pivot_bytes =
-      cache_config_page_size(cache_cfg) - header_bytes - branch_bytes;
+   uint64 trunk_node_min_size   = header_bytes + pivot_bytes + branch_bytes;
+   uint64 page_size             = trunk_page_size(trunk_cfg);
```
ajhconway, line 8977: This redirects to the `cache_cfg`, right? It feels better to use the one from the `cache_cfg`, rather than creating an implicit ordering requirement.

src/trunk.c
ajhconway: This is nice. Can you reformat it so as to print multiple 80 char lines?

src/trunk.c
ajhconway: I see that there was one already below, but I still think this is better as `cache_page_size(cache_cfg)`.

tests/unit/limitations_test.c
```diff
@@ -457,3 +509,35 @@ create_default_cfg(splinterdb_config *out_cfg, data_config *default_data_cfg)
...
+/*
+ * Helper function to parse command-line arguments to setup the configuration
+ * for SplinterDB.
+ */
+static platform_status
+parse_cmdline_args(void *datap, int unit_test_argc, char **unit_test_argv)
+{
+   // Cast void * datap to ptr-to-CTEST_DATA() struct in use.
+   struct CTEST_IMPL_DATA_SNAME(limitations) *data =
+      (struct CTEST_IMPL_DATA_SNAME(limitations) *)datap;
```
ajhconway, line 522: Rant:

I know we use this construction across tests and is out of scope, but I hate this construction everywhere in the tests. I don't understand what is good or useful about it, and it all violates both the naming conventions and the underlying philosophy. We don't use `XXXp` for pointers, no one knows what a `struct CTEST_IMPL_DATA_SNAME` is supposed to be, it isn't a name that we would use, and a newcomer could not possibly infer what is happening here. I continue to hate it and I think we should change how this works.

ajhconway (approved):
I think this is mostly good stuff.

This approval is pending minor changes.

src/trunk.c
gapisback: I'm not exactly sure what you are asking. In the old code, like below, there were two instances of this call to `cache_config_page_size()` which I tried to factor out into a local variable `page_size`.

```
8979    uint64 available_pivot_bytes =
8980       cache_config_page_size(cache_cfg) - header_bytes - branch_bytes;
[...]
8985    platform_assert(trunk_node_min_size < cache_config_page_size(cache_cfg),
```

Are you suggesting that I should retain the call to `cache_config_page_size()` and not call `trunk_page_size()` [ even though it's internally calling `cache_config_page_size()`?

I'm assuming that it's ok to refactor it out to a local variable `page_size`.

tests/unit/limitations_test.c
gapisback: Brief reason: This is something we inherited from the CTests framework, which synthesizes structures to encapsulate data structures declared in each tests' `CTEST_SETUP()`, using the test's name [i.e., say here, `limitations`] as a sub-part of the generated structure's name.

The reason why this messy stuff is here is because we want to pass-down an argument to `<Generated-struct's-name> * ptr` to a sub-function which wants to extract out individual fields. Otherwise, the interface for these minions will needlessly grow, having to explicitly list out ptrs to individual `data->member` fields.

`CTEST_IMPL_DATA_SNAME()` is a macro defined in ctests `unit/ctest.h` file which evaluates to the generated name of a C-struct that is produced by the CTests machinery.

I understand your concern about names like `XXXp` for pointers, and that can be changed easily.

I suggest you & I talk off-line, converge on an acceptable solution. And then I will log a cleanup issue to implement the corrected approach in all unit-tests.

src/trunk.c
gapisback: I interpreted the comment as follows:

`Setting hard limit` applies to stuff like this we do further below in this code blow:

```
9015    // Internally determined hard-limit, which effectively depends on the
9016    // - configured page size and trunk header size
9017    // - user-specified configured key size
9018    // - user-specified fanout
9019    trunk_cfg->hard_max_branches_per_node =
9020       bytes_for_branches / sizeof(trunk_branch) - 1;
```

The `check configuration for over-provisioning` is referring to these config params which are not practically supportable for the given page-size. This is exactly what this change is fixing, i.e., error checks of this type:

```
8982    // Deal with mis-configurations where we don't have available bytes per
8983    // pivot key
8984    uint64 available_bytes_per_pivot_key = 0;
8985    if (available_bytes_per_pivot > sizeof(trunk_pivot_data)) {
8986       available_bytes_per_pivot_key =
8987          available_bytes_per_pivot - sizeof(trunk_pivot_data);
8988    }
8989
8990    if (trunk_node_min_size >= page_size) {
8991       platform_error_log("Trunk node min size=%lu bytes "
```

That's why I thought `check configuration for over-provisioning' was reasonably accurate.

I'm marking this as resolved; Do suggest a reword if you still think it's incorrect or could be improved.

src/trunk.c
gapisback: Changed as suggested.

src/trunk.c
gapisback: Done. Revert back to use of `cache_config_page_size(cache_cfg)`

gapisback:
@ajhconway - I pushed a [new commit](https://github.com/vmware/splinterdb/pull/522/commits/e98a0f60ae47836c34eb68174d7195ec255390ea) applying diffs to address your request for changes. All have been done, except one comment correction where I felt the new comment was reasonably accurate.

Do take a 2nd look, if you wish to, and let me know. I'll try to push this to `/main` on Friday, after other things in-flight have integrated.

Issue #290, opened 2022-02-25
Revisit & revise use of 'shell', 'find' in Makefile. Replace with Makefile primitives.


gapisback:
This issue was raised during review of PR #277. Current Makefile has constructs like these:

```
 26 COMMON_TESTSRC := $(shell find $(TESTS_DIR) -maxdepth 1 -name "*.c")
 27
 28 FUNCTIONAL_TESTSRC := $(shell find $(FUNCTIONAL_TESTSDIR) -name "*.c")
 29 UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c")
```

New one is creeping in to use `egrep -v -e` to filter out some files. The concern raised is that:

Is this going to be something that will blow up when BSD or Darwin or whatever doesn't like the particular flags or something. Can't we do this in make?

--
Upon initial investigation, found `Makefile` commands like `wildcard` and `filter-out` that seem to be able to do the job.
See this post: https://stackoverflow.com/questions/10276202/exclude-source-file-in-compilation-using-makefile

Finalize this direction, and replace Unix commands with `Makefile` primitives.

PR #523, opened 2023-01-02
(#290) Replace use of 'find', 'grep' with Makefile built-in 'wildcard' command.

gapisback:
This commit replaces all uses of shell 'find', 'grep' commands in Makefile with built-ins such as 'wildcard', 'filter-out' etc. This removes any potential dependency on shell commands for 'make' invocations. A small change is to list the unit-test sources in lex-sorted order when linking them into the unit_test binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,27 +21,41 @@ UNITDIR              = unit
 UNIT_TESTSDIR        = $(TESTS_DIR)/$(UNITDIR)
 EXAMPLES_DIR         = examples
 
-SRC := $(shell find $(SRCDIR) -name "*.c")
+# Define a recursive wildcard function to 'find' all files under a sub-dir
+# See https://stackoverflow.com/questions/2483182/recursive-wildcards-in-gnu-make/18258352#18258352
+define rwildcard =
+	$(foreach d,$(wildcard $(1:=/*)),$(call rwildcard,$d,$2) $(filter $(subst *,%,$2),$d))
+endef
 
-# Generate list of common test source files, only from tests/ dir. Hence '-maxdepth 1'.
+SRC := $(call rwildcard, $(SRCDIR), *.c)
+
+# Generate list of common test source files, only from tests/ dir.
 # These objects are shared between functional/ and unit/ test binaries.
-COMMON_TESTSRC := $(shell find $(TESTS_DIR) -maxdepth 1 -name "*.c")
-FUNCTIONAL_TESTSRC := $(shell find $(FUNCTIONAL_TESTSDIR) -name "*.c")
+COMMON_TESTSRC := $(wildcard $(TESTS_DIR)/*.c)
+FUNCTIONAL_TESTSRC := $(call rwildcard, $(FUNCTIONAL_TESTSDIR), *.c)
 
 # Symbol for all unit-test sources, from which we will build standalone
 # unit-test binaries.
-UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c")
+UNIT_TESTSRC := $(call rwildcard, $(UNIT_TESTSDIR), *.c)
 TESTSRC := $(COMMON_TESTSRC) $(FUNCTIONAL_TESTSRC) $(UNIT_TESTSRC)
 
 # Some unit-tests will be excluded from the list of dot-oh's that are linked into
 # bin/unit_test, for various reasons:
-#  - Slow unit-tests will be skipped, as we want the#    resulting unit_test to
-#    run as fast as it can.
+#  - Slow unit-tests will be skipped, as we want the resulting unit_test binary
+#    to run as fast as it can.
 #  - Skip tests that are to be invoked with specialized command-line arguments.
-# These skipped tests which will have to be run stand-alone.
-FAST_UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c" | egrep -v -e"splinter_test|config_parse_test")
+# These tests which are skipped will have to be run stand-alone.
+# Construct a list of fast unit-tests that will be linked into unit_test binary,
+# eliminating a sequence of slow-running unit-test programs.
+ALL_UNIT_TESTSRC := $(call rwildcard, $(UNIT_TESTSDIR), *.c)
+SLOW_UNIT_TESTSRC = splinter_test.c config_parse_test.c
+SLOW_UNIT_TESTSRC_FILTER := $(foreach slowf,$(SLOW_UNIT_TESTSRC), $(UNIT_TESTSDIR)/$(slowf))
+FAST_UNIT_TESTSRC := $(sort $(filter-out $(SLOW_UNIT_TESTSRC_FILTER), $(ALL_UNIT_TESTSRC)))
 
-EXAMPLES_SRC := $(shell find $(EXAMPLES_DIR) -name "*.c")
+# To examine constructed variable.
+# $(info $$FAST_UNIT_TESTSRC is [${FAST_UNIT_TESTSRC}])
+
+EXAMPLES_SRC := $(call rwildcard, $(EXAMPLES_DIR), *.c)
 
 #*************************************************************#
 # CFLAGS, LDFLAGS, ETC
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,27 +21,35 @@ UNITDIR              = unit
 UNIT_TESTSDIR        = $(TESTS_DIR)/$(UNITDIR)
 EXAMPLES_DIR         = examples
 
-SRC := $(shell find $(SRCDIR) -name "*.c")
+SRC := $(wildcard $(SRCDIR)/*.c $(SRCDIR)/platform*/*.c)
```
gapisback, line 24: This works for now, but 'assumes' that there is only one sub-dir under `src/` named `platform_<_something_>`.

This creates a small dependency in future if ever we introduce a different sub-dir under `src/`. 

There are some more evolved ways of writing our own recursive `wildcard` function in `Make` lingo, but I found that needlessly complicated to introduce at this time. (Make's `wildcard ` function does not natively do a recursive search thru sub-dirs.)

Makefile
```diff
@@ -21,27 +21,35 @@ UNITDIR              = unit
...
 
 # Generate list of common test source files, only from tests/ dir. Hence '-maxdepth 1'.
 # These objects are shared between functional/ and unit/ test binaries.
-COMMON_TESTSRC := $(shell find $(TESTS_DIR) -maxdepth 1 -name "*.c")
-FUNCTIONAL_TESTSRC := $(shell find $(FUNCTIONAL_TESTSDIR) -name "*.c")
+COMMON_TESTSRC := $(wildcard $(TESTS_DIR)/*.c)
```
gapisback, line 28: The fact that `wildcard` does not recurse under sub-dirs implicitly gives us the `-maxdepth 1` functionality of the `find` command. (That's good for us!)

Makefile
```diff
@@ -21,27 +21,35 @@ UNITDIR              = unit
...
 # Symbol for all unit-test sources, from which we will build standalone
 # unit-test binaries.
-UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c")
+UNIT_TESTSRC := $(wildcard $(UNIT_TESTSDIR)/*.c)
 TESTSRC := $(COMMON_TESTSRC) $(FUNCTIONAL_TESTSRC) $(UNIT_TESTSRC)
 
 # Some unit-tests will be excluded from the list of dot-oh's that are linked into
 # bin/unit_test, for various reasons:
-#  - Slow unit-tests will be skipped, as we want the#    resulting unit_test to
+#  - Slow unit-tests will be skipped, as we want the resulting unit_test to
```
gapisback, line 38: Fixes existing typo ...

gapisback:
This issue was opened based on a review comment by @ajhconway (in an earlier PR). Am assigning the review both to him and to @rtjohnso who has done substantial work / re-work in the affected `Makefile`.

rtjohnso:
Love the objective of reducing dependencies.

One issue, as you point out, is deep directories.  There is a way to define a recursive wildcard function!  See
https://stackoverflow.com/a/18258352

I propose we use that.

Makefile
gapisback: @rtjohnso suggests that we should fix this for good w/o relying on this dependency that `platform*` is the only sub-dir seen under `src/`.

I think you meant this stackoverflow [Recursive wildcards in make](https://stackoverflow.com/questions/2483182/recursive-wildcards-in-gnu-make/18258352#18258352) post, right, which shows how to write a recursive `wildcard` function in make?

Makefile
```diff
@@ -21,27 +21,41 @@ UNITDIR              = unit
 UNIT_TESTSDIR        = $(TESTS_DIR)/$(UNITDIR)
 EXAMPLES_DIR         = examples
 
-SRC := $(shell find $(SRCDIR) -name "*.c")
+# Define a recursive wildcard function to 'find' all files under a sub-dir
+# See https://stackoverflow.com/questions/2483182/recursive-wildcards-in-gnu-make/18258352#18258352
+define rwildcard =
```
gapisback, line 26: @rtjohnso - I implemented this recursive `wildcard` function based on above mentioned stack overflow post.

I think this is what you had wanted to see fixed to avoid issues in the future if we ever were to create new sub-dirs.

This user-defined fn is now used on L30 below.

ajhconway (approved):
LGTM

Makefile
rtjohnso: Cool, but update the comment to eliminate the reference to "-maxdepth 1".

Makefile
```diff
@@ -21,27 +21,41 @@ UNITDIR              = unit
...
+endef
+
+SRC := $(call rwildcard, $(SRCDIR), *.c)
 
 # Generate list of common test source files, only from tests/ dir. Hence '-maxdepth 1'.
 # These objects are shared between functional/ and unit/ test binaries.
-COMMON_TESTSRC := $(shell find $(TESTS_DIR) -maxdepth 1 -name "*.c")
-FUNCTIONAL_TESTSRC := $(shell find $(FUNCTIONAL_TESTSDIR) -name "*.c")
+COMMON_TESTSRC := $(wildcard $(TESTS_DIR)/*.c)
+FUNCTIONAL_TESTSRC := $(wildcard $(FUNCTIONAL_TESTSDIR)/*.c)
```
rtjohnso, line 35: rwildcard is the most equivalent to the old code.

Makefile
```diff
@@ -21,27 +21,41 @@ UNITDIR              = unit
...
 
 # Symbol for all unit-test sources, from which we will build standalone
 # unit-test binaries.
-UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c")
+UNIT_TESTSRC := $(wildcard $(UNIT_TESTSDIR)/*.c)
```
rtjohnso, line 39: rwildcard

Makefile
```diff
@@ -21,27 +21,41 @@ UNITDIR              = unit
...
-#    run as fast as it can.
+#  - Slow unit-tests will be skipped, as we want the resulting unit_test binary
+#    to run as fast as it can.
 #  - Skip tests that are to be invoked with specialized command-line arguments.
-# These skipped tests which will have to be run stand-alone.
-FAST_UNIT_TESTSRC := $(shell find $(UNIT_TESTSDIR) -name "*.c" | egrep -v -e"splinter_test|config_parse_test")
+# These tests which are skipped will have to be run stand-alone.
+# Construct a list of fast unit-tests that will be linked into unit_test binary,
+# eliminating a sequence of slow-running unit-test programs.
+ALL_UNIT_TESTSRC := $(wildcard $(UNIT_TESTSDIR)/*.c)
```
rtjohnso, line 50: rwildcard

Makefile
```diff
@@ -21,27 +21,41 @@ UNITDIR              = unit
...
+SLOW_UNIT_TESTSRC = splinter_test.c config_parse_test.c
+SLOW_UNIT_TESTSRC_FILTER := $(foreach slowf,$(SLOW_UNIT_TESTSRC), $(UNIT_TESTSDIR)/$(slowf))
+FAST_UNIT_TESTSRC := $(sort $(filter-out $(SLOW_UNIT_TESTSRC_FILTER), $(ALL_UNIT_TESTSRC)))
+
+# To examine constructed variable.
+# $(info $$FAST_UNIT_TESTSRC is [${FAST_UNIT_TESTSRC}])
 
-EXAMPLES_SRC := $(shell find $(EXAMPLES_DIR) -name "*.c")
+EXAMPLES_SRC := $(wildcard $(EXAMPLES_DIR)/*.c)
```
rtjohnso, line 58: rwildcard

rtjohnso (changes requested):
Only minor changes.  Feel free to merge w/o a second review.

Thank you for eliminating these external dependencies.

Makefile
gapisback: I don't quite agree. On L35, we are hard-wiring the list of functional test sources to _only_ those that are in `functional/ test sub-dir, which currently has no further sub-dirs. So the search output is closer to what `wildcard` will provide.

In any case, I agree that it's better to use `rwildcard` here, and in similar such cases, to allow for the possibility that we may create sub-dirs under `functional/`, `unit/` and `examples/` sub-dirs.

With that in mind, I have revised the change to use `rwildcard` more often, whenever it is possible to do so, w/o breaking existing functionality.

Makefile
gapisback: Changed.

Makefile
gapisback: Changed.

Issue #156, opened 2021-11-14
`splinter_node_print_branches` does not report space. [ Fn renamed to `trunk_node_print_branches()`.]


ajhconway:
The function `trunk_node_print_branches()` does seem to print `kv_bytes_in_branch` and `kv_bytes_in_branch`. Maybe we can enhance this to aggregate the total space consumed by all branches in one trunk node and report that also.
(Updated: 30.Dec.2022)

PR #524, opened 2023-01-02
(#156) Extend trunk_node_print_branches() to print size w/units

gapisback:
Preliminary changes; Actual fix requested by issue #156, yet, to be fleshed out.
This commit adds minor improvements to trunk_node_print_branches() to:

    - Print # of kv-bytes in unit-specifiers, also
    - Print sum of # of branches, num tuples & num of kv-bytes for
      each trunk node.

    - Extend unit-test test_splinter_print_diags() to now also call
      trunk_print_branches() so we have a basic exercise of these
      print methods.

---
NOTE: These changes are built on top of in-flight PRs for issue #520 and issue #511. Those changes bring in the ability to print size values in `bytes` as a string with unit-specifiers. The actual changes related to this PR are in `trunk.c`. In the diffs, kindly ignore other changes from `btree.c` and other files which are from the above two in-flight PRs / issues.

### DIFF
--- a/src/data_internal.h
+++ b/src/data_internal.h
@@ -604,7 +604,7 @@ static inline void
 data_key_to_string(const data_config *cfg, key k, char *str, size_t size)
 {
    if (key_is_negative_infinity(k)) {
-      snprintf(str, size, "(negaitive_infinity)");
+      snprintf(str, size, "(negative_infinity)");
    } else if (key_is_negative_infinity(k)) {
       snprintf(str, size, "(positive_infinity)");
    } else {
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -1719,6 +1719,10 @@ trunk_pivot_branch_count(trunk_handle     *spl,
       spl, node->hdr->end_branch, pdata->start_branch);
 }
 
+/*
+ * Returns the # tuples (num_tuples) and # KV-bytes (num_kv_bytes) used for
+ * a single BTree rooted at 'root_addr'.
+ */
 static inline void
 trunk_pivot_btree_tuple_counts(trunk_handle *spl,
                                trunk_node   *node,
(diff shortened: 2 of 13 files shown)


src/data_internal.h
```diff
@@ -604,7 +604,7 @@ static inline void
 data_key_to_string(const data_config *cfg, key k, char *str, size_t size)
 {
    if (key_is_negative_infinity(k)) {
-      snprintf(str, size, "(negaitive_infinity)");
+      snprintf(str, size, "(negative_infinity)");
```
gapisback, line 607: Minor correction of spelling.

src/trunk.c
```diff
@@ -8816,68 +8834,101 @@ trunk_node_print_branches(trunk_handle *spl, uint64 addr, void *arg)
    trunk_node           node;
    trunk_node_get(spl->cc, addr, &node);
 
-   platform_log(
-      log_handle,
-      "------------------------------------------------------------------\n");
+   // clang-format off
+   const char *dashes = "------------------------------------------------------------------";
+   const char *dashes_long = "-------------------------------------------------------------------------------------------------";
```
gapisback, line 8839: These are formatting diffs to account for longer field-widths for certain columns printed below.

src/trunk.c
```diff
@@ -8816,68 +8834,101 @@ trunk_node_print_branches(trunk_handle *spl, uint64 addr, void *arg)
...
       trunk_branch_count_num_tuples(
          spl, &node, branch_no, &num_tuples_in_branch, &kv_bytes_in_branch);
-      uint64 kib_in_branch = 0;
+
+      nkvbytes = kv_bytes_in_branch;
+
+      sum_ntuples += num_tuples_in_branch;
+      sum_nkvbytes += kv_bytes_in_branch;
+
+      uint64 kib_in_branch = 1;
```
gapisback, line 8889: This changes old L8853 from value `0` to `1`. Otherwise, `kib_in_branch` computed below on L8891 would always come out as 0.

This change does not really fix something; just makes some things come out as non-zero.

I think this area is where some more code / logic is needed as indicated by call to non-existent function `trunk_branch_extent_count()`. 

Need some guidance from @ajhconway to understand what was expected here (from the original issue itself) and how to fix this code to report correct space amplification values.

gapisback:
Hi, @ajhconway - This is a draft PR aired to get some attention to the issue and to understand exactly what was expected from this brief description: `splinter_node_print_branches` does not report space.

Based on further discussion code in `trunk_node_print_branches()` will need to be revised.

gapisback:
This draft PR has been split-up into other PRs being independently checked-in. The main work to update `trunk_print_branches()` logic is being reviewed under PR #532

PR #526, opened 2023-01-06
Rename extent_discard (from hard_evict_extent)

jonhnet:
per discussion with Alex & Rob when exploring cache.
(based on #525 )


ajhconway (approved):
Looks great

PR #527, opened 2023-01-11
Identify Memtable v/s Branch page types via BTree-print routines.

gapisback:
This commit extends BTree-printing routines to also report the page type, whether it's a branch or a memtable BTree. As the structures and print methods are shared between two types of objects, this extra information will help in diagnostics.

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -2585,7 +2585,8 @@ btree_iterator_print(iterator *itor)
    btree_print_node(Platform_default_log_handle,
                     btree_itor->cc,
                     btree_itor->cfg,
-                    &btree_itor->curr);
+                    &btree_itor->curr,
+                    btree_itor->page_type);
 }
 
 const static iterator_ops btree_iterator_ops = {
@@ -3174,9 +3175,11 @@ static void
 btree_print_index_node(platform_log_handle *log_handle,
                        btree_config        *cfg,
                        uint64               addr,
-                       btree_hdr           *hdr)
+                       btree_hdr           *hdr,
+                       page_type            type)
 {
-   platform_log(log_handle, "**  INDEX NODE \n");
+   platform_log(
+      log_handle, "**  Page type: %s, INDEX NODE \n", page_type_str[type]);
    platform_log(log_handle, "**  Header ptr: %p\n", hdr);
    platform_log(log_handle, "**  addr: %lu \n", addr);
    platform_log(log_handle, "**  next_addr: %lu \n", hdr->next_addr);
@@ -3217,9 +3220,11 @@ static void
 btree_print_leaf_node(platform_log_handle *log_handle,
                       btree_config        *cfg,
                       uint64               addr,
-                      btree_hdr           *hdr)
+                      btree_hdr           *hdr,
+                      page_type            type)
 {
-   platform_log(log_handle, "**  LEAF NODE \n");
+   platform_log(
+      log_handle, "**  Page type: %s, LEAF NODE \n", page_type_str[type]);
    platform_log(log_handle, "**  hdrptr: %p\n", hdr);
    platform_log(log_handle, "**  addr: %lu \n", addr);
    platform_log(log_handle, "**  next_addr: %lu \n", hdr->next_addr);
@@ -3255,14 +3260,15 @@ void
 btree_print_locked_node(platform_log_handle *log_handle,
                         btree_config        *cfg,
                         uint64               addr,
-                        btree_hdr           *hdr)
+                        btree_hdr           *hdr,
+                        page_type            type)
 {
    platform_log(log_handle, "*******************\n");
    platform_log(log_handle, "BTree node at addr=%lu\n{\n", addr);
    if (btree_height(hdr) > 0) {
-      btree_print_index_node(log_handle, cfg, addr, hdr);
+      btree_print_index_node(log_handle, cfg, addr, hdr, type);
    } else {
-      btree_print_leaf_node(log_handle, cfg, addr, hdr);
+      btree_print_leaf_node(log_handle, cfg, addr, hdr, type);
    }
    platform_log(log_handle, "} -- End BTree node at addr=%lu\n", addr);
 }
@@ -3272,7 +3278,8 @@ void
 btree_print_node(platform_log_handle *log_handle,
                  cache               *cc,
                  btree_config        *cfg,
-                 btree_node          *node)
+                 btree_node          *node,
+                 page_type            type)
 {
    if (!allocator_page_valid(cache_get_allocator(cc), node->addr)) {
       platform_log(log_handle, "*******************\n");
@@ -3281,8 +3288,8 @@ btree_print_node(platform_log_handle *log_handle,
       platform_log(log_handle, "-------------------\n");
       return;
    }
-   btree_node_get(cc, cfg, node, PAGE_TYPE_BRANCH);
-   btree_print_locked_node(log_handle, cfg, node->addr, node->hdr);
+   btree_node_get(cc, cfg, node, type);
+   btree_print_locked_node(log_handle, cfg, node->addr, node->hdr, type);
    btree_node_unget(cc, cfg, node);
 }
 
@@ -3290,24 +3297,31 @@ void
 btree_print_subtree(platform_log_handle *log_handle,
                     cache               *cc,
                     btree_config        *cfg,
-                    uint64               addr)
+                    uint64               addr,
+                    page_type            type)
 {
    btree_node node;
    node.addr = addr;
-   btree_print_node(log_handle, cc, cfg, &node);
    if (!allocator_page_valid(cache_get_allocator(cc), node.addr)) {
-      platform_log(log_handle, "Unallocated BTree node addr=%lu\n", addr);
+      platform_log(log_handle,
+                   "Unallocated %s BTree node addr=%lu\n",
+                   page_type_str[type],
+                   addr);
       return;
    }
-   btree_node_get(cc, cfg, &node, PAGE_TYPE_BRANCH);
+   // Print node's contents only if it's a validly allocated node.
+   btree_print_node(log_handle, cc, cfg, &node, type);
+
+   btree_node_get(cc, cfg, &node, type);
    table_index idx;
 
    if (node.hdr->height > 0) {
       int nentries = node.hdr->num_entries;
       platform_log(log_handle,
-                   "\n---- BTree sub-trees under addr=%lu"
+                   "\n---- Page type: %s, BTree sub-trees under addr=%lu"
                    " num_entries=%d"
                    ", height=%d {\n",
+                   page_type_str[type],
                    addr,
                    nentries,
                    node.hdr->height);
@@ -3315,8 +3329,11 @@ btree_print_subtree(platform_log_handle *log_handle,
       for (idx = 0; idx < nentries; idx++) {
          platform_log(
             log_handle, "\n-- Sub-tree index=%d of %d\n", idx, nentries);
-         btree_print_subtree(
-            log_handle, cc, cfg, btree_get_child_addr(cfg, node.hdr, idx));
+         btree_print_subtree(log_handle,
+                             cc,
+                             cfg,
+                             btree_get_child_addr(cfg, node.hdr, idx),
+                             type);
       }
       platform_log(log_handle,
                    "\n} -- End BTree sub-trees under"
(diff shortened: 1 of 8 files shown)


src/btree.c
```diff
@@ -3133,9 +3134,11 @@ static void
 btree_print_index_node(platform_log_handle *log_handle,
                        btree_config        *cfg,
                        uint64               addr,
-                       btree_hdr           *hdr)
+                       btree_hdr           *hdr,
+                       page_type            type)
 {
-   platform_log(log_handle, "**  INDEX NODE \n");
+   platform_log(
+      log_handle, "**  Page type: %s INDEX NODE \n", page_type_str[type]);
```
gapisback, line 3141: Output will look like:

```
Memtable root_addr=5615779840: gen 48 ref_count 3 state 1
*******************
BTree node at addr=5615779840
{
**  Page type: memtable INDEX NODE
**  Header ptr: 0x7f592f13a000
```

src/btree.c
```diff
@@ -2552,7 +2552,8 @@ btree_iterator_print(iterator *itor)
    btree_print_node(Platform_default_log_handle,
                     btree_itor->cc,
                     btree_itor->cfg,
-                    &btree_itor->curr);
+                    &btree_itor->curr,
+                    PAGE_TYPE_BRANCH);
```
gapisback, line 2556: I suppose BTree iterators could span both memtable and branch btrees. I had to make a choice, and went with this one. 

Not sure if this has to be made more 'clever' to figure out the page's type based on which sub-tree the iterator is working upon.

gapisback, line 2556: I suppose BTree iterators could span both memtable and branch btrees. I had to make a choice, and went with this one.

Not sure if this has to be made more 'clever' to figure out the page's type based on which sub-tree the iterator is working upon.

src/btree.c
```diff
@@ -3176,9 +3179,11 @@ static void
 btree_print_leaf_node(platform_log_handle *log_handle,
                       btree_config        *cfg,
                       uint64               addr,
-                      btree_hdr           *hdr)
+                      btree_hdr           *hdr,
+                      page_type            type)
 {
-   platform_log(log_handle, "**  LEAF NODE \n");
+   platform_log(
+      log_handle, "**  Page type: %s LEAF NODE \n", page_type_str[type]);
```
gapisback, line 3186: Output will look like:

```
BTree node at addr=5615910912
{
**  Page type: memtable LEAF NODE
```

src/btree.c
```diff
@@ -3295,7 +3323,20 @@ btree_print_tree(platform_log_handle *log_handle,
...
 {
-   btree_print_subtree(log_handle, cc, cfg, root_addr);
+   btree_print_subtree(log_handle, cc, cfg, root_addr, PAGE_TYPE_BRANCH);
+}
+
+/*
+ * Routine to print a BTree of page-type 'type' starting from root_addr
+ */
+static void
+btree_print_tree_as(platform_log_handle *log_handle,
```
gapisback, line 3333: Routine is very similar to `btree_print_tree()`, but just takes an extra `page_type` arg.

I kept the name short with `_as` suffix ... 

I thought about `btree_print_tree_of_type` or `_with_type` but neither was appealing beyond `_as()` variant, without adding too long a function name.

src/trunk.c
```diff
@@ -8818,11 +8823,16 @@ trunk_node_print_branches(trunk_handle *spl, uint64 addr, void *arg)
    platform_log(
       log_handle,
       "------------------------------------------------------------------\n");
-   platform_log(
-      log_handle, "| node %12lu height %u\n", addr, trunk_height(&node));
+   platform_log(log_handle,
+                "| Page type: %s node addr=%lu height=%u next_addr=%lu\n",
```
gapisback, line 8827: I think it's useful to see both current `addr` and `next_addr` for pages that are linked, so am printing that extra field here on the top.

src/trunk.c
```diff
@@ -8149,22 +8154,22 @@ trunk_print_memtable(platform_log_handle *log_handle, trunk_handle *spl)
...
 
    uint64 mt_gen_start = memtable_generation(spl->mt_ctxt);
    uint64 mt_gen_end   = memtable_generation_retired(spl->mt_ctxt);
    for (uint64 mt_gen = mt_gen_start; mt_gen != mt_gen_end; mt_gen--) {
       memtable *mt = trunk_get_memtable(spl, mt_gen);
       platform_log(log_handle,
                    "Memtable root_addr=%lu: gen %lu ref_count %u state %d\n",
-                   mt_gen,
                    mt->root_addr,
+                   mt_gen,
```
gapisback, line 8166: Bug fix. Old code was mixing up args.

src/btree.c
```diff
@@ -3246,42 +3253,52 @@ btree_print_node(platform_log_handle *log_handle,
       platform_log(log_handle, "-------------------\n");
       return;
    }
-   btree_node_get(cc, cfg, node, PAGE_TYPE_BRANCH);
-   btree_print_locked_node(log_handle, cfg, node->addr, node->hdr);
+   btree_node_get(cc, cfg, node, type);
```
gapisback, line 3256: Use the passed-in `type` argument in preference to hard-coding as `PAGE_TYPE_BRANCH` as done on old L3249.

src/btree.c
```diff
@@ -3246,42 +3253,52 @@ btree_print_node(platform_log_handle *log_handle,
...
-      platform_log(log_handle, "Unallocated BTree node addr=%lu\n", addr);
+      platform_log(log_handle,
+                   "Unallocated %s BTree node addr=%lu\n",
+                   page_type_str[type],
+                   addr);
       return;
    }
-   btree_node_get(cc, cfg, &node, PAGE_TYPE_BRANCH);
+   // Print node's contents only if it's a validly allocated node.
+   btree_print_node(log_handle, cc, cfg, &node, type);
```
gapisback, line 3278: Relocated this here, rather than old L3262, as this call to `btree_print_node()` also makes an "is-allocated" check, as is being done on L3270 above.

So, now, we will only print the node's contents if it successfully passes the "is-allocated" test done in this function.

And, for same reason given previously, check using `type` arg rather than relying on hard-coded `PAGE_TYPE_BRANCH`.

src/data_internal.h
```diff
@@ -604,7 +604,7 @@ static inline void
 data_key_to_string(const data_config *cfg, key k, char *str, size_t size)
 {
    if (key_is_negative_infinity(k)) {
-      snprintf(str, size, "(negaitive_infinity)");
+      snprintf(str, size, "(negative_infinity)");
```
gapisback, line 607: fixes typo.

src/trunk.c
```diff
@@ -7958,13 +7966,16 @@ trunk_print_pivots(platform_log_handle *log_handle,
...
+      char buf[PIVOT_KEY_OUTBUF_LEN + 1];
+      snprintf(buf, sizeof(buf), "%-24s", key_string(spl->cfg.data_cfg, pivot));
+
       trunk_pivot_data *pdata = trunk_get_pivot_data(spl, node, pivot_no);
       if (pivot_no == trunk_num_pivot_keys(spl, node) - 1) {
          platform_log(log_handle,
-                      "| %.*s | %12s | %12s | %11s | %9s | %5s | %5s |\n",
-                      24,
-                      key_string(spl->cfg.data_cfg, pivot),
+                      "| %s | %12s | %12s | %11s | %9s | %5s | %5s |\n",
```
gapisback, line 7977: This change, and the introduction of `PIVOT_KEY_OUTBUF_LEN` and `char buf` on L7971 [etc.] is needed because the +/-ve infinity key length is not quite 24 bytes. So the outputs come out mis-aligned (slightly) with the older code. 

The outputs will be correctly aligned with this new code, as seen below:

Generated using: `./build/release/bin/unit/splinter_test test_splinter_print_diags`

```
Page type: trunk, Node addr=393216
{
----------------------------------------------------------------------------------------------------
|          |     addr     |   next addr  | height |   gen   | pvt gen |                            |
|  HEADER  |--------------|--------------|--------|---------|---------|----------------------------|
|          |       393216 |            0 |      1 |       0 |       8 |                            |
|--------------------------------------------------------------------------------------------------|
|                                       PIVOTS                                                     |
|--------------------------------------------------------------------------------------------------|
|         pivot key        |  child addr  |  filter addr | tuple count | kv bytes  |  srq  |  gen  |
|--------------------------|--------------|--------------|-------------|-----------|-------|-------|
| (negative_infinity)      |       524288 |   6200492032 |      149741 |  11841707 |    -1 |     6 |
| 0x20001c5108e65507000000 |       548864 |   6201278464 |      149188 |  11782153 |    -1 |     2 |
| Full key: 20001c51 08ffffffe65507 00000000 00000000 00000000 00000000
| 0x3ffc153e7bc3f0e2000000 |       532480 |   6202064896 |       37150 |   2937469 |    -1 |     7 |
```

src/trunk.c
```diff
@@ -7967,14 +7968,25 @@ trunk_print_pivots(platform_log_handle *log_handle,
...
-      snprintf(buf, sizeof(buf), "%-24s", key_string(spl->cfg.data_cfg, pivot));
+      memset(buf, ' ', sizeof(buf));
+      int nbytes = snprintf(buf,
+                            sizeof(buf),
+                            "%.*s",
+                            PIVOT_KEY_OUTBUF_LEN,
+                            key_string(spl->cfg.data_cfg, pivot));
+      if (nbytes <= PIVOT_KEY_OUTBUF_LEN) {
+         // Overwrite null-termination byte to blank-pad the rest of the buffer
+         buf[nbytes] = ' ';
```
gapisback, line 7982: All this jugglery is needed because we want the pivot-key output to be exactly 24-bytes.

The `key_string()` method generates shorter-strings for some cases: `"(negative_infinity)"` and 
`"(positive_infinity)"`.

So, what's happening here is that when the `snprintf()`-ed string is shorter than the requested 24 bytes, we _know_ that it's one of these two shorter-lengths. So, trim out the `\0` replacing it with a blank to get blank padded `buf`.

Then, the old code on lines L7989 and L8000 will ensure that we will print just the leading 24-bytes.

These generated strings are thereby blank-extended in the output to get the desired well-aligned output.

src/btree.c
gapisback: (Rob) Page-type should be in the iterator, try to get it from there rather than printing something misleading.

rtjohnso: The page type is in the iterator.

src/btree.c
rtjohnso: I would delete the _as suffix.

src/btree.c
```diff
@@ -3337,7 +3366,20 @@ btree_print_tree(platform_log_handle *log_handle,
                  btree_config        *cfg,
                  uint64               root_addr)
 {
-   btree_print_subtree(log_handle, cc, cfg, root_addr);
+   btree_print_subtree(log_handle, cc, cfg, root_addr, PAGE_TYPE_BRANCH);
```
rtjohnso, line 3369: btree_print_tree should take a page type

src/btree.c
gapisback: (robj) - Avoid the `_as()` suffix. See if the top-level `btree_print_tree()` itself could take a page-type and just use that.

Just have `btree_print_tree_as()` replace `btree_print_tree()` and the latter takes a page-type argument.

src/trunk.c
rtjohnso: Use printf's field width instead.

gapisback: (robj) On L 7988 below, it seems like if you change the print fmt specifier to `%*.*s`, and pass 24 twice to specify min and field-width ... it may work.

If so, the jugglery about could all go away.

rtjohnso (changes requested):

src/btree.c
gapisback: Made this change as discussed / suggested.

src/btree.c
gapisback: Done.

src/btree.c
gapisback: Yes, used that, and changed code accordingly.

src/trunk.c
gapisback: Thanks, @rtjohnso - for this tip. Now, using `%*.*s` as suggested, prints the field's output correctly aligned.

Got rid of some crufty code.

gapisback:
@rtjohnso - I have updated this BTree print page-type PR to make the changes we discussed this afternoon.
Please take a final look at this [2nd commit](https://github.com/vmware/splinterdb/pull/527/commits/176e4a4fb3b0e88df77b3542a2d2b5c1b5a0ffe5) 

With these corrections, this change set should be ready to go.

Thanks.

rtjohnso (approved):
ship it!

Issue #500, opened 2022-12-11
Improve the handling of 'hook'-related globals in task.c: hook_init_done, num_hooks, hooks[].


gapisback:
During test-dev and debugging of failing new test cases developed as part of PR #469 , the following code fragments were recognized as something that could be improved. In `task.c` we have the following globals:

```
 11 int              hook_init_done = 0;
 12 static int       num_hooks      = 0;
 13 static task_hook hooks[MAX_HOOKS];
```
These get initialized when `task_system_create()` calls `register_init_tid_hook()`.

In some unit test cases developed under above-mentioned PR, we were doing `task_system_destroy()` followed by a re-creation of the task system using `task_system_create()`.  `task_system_destroy()` destroys the task-system and releases its memory, but does not reset these global variables, especially the `hooks[]` array of functions.

At one point of test-dev, it appeared(*) that we were not invoking the thread-specific hook functions when new threads were created. Suspects were these globals, which was, eventually, found to be not the root cause.

The point of this issue is that we should still clean-up use of these globals and make them part of the `task_system{}` structure itself, so that these variable get setup and (properly) destroyed when the task-system is created / destroyed, respectively.

(*) It was later discovered to be a flaw in the test code due to a missing call to `platform_thread_join()` in the test case, which caused code-flow in the main thread to execute `CTEST_TEARDOWN()` which calls `task_system_destroy()`.

--- Updated 30.Dec.2022: Although slightly unrelated, plan to do this cleanup after the rework of fg-/bg-threads that is in-flight under PR #497 hits /main. That may affect some of this area, so want to wait till those changes are finalized. Some of this hooks-business may be slightly reworked with that upcoming change.

PR #528, opened 2023-01-11
(#500) Move hook-related global vars to task_system{} struct.

gapisback:
This commit removes the dependency of task system structures on global variables declared in `task.c` . The hook-related variables are now moved as members of the `task_system{}` struct. This removes accessing potentially stale values when task-system is destroyed and re-created. Also, `TASK_MAX_HOOKS` is now decreased from 8 to 4. This change largely has no functional impact, and is mostly a test-stabilization fix.

---
NOTE: The issue #500 describes in fuller detail why these global variables could present a problem in a testing scenario, e.g., when we do re-create a task-system by performing `task_system_destroy()` followed by `task_system_create()`.

### DIFF
--- a/src/task.c
+++ b/src/task.c
@@ -6,12 +6,6 @@
 
 #include "poison.h"
 
-#define MAX_HOOKS (8)
-
-int              hook_init_done = 0;
-static int       num_hooks      = 0;
-static task_hook hooks[MAX_HOOKS];
-
 const char *task_type_name[] = {"TASK_TYPE_INVALID",
                                 "TASK_TYPE_MEMTABLE",
                                 "TASK_TYPE_NORMAL"};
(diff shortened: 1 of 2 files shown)


src/task.c
```diff
@@ -6,12 +6,6 @@
 
 #include "poison.h"
 
-#define MAX_HOOKS (8)
-
-int              hook_init_done = 0;
-static int       num_hooks      = 0;
-static task_hook hooks[MAX_HOOKS];
-
```
gapisback, line 14: This block moves to inside the `task_system{}` struct.

rtjohnso:
How about we kill the hook stuff altogether?  It's used only for some IO registration thing, which we don't even use in our current platform.

gapisback:
@rtjohnso - You are right that we currently "... _It's used only for some IO registration thing, which we don't even use in our current platform._"

But, I'd like to retain this one last remaining hook for now. I discovered that the availability of `struct io_ops{}->io_thread_register_fn` to be handy and useful when integrating SplinterDB with the other fast-data ingestion product integration.

It was handy for me to poke a Unix process to become a thread, and to use the IO-Ops virtual functions as a way for the process to register itself as a thread and deregister (upon disconnection).  I actually extended this `io_ops` virtual function list to add a `deregister` method, too.

You don't see that code in our `/main` branch but having these virtual pointers and this hook-system will come of use in future integrations.

gapisback:
Forgot to cc: @ajhconway -- who may have another opinion on whether we should totally nuke this hooks-business it or if he would go along with my requesting to retain it (to facilitate future integration).

rtjohnso:
If you found it useful then no problem keeping it.

Best,
Rob

rtjohnso (approved):

Issue #499, opened 2022-12-11
Potential use of uninitialized variable, thread_tid, in task_register_thread() in assert code.


gapisback:
During the review of PR #469 it was noted that we are probably referencing an uninitialized variable in this code block at the top of `task_register_thread()`:

```
186    /*
187     * For all newly created threads, the thread-index will be 0 before the hook
188     * function, run below, generates this thread's index. We do not want to
189     * register a thread twice; otherwise, it will simply burn up an index and
190     * possibly also waste allocated scratch space.
191     * As threads generally will require scratch space to be allocated, an
192     * attempt to re-register a thread should find scratch space already
193     * allocated. Flag that.
194     */
195    platform_assert((thread_tid == 0)
196                       || (ts->thread_scratch[thread_tid] == NULL),
197                    "[%s:%d::%s()] Thread at index %lu is found to have scratch"
198                    " space already allocated, %p\n",
```

This assert was added previously under SHA bda1e46  as part of PR #448 .

`platform_get_tid()` returns `extern __thread threadid xxxtid` declared in `platform_linux/platform.c`: ` 10 __thread threadid xxxtid;`

While implementing the fix for PR #448, the thought was that this thread-local variable would be initialized to 0 but documentation suggests that this is not guaranteed.

It seems that we just got lucky that this assert on L195 is not tripping more often. Need to make this check more reliable and robust.

--- Updated 30.Dec.2022: Although slightly unrelated, plan to do this cleanup after the rework of fg-/bg-threads that is in-flight under PR #497 hits /main. That may affect some of this area, so want to wait till those changes are finalized.

PR #529, opened 2023-01-11
(#499) Minor cleanup of INVALID_TID, and MAX_THREADS in task-system

gapisback:
This commit applies some minor clean-up to task system as a follow-on to the larger rework done under PR #497. Consistently use STATUS_BUSY, as a way to report when all concurrent threads are in-use. This will result in a `Resource temporarily unavailable` message to the client. Minor changes to task system unit-test code & cleanup of comments.

--- NOTE to the reviewer:

Nothing is changing substantially; minor corrections to keep code & test code consistent.

### DIFF
--- a/src/task.c
+++ b/src/task.c
@@ -255,7 +255,7 @@ task_create_thread_with_hooks(platform_thread       *thread,
       platform_error_log("Cannot create a new thread as the limit on"
                          " concurrent threads, %d, will be exceeded.\n",
                          MAX_THREADS);
-      return STATUS_LIMIT_EXCEEDED;
+      return STATUS_BUSY;
    }
 
    if (0 < scratch_size) {
@@ -359,8 +359,9 @@ task_register_thread(task_system *ts,
                    thread_tid);
 
    thread_tid = task_allocate_threadid(ts);
+   // Unavailable threads is a temporary state that could go away.
    if (thread_tid == INVALID_TID) {
-      return STATUS_NO_SPACE;
+      return STATUS_BUSY;
    }
 
    platform_assert(ts->thread_scratch[thread_tid] == NULL,
--- a/tests/unit/task_system_test.c
+++ b/tests/unit/task_system_test.c
@@ -322,7 +322,7 @@ CTEST2(task_system, test_one_thread_using_extern_apis)
  * Background threads are off, by default.
  * ------------------------------------------------------------------------
  */
-CTEST2(task_system, test_multiple_threads)
+CTEST2(task_system, test_max_threads_using_lower_apis)
 {
    platform_thread new_thread;
    thread_config   thread_cfg[MAX_THREADS];
@@ -399,7 +399,7 @@ CTEST2(task_system, test_task_system_creation_with_bg_threads)
 /*
  * ------------------------------------------------------------------------
  * Test creation of task system using up the threads for background threads.
- * Verify ththe we can create just one more user-thread and that the next
+ * Verify that we can create just one more user-thread and that the next
  * user-thread creation should fail with a proper error message.
  * ------------------------------------------------------------------------
  */
@@ -410,7 +410,7 @@ CTEST2(task_system, test_use_all_but_one_threads_for_bg_threads)
    // Destroy the task system setup by the harness, by default, w/o bg threads.
    task_system_destroy(data->hid, &data->tasks);
 
-   // Consume all available threads with background threads.
+   // Consume all-but-one available threads with background threads.
    rc = create_task_system_with_bg_threads(data, 1, (MAX_THREADS - 3));
    ASSERT_TRUE(SUCCESS(rc));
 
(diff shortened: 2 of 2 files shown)


src/task.c
```diff
@@ -359,8 +359,9 @@ task_register_thread(task_system *ts,
                    thread_tid);
 
    thread_tid = task_allocate_threadid(ts);
+   // Unavailable threads is a temporary state that could go away.
    if (thread_tid == INVALID_TID) {
-      return STATUS_NO_SPACE;
+      return STATUS_BUSY;
```
gapisback, line 364: This change and L258 above, make the return status to be consistent. (This is how I had it with the previous commit of PR #469 @ SHA 5e8ec1b). The most recent change to `task.c` has introduced this inconsistency which I am reverting back to the way it was previously.

I think this is a better status as seen in the output from this unit-test:

```
TEST 5/6 task_system:test_use_all_but_one_threads_for_bg_threads
Verbose mode on.  This test exercises an error case, so on sucess it will print a message that appears to be an error.
Splinter task system created 1 background thread of type 'TASK_TYPE_MEMTABLE'.
Splinter task system created 61 background threads of type 'TASK_TYPE_NORMAL'.
Cannot create a new thread as the limit on concurrent threads, 64, will be exceeded.
Could not create a thread: Resource temporarily unavailable
```

See the bit above about `Resource temporarily unavailable`.

That comes out of `STATUS_BUSY`, which seems to be a more appropriate error return status in this case of (what usually could be) a transient error.

tests/unit/task_system_test.c
```diff
@@ -692,6 +692,9 @@ exec_one_of_n_threads(void *arg)
 {
    thread_config *thread_cfg = (thread_config *)arg;
 
+   // Before registration, thread ID should be in an uninit'ed state
+   ASSERT_EQUAL(INVALID_TID, platform_get_tid());
```
gapisback, line 696: The actual point of the original issue #499 was the abuse of potentially uninitialized return from `platform_get_tid()`.

Previous clean-up of `task.c` has tightened this up by always initializing the global variable, as seen in `platform.c`:

```
 10 __thread threadid xxxtid = INVALID_TID;
```

All this line is doing here is leveraging an existing test case to validate that the return from `platform_get_tid()` is no longer an uninitialized variable.

tests/unit/task_system_test.c
```diff
@@ -710,15 +713,14 @@ exec_one_of_n_threads(void *arg)
...
+   // have re-init'ed the thread ID.
    threadid get_tid_after_deregister = platform_get_tid();
    ASSERT_EQUAL(INVALID_TID,
                 get_tid_after_deregister,
-                "get_tid_after_deregister=%lu is != the index into"
-                " thread array, %lu ",
+                "get_tid_after_deregister=%lu should be an invalid tid, %lu",
                 get_tid_after_deregister,
-                this_threads_index);
+                INVALID_TID);
```
gapisback, line 723: Minor correction in comments and messages reported by failing assertion.

src/task.c
```diff
@@ -255,7 +255,7 @@ task_create_thread_with_hooks(platform_thread       *thread,
       platform_error_log("Cannot create a new thread as the limit on"
                          " concurrent threads, %d, will be exceeded.\n",
                          MAX_THREADS);
-      return STATUS_LIMIT_EXCEEDED;
+      return STATUS_BUSY;
```
rtjohnso, line 258: Thank you for fixing the inconsistency.

But don't you think LIMIT_EXCEEDED is the most appropriate error code here?

rtjohnso (approved):
I really do think LIMIT_EXCEEDED is the right return code, but feel free to do it either way.  No further review needed.

gapisback:
@rtjohnso - I had tried to implement 'LIMIT_EXCEEDED' in earlier rounds of my own dev. The error message reported by `strerror()` is very strange:

```
TEST 5/6 task_system:test_use_all_but_one_threads_for_bg_threads
Verbose mode on.  This test exercises an error case, so on sucess it will print a message that appears to be an error.
[...]
Cannot create a new thread as the limit on concurrent threads, 64, will be exceeded.
Could not create a thread: No space left on device
```

In above test case, when I induced this error when trying to start one thread more than `MAX_THREADS`, the error message comes out as: `No space left on device` -- which I think is quite misleading.

This is occurring because we have mapped this as follows in `platform_linux/platform_types.h`:

```
 25 #define STATUS_LIMIT_EXCEEDED CONST_STATUS(ENOSPC)
```
while STATUS_BUSY is mapped as:

```
 24 #define STATUS_BUSY           CONST_STATUS(EAGAIN)
```

... This seems more consistent.

gapisback:
@rtjohnso - I think the underlying problem is that `STATUS_LIMIT_EXCEEDED` is mapped to something that may not apply in all cases. 

Checking this [errno man page](https://man7.org/linux/man-pages/man3/errno.3.html), I could not find any `E<token>` that maps cleanly to our current usages of `STATUS_LIMIT_EXCEEDED` in our code base.

If you can suggest that is closely correct, we can amend the mapping in a different PR.

src/task.c
gapisback: Unfortunately, we have this mapping in our code base:

``` 25 #define STATUS_LIMIT_EXCEEDED CONST_STATUS(ENOSPC)`

which ends up spitting out a [misleading] `Could not create a thread: No space left on device` message.

I think we should look into properly remapping `STATUS_LIMIT_EXCEEDED` to a more appropriate errno-code, outside this PR (if you don't mind).

rtjohnso:
You're right.  I did a quick skim and didn't see any thing better. Proceed
as you had planned.  Thank you.

Best,
Rob

PR #532, opened 2023-01-15
(#156) Extend trunk_node_print_branches(). Call it from unit-test print_diags.

gapisback:
This commit adds minor improvements to trunk_node_print_branches() to:

- Print # of kv-bytes in unit-specifiers, also
- Print sum of {# of branches, num tuples & num of kv-bytes} for each trunk node.
- Add trunk_node_stats{} to gather metrics as we traverse trunk nodes during printing. Report one-line summary of aggregate metrics.

- Extend unit-test test_splinter_print_diags() to now also call trunk_print_branches() so we have a basic exercise of this print method.

--- NOTE: This PR started with an attempt to address the issue raised under #156 but upon investigation (and discussion with Alex), seems like that issue was incompletely specified. And possibly cannot be fully resolved.

So I am co-opting that issue to bring-in above diagnostic improvements to the workings of `trunk_print_branches()` and adding it to an existing unit-test case, `splinter_test test_splinter_print_diags` so that this print method continues to be exercised.


src/trunk.c
```diff
@@ -8811,75 +8824,135 @@ trunk_branch_count_num_tuples(trunk_handle *spl,
...
 
+/*
+ * Stats struct to return trunk node usage statistics via print interface
+ */
+typedef struct trunk_node_stats {
+   platform_log_handle *log_handle;
+   uint64               num_nodes;
+   uint64               num_tuples;
+   uint64               num_kv_bytes;
+} trunk_node_stats;
```
gapisback, line 8835: New structure added to gather trunk-node specific metrics as we traverse trunk nodes while printing them.

src/trunk.c
```diff
@@ -8811,75 +8824,135 @@ trunk_branch_count_num_tuples(trunk_handle *spl,
...
       trunk_branch_count_num_tuples(
          spl, &node, branch_no, &num_tuples_in_branch, &kv_bytes_in_branch);
-      uint64 kib_in_branch = 0;
+
+      nkvbytes = kv_bytes_in_branch;
+
+      sum_ntuples += num_tuples_in_branch;
+      sum_nkvbytes += kv_bytes_in_branch;
+
+      uint64 kib_in_branch = 1;
```
gapisback, line 8896: This is where a "fix" was sort-of attempted to resolve the main item raised under issue #156.

This changes old L8855 from value 0 to 1. Otherwise, kib_in_branch computed below on L8900, below, would always come out as 0.

This change does not really fix something; just makes some things come out as non-zero.

I'm not even sure if this is needed, and / or should be retained. Comments welcome during review discussion.

I think this area is where some more code / logic is needed as indicated by call to non-existent function trunk_branch_extent_count().

src/trunk.c
```diff
@@ -8811,75 +8824,135 @@ trunk_branch_count_num_tuples(trunk_handle *spl,
...
+   platform_log(log_handle, "{\n");
+   trunk_for_each_node(spl, trunk_node_print_branches, (void *)&node_stats);
+
+   platform_log(log_handle,
+                "Number of trunk nodes=%lu, Num tuples=%lu"
+                ", Num KV-bytes=%lu (%s)\n",
+                node_stats.num_nodes,
+                node_stats.num_tuples,
+                node_stats.num_kv_bytes,
+                size_str(node_stats.num_kv_bytes));
```
gapisback, line 8954: Adding a one-line summary of trunk-node metrics as gathered during print traversal.

tests/unit/splinter_test.c
```diff
@@ -661,6 +661,9 @@ CTEST2(splinter, test_splinter_print_diags)
 
    trunk_print_space_use(Platform_default_log_handle, spl);
 
+   CTEST_LOG_INFO("\n** trunk_print_branches() **\n");
+   trunk_print_branches(Platform_default_log_handle, spl);
```
gapisback, line 665: Extend test case to exercise this updated print method.

src/trunk.c
```diff
@@ -8811,75 +8824,135 @@ trunk_branch_count_num_tuples(trunk_handle *spl,
...
+
    uint16 start_branch = trunk_start_branch(spl, &node);
    uint16 end_branch   = trunk_end_branch(spl, &node);
+   uint16 num_branches = 0;
+   uint64 sum_ntuples  = 0;
+   uint64 sum_nkvbytes = 0;
+
+   size_t nkvbytes = 0;
    for (uint16 branch_no = start_branch; branch_no != end_branch;
         branch_no        = trunk_add_branch_number(spl, branch_no, 1))
```
gapisback, line 8881: @ajhconway -- We need to discuss how to re-work this iteration.

When I spoke with you briefly about this (in Dec 2022), you remarked that this iteration is possibly even incorrect, and may even have a race condition.

I recall you said something about this iteration is not paying attention to "dead pivots" [?] or something to that effect.

Would like to get some code references from `trunk.c` to places where we are doing the right iteration, so I can apply that logic here, too.

src/trunk.c
rtjohnso: Without actually counting the extents in the branch, which we're not doing any more, we cannot compute a space amp.  And kib_in_branch is superceded by nkvbytes.

So I think we can just kill 8896-8901 and 8911-8912.

rtjohnso (changes requested):

gapisback:
@rtjohnso -- As discussed in the PR meeting today, I've deleted the code in the `trunk_node_print_branches()` that was trying to compute the space amplification factor. Please see this [updated delta-commit](https://github.com/vmware/splinterdb/pull/532/commits/a3030eac43a200ade3831dac77f29ca3607acc92) for the new diffs.

To record what we discussed in the PR meeting:

I have not tried to address the potential 'race' condition Alex had remarked on earlier. We observed during our review chat that this function is doing a `trunk_node_get(spl->cc, addr, &node);` at the start, so there is some protection in-place.

Also,  all this change-set is doing is invoking this print function in a stand-alone single-user test case, so there should not be any risk due to race conditions for this controlled usage. This change-set is not worsening current behaviour. 

With this change, the new output from this print routine exercised by this `bin/unit/splinter_test test_splinter_print_diags` is as follows:

```
------------------------------------------------------------------
| Trunk node=393216, height=1
------------------------------------------------------------------
| pivots: 9
| 0: (negaitive_infinity)
| 1: 0x20001c5108e6550700000000000000000000000000000000
| 2: 0x3ffc153e7bc3f0e200000000000000000000000000000000
| 3: 0x5ffc880edb6b2b6400000000000000000000000000000000
| 4: 0x7ffe083c44f5d26e00000000000000000000000000000000
| 5: 0xa0017cf61243b18400000000000000000000000000000000
| 6: 0xc001fa28e521753f00000000000000000000000000000000
| 7: 0xdfffde43b6997c9d00000000000000000000000000000000
| 8:
-------------------------------------------------------------------------
    | branch |     addr     |  num tuples  |       num kv bytes         |
-------------------------------------------------------------------------
    |     36 |   4319608832 |        74352 |     5872541    (~5.60 MiB) |
    |     37 |   4405067776 |        74365 |     5879931    (~5.60 MiB) |
    |     38 |   4490919936 |        74507 |     5878773    (~5.60 MiB) |
    |     39 |   4577427456 |        75434 |     5950166    (~5.67 MiB) |
    |     40 |   4937744384 |       148621 |    11747471   (~11.20 MiB) |
    |     41 |   5023596544 |       149569 |    11812565   (~11.26 MiB) |
    |     42 |   5110628352 |       149979 |    11844334   (~11.29 MiB) |
    |     43 |   5197660160 |       149595 |    11811727   (~11.26 MiB) |
    |     44 |   5586550784 |       225361 |    17802848   (~16.97 MiB) |
    |     45 |   5673844736 |       223863 |    17695335   (~16.87 MiB) |
    |     46 |   5761400832 |       224262 |    17713499   (~16.89 MiB) |
    |     47 |   6170083328 |       298458 |    23577825   (~22.48 MiB) |
-------------------------------------------------------------------------
Sum |     12 |              |      1868366 |   147587015  (~140.74 MiB) |
-------------------------------------------------------------------------
```

gapisback:
Old-moldy PR. Has gone thru several start-stop attempts to get it finalized. General interest from team / colleagues has been  low on this set of changes, especially as some of the affected areas were broken even before this attempt to clean stuff up.

No more value in keeping this PR alive. The stakeholders and actors involved have move-on, and so will I shortly.

Nuke it. Let this rest.

Issue #534, opened 2023-01-18
Rework env-var interface to set log streams from unit-tests that exercise print diagnostics code.


gapisback:
Under PR #514, couple of new interfaces have been added for use (mainly) by unit-tests to set the output log streams. These functions are `set_log_streams_for_tests()`, and `set_log_streams_for_error_tests()`.

These `set` methods redirect outputs to `stdout` and `stderr` if env-var `VERBOSE` is set (1).

There are couple of unit-test cases which exercise print diagnostic functions. E.g. in `unit/btree_stress_test.c` we exercise BTree-print diagnostic code. (Under upcoming PR #531 , this test-case is being refactored into its own `test_btree_print_diags()` test-case.) Another example is `unit/splinter_test.c` which has this `test_splinter_print_diags()` test case.

Both these test cases redirect output using two different interfaces: `set_log_streams_for_tests()` and `set_log_streams_for_error_tests()`. That's a minor mis-use of these interfaces.

The real issue is that output from test cases like `test_btree_print_diags()` becomes very huge. It's over 6M usually. This causes CI browsers to become unwieldy when looking at such outputs.

The things we need to resolve under this issue are:

1. Rework the interfaces so that when these tests are run in CI (which currently invokes stuff with `VERBOSE=1`) that none of this output is generated to `stdout` or `stderr` (causing browsers to die).

2. Provide another way for people to be able to run these tests on their own dev machines and collect the outputs from `stdout`. Perhaps, we could consider different values for the env-var which will set up the desired log stream handles.

3. Consider adding a new `set_log_streams_for_diagnostic_tests()` interface. Rework the few unit-tests that are currently exercising these print diagnostics to use this interface (instead).

4. When the support for `VERBOSE=1` was added (under PR #494), the general agreement from reviewers was that it would be better to have `--cmd-line-flags` to support different verbose print options, rather than this env-var. Review that approach and come up with a solution to address all usages / requirements.

In that last PR #494, following note was recorded :

----
Some time ago, I had added this --verbose-progress option that was used initially in functional/splinter_test.c to do some verbose progress reporting while large test workloads run. This arg is parsed and tracked as a bool in test_exec_config->verbose_progress.

With some work we could co-opt this flag to also mean what this VERBOSE=1 env-var wants to do. This boolean field may be the replacement for your newly-added Ctest_verbosity flag.

---
Here are some cmdline options to consider adding:

- `--verbose-progress`: Current interface, to print verbose messages from functional tests
[...]

PR #541, opened 2023-02-07
Add initial support for message logging levels, used in unit-tests.

gapisback:
This commit slightly adjusts the way output messages from C-unit-tests will be printed. Introduce initial set of message logging levels, initially scoped only for use in unit-tests.

For test cases that invoke print / diagnostic code, add support for VERBOSE=7, under which (voluminous) output will be generated. For CI-runs, which set env-var VERBOSE=1, this change will suppress generating large outputs which would otherwise (potentially) cause browsing of test output files to fail.

Default behaviour that unit-tests will print no messages remains. Existing env-var VERBOSE=1, we continue to only print error messages.

-----
NOTE: This PR only partially address the issues raised under issue #534 . 

Currently, the main requirement for getting this re-work into /main is to suppress large outputs coming out from `splinter_test test_splinter_print_diags`, which is one of the test cases exercising print diagnostic code.

There are a few more draft PRs in the pipeline which are adding more such test cases invoking print code. The outputs from those tests, when added to `/main` will further clog up the test output files in CI, making browsing difficult.

So, this PR lays down an interim solution, introducing nascent support for log-message levels, which can be 'controlled' by existing env-var `VERBOSE`.

This way, we now have a solution for print-diagnostics test cases even while retaining the desired behaviour that most unit-tests are silent except for specialize unit-tests (`misc_test`, `limitations_test`) which want to print outputs from error messages, for example.


include/splinterdb/platform_linux/public_platform.h
```diff
@@ -78,3 +78,7 @@ typedef FILE platform_log_handle;
 void
 platform_set_log_streams(platform_log_handle *info_stream,
                          platform_log_handle *error_stream);
+
+// Reset ouput log streams to their default settings.
+void
+platform_reset_log_streams(void);
```
gapisback, line 84: Used to re-set output streams to initial (default) setting in case the env-var level is -not- what the test-case asks for.

tests/unit/btree_stress_test.c
```diff
@@ -113,7 +113,7 @@ CTEST_DATA(btree_stress)
 // Setup function for suite, called before every test in suite
 CTEST_SETUP(btree_stress)
 {
-   set_log_streams_for_error_tests(NULL, NULL);
+   set_log_streams_for_tests(MSG_LEVEL_ERRORS);
```
gapisback, line 116: First, this is a bug-fix to use the proper interface.

Second, the way to understand this call is that: The test is saying that: "I want to see output ONLY when the output logging-level is MSG_LEVEL_ERRORS (i.e., 3)".

Maybe I could change the check in `set_log_streams_for_tests` to check for:

   `if (Ctest_verbosity >= exp_msg_level)`
   
... But I wanted to have the reviewers give their opinion before going down that route.

tests/unit/ctest.h
```diff
@@ -79,14 +79,24 @@ struct ctest {
    unsigned int magic;
 };
 
+/*
+ * Message verbosity levels:  Modeled on syslog's severity levels.
+ * See: https://en.wikipedia.org/wiki/Syslog
+ */
+typedef enum msg_level {
```
gapisback, line 86: I considered the option of externalizing these by adding them to `include/splinterdb/platform_linux/public_platform.h`, but given that this is still preliminary code, and still cooking ... I figured it's best to scope these levels only to CTest code. For now.

Once we get a better handle on how all this plays out ... we could look into formalizing these levels in an external `.h` file.

tests/unit/unit_tests_common.c
```diff
@@ -17,10 +17,12 @@
  * when the caller sets the VERBOSE env var to opt-in.
  */
 void
-set_log_streams_for_tests()
+set_log_streams_for_tests(msg_level exp_msg_level)
 {
-   if (Ctest_verbose) {
+   if (Ctest_verbosity == exp_msg_level) {
```
gapisback, line 22: Qs. Do you want to change this to `>=` here?

If we did that, then, when someone runs: `VERBOSE=7 ./bin/unit/misc_test`, they will start seeing error messages.

It will not be an issue for normal usages and for CI, but that's what will happen.

Let's decide which way to go ... so I can finalize the check here.

rosenhouse: Yes, `>=` is the normal semantic.

tests/unit/main.c
```diff
@@ -399,20 +400,44 @@ ctest_is_unit_test(const char *argv0)
...
-// Determine the log message verbosity to use for this test run.
-// If env var VERBOSE is unset, or is set to "0" or "false", the be quiet.
-// Otherwise, be verbose.
-static bool
+/*
+ * Determine the log message verbosity level to use for this test run.
+ *
+ * If env var VERBOSE is unset, or is set to "0" or "false", then be quiet.
+ * If env var VERBOSE=1 or "true", be verbose reporting only errors.
+ * Otherwise, settle for the specified verbosity level.
```
rosenhouse, line 408: I think it's ok to break the "false" and "true" usage in service of having a simple numeric representation.

tests/unit/ctest.h
```diff
@@ -79,14 +79,24 @@ struct ctest {
...
-extern _Bool Ctest_verbose;
+extern msg_level Ctest_verbosity;
 
-// For immedate logging to stdout
+// For immediate logging to stdout at any message verbosity level.
 // (contrast with CTEST_LOG which does buffered/delayed logging to stderr)
 #define CTEST_LOG_INFO(...)                                                    \
    do {                                                                        \
-      if (Ctest_verbose) {                                                     \
+      if (Ctest_verbosity != MSG_LEVEL_SILENT) {                               \
```
gapisback, line 99: Old code is "fine" but the rewrite is more true to form.

tests/unit/splinter_test.c
```diff
@@ -634,6 +634,8 @@ CTEST2(splinter, test_lookups)
  */
 CTEST2(splinter, test_splinter_print_diags)
 {
+   set_log_streams_for_tests(MSG_LEVEL_DEBUG);
```
gapisback, line 637: This call will setup the output log handles if test-case is run with `VERBOSE=7`. This allows for manual inspection of the outputs while suppressing tons of output in CI (which invokes tests using `VERBOSE=1`).

rosenhouse (changes requested):
This code does not work the way I expect it too, I think there's a regression here.  Try running 
`build/release/bin/unit/limitations_test` without setting `VERBOSE` at all.  Test on `main` vs on this branch.  On `main`, the output is 14 lines showing only the tests passing.  On this branch the output is 53 lines long and includes all the error messages that I'd expect to be supressed.

tests/unit/btree_stress_test.c
rosenhouse: I'd describe the meaning of this call differently.  It isn't about the test saying what output it wants to see.  It's about the test case declaring that it will be logging messages to the log-stream at a given severity (or at a more severe / lower number).  Therefore, the test-runner ought to have the log-stream map to stdout if and only if the user has requested to see messages at that severity (or less severe / higher number).  Otherwise, it should map to /dev/null.

Ideally, that logic would be on a per-call basis.  But since that would be a lot more work than we want to do right now, we instead have it at the granularity of per-test-case.

tests/unit/main.c
gapisback: Applied this change in latest upload.

tests/unit/unit_tests_common.c
gapisback: Changed accordingly. See latest upload.

gapisback:
You are right, @rosenhouse -- A change I made in a hurry to `limitations_test.c` had regressed behaviour. It's fixed now w/ the latest upload.

tests/unit/ctest.h
rosenhouse: And in that spirit, I think `>=` would better convey the overall semantics.

tests/unit/main.c
```diff
@@ -399,20 +400,40 @@ ctest_is_unit_test(const char *argv0)
...
-   return 1;
+
+   if (STRING_EQUALS_LITERAL(val, "1")) {
+      return MSG_LEVEL_ERRORS;
+   }
+   if (STRING_EQUALS_LITERAL(val, "7")) {
+      return MSG_LEVEL_DEBUG;
+   }
+   // Unknown verbosity level; fall-back to silent output mode.
+   return MSG_LEVEL_SILENT;
```
rosenhouse, line 436: Is it much harder to use `atoi` or the modern secure equivalent?  This "exact-match only with fallback" behavior is pretty confusing e.g. `VERBOSE=3`  is the same as `VERBOSE=0`.

tests/unit/main.c
```diff
@@ -545,6 +566,13 @@ ctest_usage(const char *progname, int program_is_unit_test)
    } else {
       printf("[ --list | [ --<config options> ]* <test-case-name> ]\n");
    }
+   printf(
+      "\nUse env-var VERBOSE=<n> to control verbosity of output from tests.\n");
+   printf("  %d - Or, unset, turns off all output from tests. (Default).\n",
+          MSG_LEVEL_SILENT);
+   printf("  %d - Only error messages are printed.\n", MSG_LEVEL_ERRORS);
```
rosenhouse, line 573: This isn't correct right now, since `MSG_LEVEL_ERRORS` is 3, but the `ctest_get_verbosity()` function silently treats 3 the same as 0.  I suggest fixing `ctest_get_verbosity()`.

tests/unit/btree_stress_test.c
rosenhouse: Yeah, given the current implementation, this argument needs to be `MSG_LEVEL_DEBUG`.  Because on the current commit, 
```
VERBOSE=1 build/release/bin/unit/btree_stress_test
```
gives all the debug output.

rosenhouse (changes requested):
`btree_stress_test` is setting the wrong level.

tests/unit/btree_stress_test.c
gapisback: Fixed the bug in `btree_stress_test.c`. Thanks for catching it.

Also took the opportunity to add a prolog comment to `set_log_streams_for_tests()`, to help future authors. 

Let me know if you wish to wordsmith that even more.

tests/unit/ctest.h
gapisback: But then `>=` implicitly assumes that `MSG_LEVEL_SILENT` is  == 0 . 

W/o relying on the value of this token, it's better to say `!=`;  I think.

tests/unit/main.c
gapisback: I actually had coded it up to use `atoi` in an earlier version.

But then I felt `VERBOSE=3` does nothing as that is an unsupported verbosity level. So, supporting that (or even allowing it thru this interface) appeared confusing by itself to me.

Which is why I went the route of only 'honoring' verbosity levels we know of now and support. The remaining unknown ones default (fall-back) to silent.

This way, one day when we do fill out the `MSG_LEVEL_<>` stream, the usage of `VERBOSE=4` will properly mean and map to something supported.

tests/unit/main.c
gapisback: Not sure how to fix this in `ctest_get_verbosity` for  handling 3. Only `VERBOSE=1` is being allowed now and the `1` is mapped to `MSG_LEVEL_ERRORS`.

I agree the help message reporting 3 is incorrect. I think I should just change that to hard-coded 1 as that's how env-var `VERBOSE=1` is operating.

Let's talk ...

gapisback: @rosenhouse : I fixed the `--help` output to now report 1 for printing errors. The output looks like:
```
Use env-var VERBOSE=<n> to control verbosity of output from tests.
  0 - Or, unset, turns off all output from tests. (Default).
  1 - Only error messages are printed.
  7 - All messages, including debugging output is printed.
```

I think that should be sufficient.

I did think about your comment of `suggest fixing ctest_get_verbosity().` I thought of amending this code-fragment in that function:

```
 428    if (STRING_EQUALS_LITERAL(val, "1")) {
 429       return MSG_LEVEL_ERRORS;
 430    }
```

to also allow 3; i.e. change L428 to : `if (STRING_EQUALS_LITERAL(val, "1") || STRING_EQUALS_LITERAL(val, "3"))`

That would take us a bit closer to the eventual msg-level of `error == 3`.

But I figured that as eventually we plan to remove this `VERBOSE=` env-var support in lieu of a proper `--verbose*` cmd-line flag, I felt it's not that critical to "improve" the handling at this time.

Let me know if you disagree and still wish to see something else done here.

gapisback:
@rosenhouse -- I cleaned up the code / comments based on our discussion this morning. Have run some sanity tests to verify that the outputs are coming out 'as expected'. Just pushed a new commit. 

Please take a final look and re-test to ensure that things are working as we expected / converged on.

tests/unit/unit_tests_common.c
```diff
@@ -12,71 +12,30 @@
...
  *
  * Some unit tests exercise error cases, so even when everything succeeds, they
  * generate lots of "error" messages. By default, that would go to stderr, which
  * would pollute the test output. This function sets up output file handles such
  * that those expected error messages are only printed when the caller sets the
- * VERBOSE env var to opt-in. By default, execution from unit-tests will be
- * silent, redirecting output handles to /dev/null.
- *
- * Returns: Output file handles (out/error) if caller has requested for them.
+ * VERBOSE env var to opt-in.
```
gapisback, line 24: You may want to review this edited prolog comment block ... for accuracy.

tests/unit/unit_tests_common.c
```diff
@@ -12,71 +12,30 @@
...
-      if (log_stderr) {
-         *log_stderr = stderr;
-      }
-      CTEST_LOG_INFO("\nVerbose mode on. This test may exercise an error case, "
-                     "so on success it will print a message that appears to "
-                     "be an error. "
-                     "Or, the test may exercise print-diagnostic code.\n");
+      CTEST_LOG("Verbose mode on. This test may exercise an error case, so on"
+                " success it may print a message that appears to be an error."
+                " Or, the test may exercise print-diagnostic code.\n");
```
rosenhouse, line 35: One thing I didn't consider when we discussed this before is that, now that we've combined these two functions together, if you run a non-error-exercising test with `VERBOSE=3` you'll see this message, which is kind of noise since it is never true.  

I think I'd be ok with either of the following "fixes":
1. Just shorten this message so it's less than 80 characters, so at least its not too much noise
2. Add a boolean argument to this function, something like `bool may_print_errors`, which would control whether or not this log line is shown.  And then pass in `true` from `limitations_test` and such, but `false` otherwise.

include/splinterdb/platform_linux/public_platform.h
```diff
@@ -78,3 +78,6 @@ typedef FILE platform_log_handle;
 void
 platform_set_log_streams(platform_log_handle *info_stream,
                          platform_log_handle *error_stream);
+
+platform_log_handle *
+platform_get_stdout_stream(void);
```
gapisback, line 83: See if this can be located to `platform.h`

gapisback: Reworked to locate prototype to `platform.h`; that worked out ok.

tests/unit/unit_tests_common.c
gapisback: @rosenhouse - I did a large'ish rework of all this handling and updated `set_log_streams_for_tests()` to a large extent. Lots has changed.

See [this commit](https://github.com/vmware/splinterdb/pull/541/commits/00d9c5abc0e29fdc287b580f91a33e0ea14e34f5).

tests/unit/main.c
```diff
@@ -399,20 +407,30 @@ ctest_is_unit_test(const char *argv0)
...
-static bool
+/*
+ * Determine the log message verbosity level to use for this test run.
+ *
+ * If env var VERBOSE is unset, or is set to "0" then be quiet.
+ * If env var VERBOSE=3 then be verbose reporting only errors.
+ * Otherwise, settle for the specified verbosity level.
+ *
+ * NOTE: Currently, the only other level we support is level=7 to print all
+ *       debugging output.
```
rosenhouse, line 418: Revise this comment line, since other values are used and are ok.

tests/unit/unit_tests_common.c
rosenhouse: I think it is accurate, but perhaps incomplete.  It would be good to call out some actionable instructions for unit test authors.  For example, we could add something like:
> Tests which exercise error paths in the library should first call this function with MSG_LEVEL_ERRORS, so that any error messages generated by the library are only shown when the test is run with VERBOSE >= 3.
>
> Tests which cause the library to generate debug output (e.g. by printing internal data structures) should call this function with MSG_LEVEL_DEBUG, so that that output is only shown when the test is run with VERBOSE>=7.

rosenhouse (approved):
Fantastic.  I just tested it out and it works exactly how I hoped.  Thank you!  I left a couple small suggestions, but overall it looks good.

tests/unit/main.c
gapisback: I just deleted this NOTE: line as trying to improve it got harder than the value added.

rosenhouse:
I have applied the CI config changes here to our Concourse server, so the new `VERBOSE: 3` level should be used along with this code the next time CI is triggered.

PR #542, opened 2023-02-07
Fixups to enable building this branch with clang

rosenhouse:
Note this is a PR against your `pg-splinterdb-shmem-integn-1` branch.

My editor uses clang for various IDE-like things.  It's nice to have this branch buildable with clang.


src/platform_linux/shmem.c
```diff
@@ -218,20 +218,6 @@ platform_heap_id_to_shmaddr(platform_heap_id hid)
...
-   return (void *)platform_heap_id_to_handle(hid);
-}
-
-/* Evaluates to valid 'high' address within shared segment. */
-static inline void *
-platform_shm_hip(platform_heap_id hid)
-{
-   return (((shmem_info *)platform_heap_id_to_shmaddr(hid))->shm_end - 1);
-}
-
```
gapisback, line 234: I suppose you are deleting these functions as they are defined as `static` but are not used anywhere in this file.

I thought I had used it for some assertion checks / diagnostics, but can't find those usages now.

I have found these to be useful during development and would like to retain them, just so that they are available next time I have to add some diagnostic xchecks to verify that an address is within a shared segment region.

What if you just make them non-`static` and let them be in this file. Or, even remove the `inline` qualifier, if needed. Will that allow compilation to go through? If so, can we retain these two.

src/platform_linux/shmem.c
```diff
@@ -517,11 +503,12 @@ platform_shm_alloc(platform_heap_id hid,
    }
 
    // Optimistically, allocate the requested 'size' bytes of memory.
-   retptr = __sync_fetch_and_add(&shminfop->shm_next, size);
+   _Static_assert(sizeof(void*)==sizeof(size_t), "check our casts are valid");
```
gapisback, line 506: It's ok to add it here; I understand why you are doing it. But, it's a very low-level assertion and I think it probably belongs elsewhere in `platform.c`.

You may choose to not do anything right now as this is proto code. I think it would be better to create a new function in `platform.c` named:

`void __attribute__((constructor)) platform_asserts()`

and stuff all such platform-wide asserts in one place.

(This is just a suggestion; feel free to file it away for a future date.)

gapisback, line 506: Please run `clang-format` on this .. not sure if CI is running on this fix.

gapisback (approved):
@rosenhouse - thanks for this cleanup to get `clang` builds working. I have put down minor comments, some of which can be ignored for now as this is proto code.

src/platform_linux/shmem.c
rosenhouse: Ok, I removed the `static` and that was enough to silence the compiler errors

src/platform_linux/shmem.c
rosenhouse: I'm going to leave this as-is for now.  We can return to it if/when we're prepping this branch for integration to `main`

rosenhouse: Added a separate commit for this, since it fixed up other files too.

rosenhouse:
Turns out removing the `static` broke the GCC build.  So I undid that and fixed it a different way: https://github.com/vmware/splinterdb/commit/b9ea7ef29acdd0d5ce6c34211bfde3d14adb7093

Pushed that fix directly to the branch since it was so small.

PR #543, opened 2023-02-13
Replace all instances of platform_default_log() with CTEST_LOG_INFO()

gapisback:
Also, minor fix in exec_worker_thread() in large_inserts_bugs_stress_test.c to avoid divide-by-0 error when elapsed_ns is too small.

-- Note to @rosenhouse : With this change, all unit-tests added newly by my shared-memory prototype work are changed to consistently use `CTEST_LOG_INFO()`. This brings them in-sync with recently added unit-test verbosity levels interfaces and behaviour.

Here is the confirmation:

```
agurajada/gabe/shmem-squash-logging-ovhds-poc

sdb-fdb-build:[16] $ grep -l -w platform_log *.c
sdb-fdb-build:[17] $ grep -l -w platform_default_log *.c
```

### DIFF
--- a/tests/unit/large_inserts_bugs_stress_test.c
+++ b/tests/unit/large_inserts_bugs_stress_test.c
@@ -198,7 +198,7 @@ CTEST2_SKIP(large_inserts_bugs_stress,
       uint64 elapsed_ns      = platform_timestamp_elapsed(start_time);
       uint64 test_elapsed_ns = platform_timestamp_elapsed(test_start_time);
 
-      platform_default_log(
+      CTEST_LOG_INFO(
          PLATFORM_CR
          "Inserted %lu million KV-pairs"
          ", this batch: %lu s, %lu rows/s, cumulative: %lu s, %lu rows/s ...",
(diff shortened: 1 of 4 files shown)


tests/unit/splinter_test.c
```diff
@@ -129,9 +129,6 @@ CTEST_SETUP(splinter)
    heap_capacity        = MAX(heap_capacity, 2 * GiB);
 
    bool use_shmem = test_using_shmem(Ctest_argc, (char **)Ctest_argv);
-   if (use_shmem) {
-      platform_default_log("Run test using shared memory segment.\n");
```
gapisback, line 133: Not needed any more, as CTests' main() also checks and reports when tests are running with `--use-shmem` option.

tests/unit/large_inserts_bugs_stress_test.c
```diff
@@ -711,15 +710,19 @@ exec_worker_thread(void *w)
       }
    }
    uint64 elapsed_ns = platform_timestamp_elapsed(start_time);
+   uint64 elapsed_s  = NSEC_TO_SEC(elapsed_ns);
+   if (elapsed_ns == 0) {
+      elapsed_s = 1;
```
gapisback, line 715: Bug-fix to deal with very small `elapsed_ns` time values.

rosenhouse (approved):
I've manually run the affected tests and they all work as expected.

Issue #546, opened 2023-03-11
Segmentation fault after splinterdb_stats_reset()


deukyeon:
There is a bug that causes a segmentation fault when recording histograms again after calling `splinterdb_stats_reset()`, as shown in the gdb output:
```
Thread 30 "ycsbc" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffe8bbff640 (LWP 1843516)]
trunk_insert (spl=<optimized out>, tuple_key=..., data=...) at src/trunk.c:6188
6188                platform_histo_insert(spl->stats[tid].update_latency_histo,
```

The issue seems to be that the histogram is zeroed out in `trunk_reset_stats()`, which is called by `splinterdb_stats_reset()`, as shown below:
```
void
trunk_reset_stats(trunk_handle *spl)
{
   if (spl->cfg.use_stats) {
      memset(spl->stats, 0, MAX_THREADS * sizeof(trunk_stats));
   }
}
```

PR #547, opened 2023-03-11
(#546) Fix the segmentation fault after splinterdb_stats_reset()

deukyeon:
Previously, when `trunk_stats_reset()` was called by splinterdb_stats_reset(), the entire statistics of a trunk, including the histogram handles, were reset to zero.


rtjohnso (approved):
LGTM

Issue #548, opened 2023-03-20
Definition of  typedef bool  as int32 on Linux causes integration issues w/ other s/w that define bool as BYTE.


gapisback:
The definition of `typedef int32 bool` in our `splinterdb/platform_linux/public_platform.h` caused me a very basic integration issue with other s/w code which links with SplinterDB library. The problem is that the \<some other\> code defines this as `unsigned char`, which is 1-byte. Splinter's definition of `bool` as `int32` throws-off some structure layouts.

Specifically, see the layout of public `splinterdb_config{}` as of SHA d9fcc407 in `/main`, defined in `include/splinterdb/platform_linux/public_platform.h`:

```
 23 // Configuration options for SplinterDB
 24 typedef struct {
 25    // required configuration
 26    const char *filename;
 27    uint64      cache_size;
 28    uint64      disk_size;
[...]
 49    // cache
 50    bool        cache_use_stats;
 51    const char *cache_logfile;
[...]
 65    uint64 num_memtable_bg_threads;
 66    uint64 num_normal_bg_threads;
 67
 68    // btree
 69    uint64 btree_rough_count_height;
 70
 71    // filter
 72    uint64 filter_remainder_size;
 73    uint64 filter_index_size;
 74
 75    // log
 76    bool use_log;
 77
 78    // splinter
 79    uint64 memtable_capacity;
```

Due to the differing definitions of `typedef bool`, the offset of fields past the 1st instance of `bool` (i.e. L50 `bool cache_use_stats`) will differ in the `.o`-s compiled into the library and in the objects for the \<other code\> base.

This makes the system unstable.

Asking internally with SplinterDB core-dev team, we agreed to stabilize on 1-byte `bool` fields. 

Longer-term may be the better way would be to `#include <stdbool.h>` but that's not being followed through right now.

gapisback:
For historical reference, this issue was aired out internally in these threads:

1. 21.Dec.2022: [Original slack post](https://vmware.slack.com/archives/G01MEDQ05MX/p1671671258434059)
2. 20.Mar.2023: Follow-up [post](https://vmware.slack.com/archives/G01MEDQ05MX/p1679340319477579), leading to closure and creation of this issue

PR #549, opened 2023-03-20
(#548) Change typedef of bool from int32 to 1-byte 'unsigned char'.

gapisback:
In SplinterDB's public splinterdb_config{} config, we have few fields defined as 'bool' which is typedef'ed to int32 on our side. This creates compatibility problems when linking this library with other s/w which may have defined 'bool' as 1-byte field. (Offsets of fields in the splinterdb_config{} struct following 1st field defined as 'bool' changes across objects. This commit addresses this issue by defining boolean fields in public struct to be of type `_Bool`. This should, hopefully,reduce the risk of such incompatibilities. 

(Splinter's internal definition of `bool` remains as `int32`, to avoid any other product change / compatibility issues.)

Found only one instance where 'bool' return value from routing_filter_is_value_found() had to be adjusted to fit 1-byte return value.

### DIFF
--- a/src/routing_filter.h
+++ b/src/routing_filter.h
@@ -122,7 +122,7 @@ routing_filter_get_next_value(uint64 found_values, uint16 last_value)
 static inline bool
 routing_filter_is_value_found(uint64 found_values, uint16 value)
 {
-   return found_values & (1 << value);
+   return ((found_values & (1 << value)) != 0);
 }
 
 
--- a/include/splinterdb/data.h
+++ b/include/splinterdb/data.h
@@ -90,10 +90,10 @@ merge_accumulator_length(const merge_accumulator *ma);
 slice
 merge_accumulator_to_slice(const merge_accumulator *ma);
 
-bool
+_Bool
 merge_accumulator_copy_message(merge_accumulator *ma, message msg);
 
-bool
+_Bool
 merge_accumulator_resize(merge_accumulator *ma, uint64 newsize);
 
 void
--- a/include/splinterdb/platform_linux/public_platform.h
+++ b/include/splinterdb/platform_linux/public_platform.h
@@ -61,9 +61,6 @@ static_assert(sizeof(uint64) == 8, "incorrect type");
 #   define FALSE (0)
 #endif
 
-#if !defined(__cplusplus)
-typedef int32 bool;
-#endif
 typedef uint8 bool8;
 
 typedef FILE platform_log_handle;
--- a/include/splinterdb/public_util.h
+++ b/include/splinterdb/public_util.h
@@ -27,7 +27,7 @@ typedef struct slice {
 #define NULL_SLICE    ((slice){.length = 0, .data = NULL})
 #define INVALID_SLICE ((slice){.length = (uint64)-1, .data = NULL})
 
-static inline bool
+static inline _Bool
 slice_is_null(const slice b)
 {
    return b.length == 0 && b.data == NULL;
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -47,7 +47,7 @@ typedef struct {
    uint64 io_async_queue_depth;
 
    // cache
-   bool        cache_use_stats;
+   _Bool       cache_use_stats;
    const char *cache_logfile;
 
    // task system
@@ -73,7 +73,7 @@ typedef struct {
    uint64 filter_index_size;
 
    // log
-   bool use_log;
+   _Bool use_log;
 
    // splinter
    uint64 memtable_capacity;
@@ -237,7 +237,7 @@ void
 splinterdb_lookup_result_deinit(splinterdb_lookup_result *result); // IN
 
 // Returns true if the result was found
-bool
+_Bool
 splinterdb_lookup_found(const splinterdb_lookup_result *result); // IN
 
 // Decode the value from a found result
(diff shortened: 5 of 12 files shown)


src/routing_filter.h
```diff
@@ -122,7 +122,7 @@ routing_filter_get_next_value(uint64 found_values, uint16 last_value)
 static inline bool
 routing_filter_is_value_found(uint64 found_values, uint16 value)
 {
-   return found_values & (1 << value);
+   return ((found_values & (1 << value)) != 0);
```
gapisback, line 125: @rtjohnso - Weirdly, this one test to do with filters is tripping due to above code. The old code would return any int value > 255 to mean `TRUE`'. Now, if `value` is 8, `(1 << value)` becomes 256, and so `found_values & (1 << value)` no longer fits in 1-byte unsigned char return type. 

This fn could incorrectly return `FALSE` in cases where it should have returned `TRUE`.

With this fix, my manual runs of `test.sh` are clean, so I am now going to rely on CI to ferret out bugs due to any other such wrong code.

gapisback:
@rtjohnso - Triggered by the fix I discovered as being necessary in `src/routing_filter.h`, I am doing a code walk thru myself, cross-checking all functions that are defined to return `bool`, to see if other similar issues are lurking elsewhere.

I am not finding any ... hopefully, this fix in `src/routing_filter.h` is the ONLY place where code needs to be adjusted.

gapisback:
@rtjohnso - I completed my code-walk through for all instances of `bool` in the code. I found a few straggling places where code was using hard-coded `0` or `1` for boolean variables. I pushed [this commit](https://github.com/vmware/splinterdb/commit/ffb258eac3eb59f5705f1b8113a788f4b86b957e) to do some more minor clean-up.

None of this changes the logic / semantics, but will be good to make sure I get one more pair of eyes over it (lest, I'm making some silly mistakes).

gapisback:
HI, @rtjohnso - I reworked this fix as per what we discussed this past Tue's PR meeting. Now, in public config struct all boolean fields are defined as `_Bool`.

The rest of the change set stays as it was before, including minor cleanup for (what was previously) one instance of a bug, and other cleanup of hard-coded uses of 0 /1 with FALSE / TRUE.

src/routing_filter.h
rtjohnso: Good catch!

tests/unit/misc_test.c
```diff
@@ -267,6 +267,23 @@ CTEST2(misc, test_size_to_str)
...
+{
+   /*
+    * Re-confirming what this compiler implements for the C-standard
+    * _Bool type. We had issues where the use of our internal 'bool'
+    * type in public structs would throw-off field offsets. The fix
+    * was to use _Bool, so that there is no difference in the definition
+    * of this type in our library v/s in other s/w code that links
+    * with splinterdb library.
+    */
+   ASSERT_EQUAL(1, sizeof(_Bool));
```
rtjohnso, line 280: We don't care what size _Bool is.  Delete this one.

tests/unit/misc_test.c
```diff
@@ -267,6 +267,23 @@ CTEST2(misc, test_size_to_str)
...
+
+   // Our internal usage continues to be this, for historical reasons
+   ASSERT_EQUAL(4, sizeof(bool));
```
rtjohnso, line 283: Can we make this a static assert?

rtjohnso (changes requested):
Don't forget to move the definition of `bool` from the public platform to the private platform.

gapisback:
@rtjohnso - It's not straightforward w/o too much work to address your suggestion of : "_Don't forget to move the definition of bool from the public platform to the private platform._"

There are some cascading dependencies which cause compilation problems.

For instance, in `include/splinterdb/public_platform.h` we explicitly document a dependency that:

```
 14 The external definitions include:
 15 - uint64, which is used for buffer sizes, etc.
 16 - bool, typedef'd to int32 on linux
```

I had tried to move this typedef to `src/platform_linux/platform.h` and this immediately results in a compilation failure:

```
Compiling            src/btree.c                                        [build/release/obj/src/btree.o]
In file included from include/splinterdb/data.h:23,
                 from src/btree_private.h:14,
                 from src/btree.c:4:
include/splinterdb/public_util.h:30:15: error: unknown type name ‘bool’
   30 | static inline bool
      |               ^~~~
```

This is (partly) why in my original version of the fix, I had retained the location of `typedef bool` to be in `include/splinterdb/platform_linux/public_platform.h` but changed it to `unsigned char`.

Unless I'm missing something, I suspect to implement this new approach (of using `_Bool` in extern headers, and to use `bool` as `int32` for internal usages) will need some `.h` file reorganization.

I'm not sure it's a good idea to take that on right now. It should be a separate item.

rtjohnso:
If we leave our definition of bool in the public header, then won't you
still have your postgres problem?

Just change the static inline bool functions in public headers to use _Bool
instead.

Best,
Rob

gapisback:
If we leave our definition of bool in the public header, we won't have the field-offset mismatch when integrating with Postgres because the boolean fields in the public `splinterdb_config{}` have been all changed to `_Bool`.

So irrespective of whether struct is compiled with our library of PG-glue-code, we should see the same field definition.

gapisback:
@rtjohnso - I amended the code to use `_Bool` for few methods whose protos are in `public_util.h` . This allows me relocate `typedef bool int32` to `platform.h` and to go through compilation.

(Removed the newly added test-case.)

Let me know if you find this new version of the fix acceptable.

rtjohnso (approved):
Ship it!

PR #552, opened 2023-03-21
Core changes to support running Splinter with allocated shared memory.

gapisback:
This commit brings in basic support to create a shared memory segment and to redirect all memory allocation primitives to shared memory. Currently, we only support a simplistic memory mgmt; i.e. only-allocs, and no frees. With shared segments of 1-2 GiB we can run many functional and unit tests.

Some cavaets and fixes implemented to get some tests to run cleanly:

o Redirect large allocations to use malloc(), and not the shmem allocation.

  We currently do not support free() for shared memory based allocation.
  Some sub-systems that allocate and free large chunks repeatedly use
  up too much shared memory, causing tests to fail  .

  This commit fixes that issue by re-directing following clients to
  fall-back to using regular malloc() / free() interfaces:

  - merge_iterator_create()
  - merge_iterator_destroy(),
  - trunk_range() - trunk_range_iterator
  - trunk_range_iterator_deinit()
  - btree_pack_req_init()
  - btree_pack_req_deinit()

  With these changes, unit-test splinter_test now succeeds using shared memory:

sdb-fdb-build:[10] $ build/debug/bin/unit/splinter_test Running 1 CTests, suite name 'splinter', test case 'all'. TEST 1/3 splinter:test_inserts [OK]
TEST 2/3 splinter:test_lookups [OK]
TEST 3/3 splinter:test_splinter_print_diags [OK]
RESULTS: 3 tests (3 ok, 0 failed, 0 skipped) ran in 1136485 ms

o Rework existing tests to support --use-shmem arg to run w/ shared segment.

  Apply --use-shmem as the very 1st arg for different tests so that they
  can now leverage shared memory support. Following tests are / should be
  working with this fix:

 - limitations_test
 - splinterdb_quick_test
 - writable_buffer_test
 - task_system_test
 - splinter_test
 - btree_test

With this argument, splinterdb_create_or_open() will now create a heap, i.e., allocate shared segment, if the calling test has not already done so.

NOTE: unit/btree_test still goes through shm-free code due to the use
      of scratch_buffer space in this test masquerading as heap-id.
      Investigate and resolve.

o Fix bugs & refactoring to enable unit stress-tests.

  Minor rearrangement of code and fixing bugs. This commit now
  enables running btree_test and splinterdb_stress_test unit-tests
  with --use-shmem option.

o Enhances memory alloc/free diags to print name of object.

  To better understand for which object memory allocation
  or free is occurring, this commit makes minor changes to
  caller-macros that allocate / free memory to
  also print object's name using STRINGIFY() on the parameter 'v'.

o Rework unit-testing, main.c, Makefile to support --use-shmem

  Change test.sh also to run unit-tests with --use-shmem

Fix build failures, shfmt issues and chmod +x test.sh issues.

Fix chmod perms to +x

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,7 +386,8 @@ $(foreach unit,$(UNIT_TESTBINS),$(eval $(call unit_test_self_dependency,$(unit))
 #
 # These will need to be fleshed out for filters, io subsystem, trunk,
 # etc. as we create mini unit test executables for those subsystems.
-PLATFORM_SYS = $(OBJDIR)/$(SRCDIR)/$(PLATFORM_DIR)/platform.o
+PLATFORM_SYS = $(OBJDIR)/$(SRCDIR)/$(PLATFORM_DIR)/platform.o \
+			   $(OBJDIR)/$(SRCDIR)/$(PLATFORM_DIR)/shmem.o
 
 PLATFORM_IO_SYS = $(OBJDIR)/$(SRCDIR)/$(PLATFORM_DIR)/laio.o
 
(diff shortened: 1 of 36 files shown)


gapisback:
Incorrect target of PR. Accidentally fired this off to pull into `/main`, whereas it should have been a PR into `/shmem`. 

Nuke this.

PR #557, opened 2023-03-23
CI: use gcc for ASAN jobs

rosenhouse:
This change has already been applied to CI, so we can see the results of it live.


gapisback (approved):
Looks good ... Thanks for the quick turnaround !

PR #563, opened 2023-04-05
Fix bug in output formatted by size_to_str() helper.

gapisback:
Fractional portion of value formatted by size_to_str() was incorrect. We were losing the scale for things which were supposed to be "xx.07", we were reporting "xx.7", which is incorrect.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -407,7 +407,7 @@ size_to_str(char *outbuf, size_t outbuflen, size_t size)
    }
 
    if (frac_val || is_approx) {
-      snprintf(outbuf, outbuflen, "~%ld.%ld %s", unit_val, frac_val, units);
+      snprintf(outbuf, outbuflen, "~%ld.%02ld %s", unit_val, frac_val, units);
    } else {
       snprintf(outbuf, outbuflen, "%ld %s", unit_val, units);
    }
--- a/tests/unit/misc_test.c
+++ b/tests/unit/misc_test.c
@@ -244,13 +244,19 @@ CTEST2(misc, test_size_to_str)
 
         , { KiB + 128                      , "~1.12 KiB"  }
         , { KiB + ((25 * KiB) / 100)       , "~1.25 KiB"  }
-        , { MiB + 128                      , "~1.0 MiB"   }
+        , { MiB + 128                      , "~1.00 MiB"   }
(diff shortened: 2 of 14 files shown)


gapisback:
This was fortunately recognized and reported by @rosenhouse, in an [internal slack thread](https://vmware.slack.com/archives/D02D7UWU9U3/p1680641446893019):

Really small thing, but looking at the “Shared memory usage stats”, I see:

```
Shared memory usage stats shmid=27: Total=2222981120 bytes (~2.7 GiB), Used=1527320576 bytes (~1.42 GiB, 68.71 %), Free=695654272 bytes (~663.42 MiB, 31.29 %), nfrees=202000, nf_search_skipped=197988 (98 %), Large fragments in-use HWM=3616 (found in-use=0), consumed=1512585728 bytes (~1.40 GiB).
```

where that first Total bit `Total=2222981120 bytes (~2.7 GiB)` looks wrong: `2222981120 / 2^30` is about `2.07`, not `2.7`

src/util.c
```diff
@@ -407,7 +407,7 @@ size_to_str(char *outbuf, size_t outbuflen, size_t size)
    }
 
    if (frac_val || is_approx) {
-      snprintf(outbuf, outbuflen, "~%ld.%ld %s", unit_val, frac_val, units);
+      snprintf(outbuf, outbuflen, "~%ld.%02ld %s", unit_val, frac_val, units);
```
gapisback, line 410: Here is the bug. It's a formatting issue. We want 2-digit scale, but old code will only print "7" rather than "07".

With fix, you will now see "07" when appropriate.

tests/unit/misc_test.c
```diff
@@ -244,13 +244,19 @@ CTEST2(misc, test_size_to_str)
 
         , { KiB + 128                      , "~1.12 KiB"  }
         , { KiB + ((25 * KiB) / 100)       , "~1.25 KiB"  }
-        , { MiB + 128                      , "~1.0 MiB"   }
+        , { MiB + 128                      , "~1.00 MiB"   }
```
gapisback, line 247: With fix, decimals like "XX.003" will appear like "XX.00", hence the need to adjust expected output slightly.

rtjohnso (approved):
LGTM

PR #564, opened 2023-04-10
Rework fingerprint array as derived object from memfrag{} object.

gapisback:
Refactor fingerprint accessor / code. Fix bugs. Rework / refactor stats-usage structs. Stabilization fixes.


gapisback:
This whole fingerprint array memory mgmt design has been re-implemented and productized.

PR #569, which is under review now, carries forward the final code changes for this sub-system.

This dev work is no longer needed.

PR #565, opened 2023-04-12
Stabilization changes for platform_free() and free-list fragment management.

gapisback:


gapisback:
These dev change-set changes have been applied to the final version of part-3 support for shared memory. 

PR #569 is now ready for review, so this temp dev work can now be deleted.

PR #568, opened 2023-04-14
Support for multi-process, forked processes using shared memory

gapisback:
This commit extends core shared memory support to now allow for
a multi-process execution model, where multiple processes can now
attach to Splinter shared memory. Core thread-specific
concurrency primitives are modified, slightly, to now also support
a multi-process execution model.

- This commit sets up the stage to support fork()'ed or other OS-processes
  running with --use-shmem option, where each process will [in future]
  masquerade as a Splinter thread. A core change needed to move to that
  execution model is to support thread-specific IO-context structures.
  Otherwise, if an/other OS-process tries to do IO using AIO-context
  established by the main thread (i.e. by the process that started up
  SplinterDB), we will immediately run into hard IO-system call errors.

  This commit:
   - Performs an io_setup() for each thread / process
   - Manages this AIO-context tightly bound to Splinter's thread context
   - Does required book-keeping to keep this IO-context state kosher
     in conjunction with thread registration / de-registration.
   - Updates existing io_apis_test to deal with thread-specific IO
     context handles.

   An alternative could be localize this change-in-behaviour (of setting
   up thread-specific IO-context structs) only when the process-model of
   execution comes around. That execution model requires configuring
   SplinterDB with shared-memory support. But, just by looking at
   --use-shmem (or corresponding config setting), we cannot be sure that
   the process-model will be used or if we are just re-running rest of
   the test suites with shared-segment enabled. So, without trying to
   further complicate this choice-making, with this commit we will always
   set up thread-specific AIO-context structures, whenever shared memory
   configuration is detected.

Collection of lower-level changes to move to this execution model:

 - platform_buffer_init() that mmap()s' memory for the buffer cache
   will now use MAP_SHARED (v/s MAP_PRIVATE).
   The issue is that some parts of structures, e.g. buffer cache, are
   allocated using mmap(). The flags for this were MAP_PRIVATE, which
   means this memory is only accessible to the main process that set up
   Splinter. All child threads work on a COW-version of this mapped memory.
   So the changes done by the child process to the BTree in the buffer
   cache are not visible to the parent process.

 - Convert synchronization primitives to be shared across processes.

   This commit reworks core synchronization APIs to use interfaces
   that allow the sync-hook across child processes. This affects:

     - platform_mutex_init()
     - platform_spinlock_init()
     - platform_condvar_init()
     - platform_semaphore_init()
     - Add corresponding API-exerciser unit-tests for sanity coverage

 - Now that we have thread-specific IO-context setup, as part of
   thread register / deregister, we now also do io_register_thread(),
   io_deregister_thread(). This is basically book-keeping state of
   thread w.r.t IO setup & context.

Testing changes added:

  - Support --fork-child to test execution options. Some new tests
    will honor this argument, and will exercise activity using
    a forked-process execution model.

  - New test splinterdb_forked_child_test added: This covers the
    cases to show that IO errors could be repro'ed when running Splinter
    activity from a forked child process. Many other cases are added
    to this framework to exercise different cases of forked process
    doing SplinterDB activity. Much code/dev stabilization was achieved
    through this single new test.

  - Add case test_seq_key_seq_values_inserts_forked to large_inserts_stress
    test.

  - Existing functional io_apis_test to run with --fork-child option,
    thereby creating the scenario(s) of forked processes exercising
    the basic IO APIs.

  - Add new & extended tests to test.sh, for extended coverage using
    shared-memory and multi-process execution.

  - Add support for --wait-for-gdb and wait_for_gdb_hook() function.

    To debug forked child processes, add support for new command-line
    flag: --wait-for-gdb . And add a looping function where we can
    set a breakpoint, wait_for_gdb_hook(). Use this facility in
    splinterdb_forked_child_test.c, which has helped debug errors
    seen while running test_multiple_forked_process_doing_IOs().


src/btree.c
```diff
@@ -1778,7 +1778,23 @@ btree_insert(cache              *cc,         // IN
...
-   debug_assert(allocator_page_valid(cache_get_allocator(cc), child_node.addr));
+#if SPLINTER_DEBUG
+   bool child_node_is_valid =
+      allocator_page_valid(cache_get_allocator(cc), child_node.addr);
+   if (!child_node_is_valid) {
+      btree_print_tree(Platform_default_log_handle,
+                       cc,
+                       (btree_config *)cfg,
+                       parent_node.addr,
+                       PAGE_TYPE_MEMTABLE);
```
gapisback, line 1789: I had to deep-debug this area before I could id the fix for `mmap()` control flags. 

The new test case `test_one_insert_then_close_bug` in `splinterdb_forked_child_test.c` goes towards identifying this bug, where this debug code came very much in handy.

src/io.h
```diff
@@ -58,9 +58,11 @@ typedef platform_status (*io_write_async_fn)(io_handle     *io,
                                              uint64         addr);
 typedef void (*io_cleanup_fn)(io_handle *io, uint64 count);
 typedef void (*io_cleanup_all_fn)(io_handle *io);
-typedef void (*io_thread_register_fn)(io_handle *io);
+typedef void (*io_register_thread_fn)(io_handle *io);
+typedef void (*io_deregister_thread_fn)(io_handle *io);
```
gapisback, line 62: Minor renaming of interface functions for naming consistency.

src/platform_linux/laio.c
```diff
@@ -67,40 +70,140 @@ laio_cleanup(io_handle *ioh, uint64 count);
...
+   .deregister_thread = laio_deregister_thread,
+   .get_context       = laio_get_context,
+   .get_io_async_req  = laio_get_io_async_req,
 };
 
+/*
+ * Given an AIO handle, set up a thread-specific IO context opaque handle.
+ */
+static platform_status
+io_context_setup(laio_handle *io)
```
gapisback, line 105: New thread-specific setup routine. Review carefully.

src/platform_linux/laio.c
```diff
@@ -67,40 +70,140 @@ laio_cleanup(io_handle *ioh, uint64 count);
...
+ *
+ * This routine will also be called when Splinter is being shutdown as
+ * part of deinitializing the IO system. For that usage, allow for a case when
+ * some threads' IO-context may not have been cleaned up previously; so be
+ * a bit little lax on the assertion checks.
+ *
+ * Returns: 0, for successful destroy; 1, otherwise.
+ */
+static int
+io_context_cleanup(laio_handle *io, threadid tid, bool from_deregister)
```
gapisback, line 143: Again, needs close review.

src/platform_linux/laio.c
```diff
@@ -378,34 +518,58 @@ laio_cleanup(io_handle *ioh, uint64 count)
...
 
+   threadid tid = platform_get_tid();
+
    io = (laio_handle *)ioh;
 
    // Check for completion of up to 'count' events, one event at a time.
    // Or, check for all outstanding events (count == 0)
    for (i = 0; ((count == 0) || (i < count)); i++) {
-      status = io_getevents(io->ctx, 0, 1, &event, NULL);
+      status = io_getevents(io->ctx[tid], 0, 1, &event, NULL);
```
gapisback, line 528: Each thread will now poll for the completion of its IO events.

@ajhconway -- it's subtle stuff like this that will need close examination. Will be good if you can dwell into this thread-specific IO context re-design, to see if you can spot any gaping holdes.

src/platform_linux/laio.c
```diff
@@ -378,34 +518,58 @@ laio_cleanup(io_handle *ioh, uint64 count)
...
+
+         // We got a hard-error probably because some code-flow messed-up
+         // using the per-thread IO-context. No point in trying again and
+         // again if we are checking for completion of all outstanding
+         // events.
+         if (count == 0) {
+            break;
+         } else {
+            // Retry a few times for a finite # of pending events
+            continue;
```
gapisback, line 550: NOTE to myself: I think this re-try logic can come out now. I had added it during code-dev stabilization. This area is quite firm now.

Comments?

src/platform_linux/laio.c
```diff
@@ -378,34 +518,58 @@ laio_cleanup(io_handle *ioh, uint64 count)
...
  * laio_cleanup_all() - Handle completion of outstanding IO requests,
  * for all async requests in the queue.
+ *
+ * RESOLVE: Note this interface is potentially problematic when deployed
+ * using a process model. The expectation is that only the parent process
+ * will invoke this to do a final cleanup. But if there are any pending
+ * aios left behind by a process that exited before deregistering itself
+ * cleanly, we may find that AIO-request busy below, and will try to call
+ * the cleanup function. As the IO-context is no longer correct, we will
+ * get a hard-error from io_getevents() call.
```
gapisback, line 572: ATTN: Let's discuss this point ... I had some misgivings in this area, so put down this comment.

We should talk ... to see if there are any latent problems lurking here ... ???

src/platform_linux/platform.c
```diff
@@ -124,8 +153,17 @@ platform_buffer_init(buffer_handle *bh, size_t length)
...
                             length,
                             strerror(errno));
-         munmap(bh->addr, length);
+         // Save off rc from mlock()-failure, so we return that as status
          rc = CONST_STATUS(errno);
+
+         int rv = munmap(bh->addr, length);
+         if (rv != 0) {
+            // We are losing rc from this failure; can't return 2 statuses
+            platform_error_log("munmap %p (%lu bytes) failed with error: %s\n",
```
gapisback, line 162: Potential missing `rc` issue also in `/main`. Hard to repro ... figured this out from code-walk through.

src/platform_linux/platform.c
```diff
@@ -211,11 +249,28 @@ platform_thread_id_self()
...
+   pthread_mutexattr_init(&mattr);
+   status.r = pthread_mutexattr_setpshared(&mattr, PTHREAD_PROCESS_SHARED);
+   if (!SUCCESS(status)) {
+      return status;
+   }
+
+   // clang-format off
+   status.r = pthread_mutex_init(&lock->mutex,
+                                 ((heap_id == PROCESS_PRIVATE_HEAP_ID)
+                                    ? NULL : &mattr));
```
gapisback, line 266: Here onwards, collection of changes to migrate thread-specific synchronization logic to also work in multi-process model.

src/platform_linux/platform.c
```diff
@@ -329,51 +472,6 @@ platform_sort_slow(void               *base,
    return qsort_r(base, nmemb, size, cmpfn, cmparg);
 }
 
-platform_status
-platform_condvar_init(platform_condvar *cv, platform_heap_id heap_id)
-{
```
gapisback, line 304: This chunk moved up, to be located alongside their sibling other-sync mechanism functions.

src/platform_linux/shmem.c
```diff
@@ -224,17 +226,93 @@ platform_valid_addr_in_shm(platform_heap_handle heap_handle, void *addr)
...
 
 /*
- * platform_valid_addr_in_heap(), platform_valid_addr_in_shm()
- *
  * Validate that input address 'addr' is a valid address within shared segment
  * region.
  */
-static inline bool
-platform_valid_addr_in_heap(platform_heap_id heap_id, void *addr)
+bool
```
gapisback, line 232: In part-2, the isn't much real logic change in `shmem.c`, other than adding more tracing, testing hooks and better metrics.

gapisback:
**NOTE: to the reviewers: @rtjohnso @ajhconway @rosenhouse :**

This is part-2 of the shared memory support dev work: Support for multi-process, forked-processes executing with shared memory

I have layered this on top of the `aguraajda/shmem-core-Rev`, so you will see  2 commits. ONLY review the diffs in the [2nd commit](https://github.com/vmware/splinterdb/pull/568/commits/aff67bde838648d7ff1f20f8ec1cad748c46681d), as the 1st commit will be reviewed under part-1's PR #556 

Here is the order to review files in, to get a good grip on this change-set:

1. `laio.h`, `laio.c` - see my annotations
2. Read / Write APIs tests in `functional/io_apis_test.c` have been enhanced to run in `--fork-child` execution mode, which exercises all this IO sub-system code in its full glory. I'm pretty comfortable overall with these changes.
3. `task.c` - see my annotations
4. `shmem.c` - In part-2, the isn't much real logic change , other than adding more tracing, testing hooks and better metrics. You can give this a pass or just a quick-look.
5. Tests: Review or take a look at `functional/io_apis_test.c`, `unit/splinterdb_forked_child_test.c`.
6. `test.sh` - Review the new tests being folded-in as part of normal CI runs.

src/platform_linux/platform.h
```diff
@@ -669,6 +671,12 @@ platform_heap_create(platform_module_id module_id,
 void
 platform_heap_destroy(platform_heap_id *heap_id);
 
+void
+platform_shm_set_splinterdb_handle(platform_heap_id heap_id, void *addr);
+
+shmem_heap *
+platform_heap_id_to_shmaddr(platform_heap_id hid);
```
gapisback, line 678: Add extern references to these interfaces which are used [mainly] in unit-tests.

src/platform_linux/platform.h
```diff
@@ -711,16 +719,6 @@ platform_thread_id_self();
...
 platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
 
-void
-platform_enable_tracing_shm_ops();
-
-void
-platform_enable_tracing_shm_allocs();
-
-void
-platform_enable_tracing_shm_frees();
```
gapisback, line 721: Deleted and moved to `shmem.h`

tests/splinterdb_test_apis.h
```diff
@@ -17,8 +17,8 @@
 void
 splinterdb_cache_flush(const splinterdb *kvs);
 
-void *
-splinterdb_get_heap_handle(const splinterdb *kvs);
+platform_heap_id
+splinterdb_get_heap_id(const splinterdb *kvs);
```
gapisback, line 21: As `heap_handle` has been obsoleted as part of previous commit, rework this interface to now return `heap_id`.

tests/unit/splinter_shmem_test.c
```diff
@@ -578,37 +580,6 @@ CTEST2(splinter_shmem, test_realloc_of_free_fragments_uses_first_fit)
...
-/*
- * ---------------------------------------------------------------------------
- * Test case to verify that configuration checks that shared segment size
- * is "big enough" to allocate memory for RC-allocator cache's lookup
- * array. For very large devices, with insufficiently sized shared memory
- * config, we will not be able to boot-up.
- * RESOLVE - This test case is still incomplete.
- * ---------------------------------------------------------------------------
- */
-CTEST2(splinter_shmem, test_large_dev_with_small_shmem_error_handling)
```
gapisback, line 590: Deleted this test case as it's too low-level detail. (And a bit time consuming to hammer out the logic to exercise this scenario. Not very important right now.)

gapisback:
@rtjohnso - I am assigning this PR to you for review. All tests have passed in previous runs, but after the most recent cleanup change (to engage some new unit-tests), one `splinter_test --perf --use-shmem ` [job](https://runway-ci.eng.vmware.com/builds/3591606670) has failed as follows:

```
+ build/release/bin/driver_test splinter_test --perf --use-shmem --max-async-inflight 0 --num-insert-threads 4 --num-lookup-threads 4 --num-range-lookup-threads 0 --tree-size-gib 2 --cache-capacity-mib 512
build/release/bin/driver_test: splinterdb_build_version 4876ae59
Dispatch test splinter_test
[...]
splinter_perf_inserts() starting num_insert_threads=4, num_threads=4, num_inserts=27185152 (~27 million) ...
Thread 2 inserting  34% complete for table 0 ... OS-pid=1084, OS-tid=1085, Thread-ID=1, Assertion failed at src/routing_filter.c:184:routing_get_header(): "hdr_raw_addr != 0". 
./test.sh: line 110:  1084 Aborted                 "$@"
```

I recall having seen such failures previously, too, even without shared memory configured, but could not spot an identical open issue reporting this failure.

I am re-running this job, just to see if will succeed. I suspect there is a lurking issue that may randomly pop-up.

I suggest you do start on the review while I try to figure this out in the background.

gapisback:
HI, @rtjohnso - Update on testing status: My [re-run](https://runway-ci.eng.vmware.com/teams/splinterdb/pipelines/splinterdb/jobs/main-pr-gcc/builds/975.1) of this failed CI/gcc job now succeeded.

As suspected there is some flaky condition that seems to pop-up occasionally.

src/platform_linux/laio.c
gapisback: See comments added to the files requiring changes.

src/platform_linux/laio.c
gapisback: @rtjohnso - Remove this new changed behaviour and revert to the old code flow.

include/splinterdb/splinterdb.h
```diff
@@ -60,7 +60,8 @@ typedef struct splinterdb_config {
 
    // Shared memory support
    uint64 shmem_size;
-   _Bool  use_shmem; // Default is FALSE.
+   _Bool  use_shmem;  // Default is FALSE.
+   _Bool  fork_child; // Default is FALSE
```
gapisback, line 64: re-visit -- whether this field is needed any more in this global struct. Delete, if not needed.

src/platform_linux/platform_inline.h
```diff
@@ -110,12 +110,21 @@ platform_semaphore_destroy(platform_semaphore *sema)
...
-{
-   __attribute__((unused)) int err = sem_init(sema, 0, value);
+                        platform_heap_id    heap_id)
+{
+   // If we are running with a shared segment, it's likely that we
+   // may also fork child processes attaching to Splinter's shmem.
+   // Then, use 1 => spinlocks are shared across process boundaries.
+   // Else, use 0 => spinlocks are shared between threads in a process.
+   __attribute__((unused)) int err =
+      sem_init(sema, ((heap_id == PROCESS_PRIVATE_HEAP_ID) ? 0 : 1), value);
```
gapisback, line 127: @rtjohnso - Investigate why we had to this business of 0 / 1. Can't we always initialize sem_init() the same way, irrespective of whether it's process or threaded-exec model.

src/splinterdb.c
```diff
@@ -785,3 +789,57 @@ splinterdb_close_print_stats(splinterdb *kvs)
...
 }
+
+/*
+ * -------------------------------------------------------------------------
+ * External "APIs" provided mainly to invoke lower-level functions intended
+ * for use -ONLY- as testing interfaces.
+ * -------------------------------------------------------------------------
+ */
+void
+splinterdb_cache_flush(const splinterdb *kvs)
```
gapisback, line 800: @rtjohnso : Review why this externalization is needed.

And what is the intent behind the usage of this interface in the newly added `test_completion_of_outstanding_async_IOs_from_process_bug()` test case.

gapisback (commented):
Address comments and circle back with RobJ.

src/platform_linux/laio.c
gapisback: Done.

src/platform_linux/laio.c
```diff
@@ -67,37 +70,137 @@ laio_cleanup(io_handle *ioh, uint64 count);
...
+   .get_async_req     = laio_get_async_req,
+   .get_metadata      = laio_get_metadata,
+   .read_async        = laio_read_async,
+   .write_async       = laio_write_async,
+   .cleanup           = laio_cleanup,
+   .cleanup_all       = laio_cleanup_all,
+   .register_thread   = laio_register_thread,
+   .deregister_thread = laio_deregister_thread,
+   .get_context       = laio_get_context,
+   .get_io_async_req  = laio_get_io_async_req,
```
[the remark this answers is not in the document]
gapisback: Done. `io_get_io_async_req()` and all its references and IO->fn ptrs are now withdrawn.

include/splinterdb/splinterdb.h
gapisback: Good catch, @rtjohnso - yes, this can come out. Deleted.

src/platform_linux/platform_inline.h
gapisback: @rtjohnso : Went back and took a re-look at the [man page](https://man7.org/linux/man-pages/man3/sem_init.3.html) mentioned a few lines above in the comment. 

The documented interface & its semantic is:

```
If pshared has the value 0, then the semaphore is shared between
       the threads of a process, and should be located at some address
       that is visible to all threads (e.g., a global variable, or a
       variable allocated dynamically on the heap).

       If pshared is nonzero, then the semaphore is shared between
       processes, and should be located in a region of shared memory.
```

So, no, doesn't seem like we can unconditionally always use either 0 or 1, irrespective of threaded- v/s process-execution model.

src/splinterdb.c
gapisback: Revisited both newly added test case:

- `test_completion_of_outstanding_async_IOs_from_process_bug()`
- `test_multiple_forked_process_doing_IOs()`

The point of using `cache_flush()` is so that we can invoke the call to `io_cleanup_all()` -> `io_cleanup()` to complete all pending AIOs.

Both test cases essentially are exercisers for these interfaces to cross-check the behaviour of whether we can issue any IO-cleanup related calls after thread deregistration.

The tests are somewhat of a negative test in that they verify an expected code-flow of execution.

These test cases seem useful as a way of self-documenting stabilization test cases, hence, I'd like to retain these.

src/platform_linux/laio.c
gapisback: @rtjohnso and I walked through the logic changes during our review discussion and agreed to the nature of these changes.

All comments added to clean-up code raised during this review discussion have since been addressed, with code rework.

src/platform_linux/laio.c
gapisback: @rtjohnso and I walked through the logic changes during our review discussion and agreed to the nature of these changes.

src/platform_linux/shmem.c
```diff
@@ -169,12 +169,13 @@ platform_shm_heap_valid(shmem_heap *shmheap);
...
  * location of the next-free-byte can be tracked.
  */
-static inline shmem_heap *
+shmem_heap *
 platform_heap_id_to_shmaddr(platform_heap_id hid)
 {
    debug_assert(hid != NULL);
-   void *shmaddr = ((void *)hid - offsetof(shmem_heap, shm_id));
-   return (shmem_heap *)shmaddr;
+   shmem_heap *shmaddr =
```
rtjohnso, line 176: I don't understand why you don't just make the hid for a shared-mem heap to be equal to the address of its control block.

In other words, why isn't this code just
```
   return (shmem_heap *)hid;
```

src/routing_filter.c
```diff
@@ -8,7 +8,7 @@
  *     This file contains the implementation for a routing filter
  *----------------------------------------------------------------------
  */
-
+#include <unistd.h>
```
rtjohnso, line 11: Platformize the functionality you need here.

src/io.h
```diff
@@ -58,9 +58,12 @@ typedef platform_status (*io_write_async_fn)(io_handle     *io,
...
 typedef bool32 (*io_max_latency_elapsed_fn)(io_handle *io, timestamp ts);
+
 typedef void *(*io_get_context_fn)(io_handle *io);
+typedef io_async_req *(*io_get_io_async_req_fn)(io_handle *io);
```
rtjohnso, line 66: Seems to be unused.

src/splinterdb.c
```diff
@@ -785,3 +789,57 @@ splinterdb_close_print_stats(splinterdb *kvs)
...
+   cache_flush(kvs->spl->cc);
+}
+
+platform_heap_id
+splinterdb_get_heap_id(const splinterdb *kvs)
+{
+   return kvs->heap_id;
+}
+
+const void *
```
rtjohnso, line 811: Why are these `void *` instead of the correct type?

tests/config.c
```diff
@@ -370,13 +371,30 @@ config_parse(master_config *cfg, const uint8 num_config, int argc, char *argv[])
...
+            }
+         }
 
          config_set_uint64("key-size", cfg, max_key_size) {}
          config_set_uint64("data-size", cfg, message_size) {}
 
          // Test-execution configuration parameters
          config_set_uint64("seed", cfg, seed) {}
          config_set_uint64("num-inserts", cfg, num_inserts) {}
+         config_set_uint64("num-forked-processes", cfg, num_forked_processes) {}
```
rtjohnso, line 397: I think "num-processes" would be better.

tests/splinterdb_test_apis.h
```diff
@@ -0,0 +1,45 @@
...
+#include "splinterdb/splinterdb.h"
+
+// External APIs provided -ONLY- for use as a testing hook.
+void
+splinterdb_cache_flush(const splinterdb *kvs);
+
+platform_heap_id
+splinterdb_get_heap_id(const splinterdb *kvs);
+
+const void *
```
rtjohnso, line 23: Move this file to `src/splinterdb_private.h` and `#include` it in `splinterdb.c` so that the decls get type-checked against the definitions.

rtjohnso (changes requested):
Mostly minor cleanups.

src/platform_linux/platform.c
```diff
@@ -395,6 +430,88 @@ platform_batch_rwlock_unget(platform_batch_rwlock *lock, uint64 lock_idx)
...
+   if (!SUCCESS(status)) {
+      return status;
+   }
+
+   // clang-format off
+   status.r = pthread_mutex_init(&cv->lock,
+                                ((heap_id == PROCESS_PRIVATE_HEAP_ID)
+                                    ? NULL : &mattr));
+   // clang-format on
+   __attribute__((unused)) int rv = pthread_mutexattr_destroy(&mattr);
```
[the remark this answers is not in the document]
gapisback: Done. Changed to use `debug_only` and that seems to work fine.

src/platform_linux/platform.c
```diff
@@ -395,6 +430,88 @@ platform_batch_rwlock_unget(platform_batch_rwlock *lock, uint64 lock_idx)
...
+   if (!SUCCESS(status)) {
+      return status;
+   }
+
+   // Init condition so it can be shared between processes, if so configured
+   pthread_condattr_t cattr;
+   pthread_condattr_init(&cattr);
+   status.r = pthread_condattr_setpshared(&cattr, PTHREAD_PROCESS_SHARED);
+   if (!SUCCESS(status)) {
+      return status;
```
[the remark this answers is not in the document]
gapisback: Reworked as suggested to use `goto` labels, which also brings-in the missing cleanup for `pthread_mutex_destroy(&cv->lock)` in one error code path.

src/platform_linux/shmem.c
gapisback: OK; Done. This needed a small fix in `platform_shmcreate()`, to return the heap_id correctly when splinter is first created. That new code now looks like:

```
 331 platform_status
 332 platform_shmcreate(size_t            size,
 333                    platform_heap_id *heap_id) // Out
[...]
 377    // Return 'heap-ID' handle pointing to start addr of shared segment.
 378    if (heap_id) {
 379       *heap_id = (platform_heap_id *)shmaddr;
 380    }
```

Unit-tests seem to pass with this new approach. Will watch CI ...

src/io.h
gapisback: yes, right. Deleted.

src/splinterdb.c
gapisback: These accessor methods are used in the `unit/splinterdb_forked_child_test.c` test, to validate that the sub-system's handle address is valid.

We cross-check using `platform_valid_addr_in_shm()`, which just needs a `void *` ptr of the address being validated.

If each of these `splinterdb_get...()` methods were to return the correct type, it will require further casting when calling the validation function named above.

As these interfaces are meant only for testing purposes, I felt it was acceptable just to return `void *`.

tests/config.c
gapisback: Changed as suggested.

Also, discovered that there was no invocation of a test with this argument. ... Good!

Adjusted `test.sh/run_slower_forked_process_tests()`, to add in a call to `unit/splinterdb_forked_child_test` with this new `--num-processes` argument, for improved coverage (of this new config option).

tests/splinterdb_test_apis.h
gapisback: Done, with one minor diff. I wanted to retain the fact that this `.h` exists solely for use by tests, so renamed this file to `src/splinterdb_tests_private.h`.

src/routing_filter.c
gapisback: This #include seems like left-over cruft. 

I went back and took a look at the diffs introduced, which is just an elaborated print statement.

I don't need L11 anymore, so took it out.  Dev builds seem to work ok, will cross-check in CI.

Thanks for catching this ...

gapisback: @rtjohnso --  I had to re-introduce this `#include`. It's needed to resolve extern reference to `getpid()` that is used in the error message. Otherwise, CI quick_check job failed in clang debug build mode. 

[ Strangely, I can't repro that failure on my Linux dev box in the exact same build mode. ]

This isn't new with this change-set. Other files are also #include'ing `unistd.h` whenever `getpid()` is used.

src/splinterdb.c
rtjohnso: Would you mind changing them anyway?  I feel strongly about this.

src/routing_filter.c
rtjohnso: I think all the other instances of `#include <unistd.h>` were introduced by you in the previous PR for multi-process support.

We need to clean them all out by platformizing `getpid`.

I'd prefer to do it now, but I understand that doing it now may create additional difficulty in rebasing and merging the third PR in this line of work.

Would you mind doing it as a follow-on PR after we land the third one?

rtjohnso:
The latest code changes look good.

However, note my two comments on the `void *` accessor methods and the `#include <unistd.h>` stuff.  Both of those need to be fixed, although the latter could occur later if it would cause difficulty merging the third multiprocess PR.

src/routing_filter.c
gapisback: @rtjohnso - I peeled this cleanup item out on its own new issue. See issue #599 where I've described a possible solution.

I will clean this up after part-3 lands.

src/splinterdb.c
gapisback: @rtjohnso -- Changed as requested. Please see the delta-diffs in [this commit](https://github.com/vmware/splinterdb/pull/568/commits/1795b6acaa91b0841c6ab57c766d7d4a1b3bb9ef).

Curiously, I did not seem to have to change the callers to cast the return ptr to `(void *)`. Example, this one in `splinterdb_forked_child_test.c`:

```
152       ASSERT_TRUE(platform_valid_addr_in_heap(
153          spl_heap_id, splinterdb_get_task_system_handle(spl_handle)));
```

I would have expected to have to cast the return to `void *`, but I didn't get any compilation errors in my dev builds. Will see how things fare in CI.

For proper-form, I could do the casting, but seems unnecessary as compilation is not failing.
Let me know if you feel strongly that the call-sites should be amended, too.

gapisback:
@rtjohnso -- I've made the one code change you requested (see [this commit](https://github.com/vmware/splinterdb/pull/568/commits/1795b6acaa91b0841c6ab57c766d7d4a1b3bb9ef)), and peeled-off new issue #599 to platformize the use of `getpid()`.

Let me know if you see need for any further changes.

src/splinterdb_tests_private.h
```diff
@@ -20,26 +24,20 @@ splinterdb_cache_flush(const splinterdb *kvs);
...
-const void *
+const task_system *
 splinterdb_get_task_system_handle(const splinterdb *kvs);
 
-const void *
+const platform_io_handle *
 splinterdb_get_io_handle(const splinterdb *kvs);
 
-const void *
+const rc_allocator *
```
rtjohnso, line 33: Actually, can you change these to return `cache *` and `allocator *` (i.e. the abstract types rather than the concrete implementations)?  Thank you!

rtjohnso (approved):
Note one minor change.  Sorry I wasn't clear on that before.

No need to check with me again.  Thanks!

src/splinterdb_tests_private.h
gapisback: Done in final round of changes. 

Fyi; @rtjohnso -- Final CI job is running ... Once it passes, I will push this to `/main`.

PR #569, opened 2023-04-14
Support free-fragment recycling in shared-segment. Add fingerprint object management.

gapisback:
The main change with this commit is the support for free-fragment lists
and recycling of small fragments from shared memory. This was a main
limitation of the support added in previous commits. Another driving
factor for implementing some free-list support was that previous
multi-user concurrent insert performance benchmarking was not functional
beyond a point, and we'd frequently run into shmem Out-Of-Memory (OOMs),
even with shmem sizes > 8 GiB (which worked in a prior dev/perf-test cycle).

The main design changes to manage small-fragments are follows:

**Managing memory allocation / free using platform_memfrag{} fragments:**

- Allocation and free of memory is dealt with in terms of "memory
  fragments", a small structure that holds the memory->{addr, size}.
  All memory requests (as is being done previously) are aligned to
  the cacheline.

  - Allocation: All clients of memory allocation have to "hand-in"
    an opaque platform_memfrag{} handle, which will be returned populated
    with the memory address, and more importantly, the size-of-the-fragment
    that was used to satisfy the memory request.

  - Free: Clients now have to safely keep a handle to this returned
    platform_memfrag{}, and hand it back to the free() method.
    free() will rely "totally" on the size specified in this input
    fragment handle supplied. And the free'd memory fragment will
    be returned to the corresponding free-list bucket.

- Upon free(), the freed-fragment is tracked in a few free-lists
  bucketed by size of the freed-fragment. For now, we support 4 buckets,
  size <= 64, <= 128, <= 256 & <= 512 bytes. (These sizes are sufficient
  for current benchmarking requirements.) A free'd fragment is hung off
  of the corresponding list, threading the free-fragments using
  the fragment's memory itself. New struct free_frag_hdr{} provides the
  threading structure. It tracks the current fragment's size and
  free_frag_next pointer. The 'size' provided to the free() call is
  is recorded as the free'd fragment's size.

- Subsequently, a new alloc() request is 1st satisfied by searching the
  free-list corresponding to the memory request. For example, a request
  from a client for 150 bytes will be rounded-up a cacheline boundary,
  i.e. 192 bytes. The free-list for bucket 256 bytes will be searched
  to find the 1st free-fragment of the right size. If no free fragment
  is found in the target list, we then allocate a new fragment.
  The returned fragment will have a size of 256 (for an original request
  of 150 bytes).

- An immediate consequence of this approach is that there is a small,
  but significant, change in the allocation, free APIs; i.e. TYPED_MALLOC(),
  TYPED_ARRAY_MALLOC() and TYPED_FLEXIBLE_STRUCT_MALLOC(), and their 'Z'
  equivalents, which return 0'ed out memory.

- All existing clients of the various TYPED*() memory allocation
  calls have been updated to declare an on-stack platform_memfrag{}
  handle, which is passed back to platform_free().

  - In some places memory is allocated to initialize sub-systems and
    then torn down during deinit(). In a few places existing structures
    are extended to track an additional 'size' field. The size of the
    memory fragment allocated during init() is recorded here, and then
    used to invoke platform_free() as part of the deinit() method.
    An example is clockcache_init() where this kind of work to record
    the 'size' of the fragment is done and passed-down to clockcache_deinit(),
    where the memory fragment is then freed with the right 'size'.
    This pattern is now to be seen in many such init()/deinit() methods
    of diff sub-systems; e.g. pcq_alloc(), pcq_free(), ...

- Cautionary Note:

  If the 'ptr' handed to platform_free() is not of type platform_memfrag{} *,
  it is treated as a generic <struct> *, and its sizeof() will be used
  as the 'size' of the fragment to free. This works in most cases. Except
  for some lapsed cases where, when allocating a structure, the allocator
  ended up selecting a "larger" fragment that just happened to be
  available in the free-list. The consequence is that we might end-up
  free'ing a larger fragment to a smaller-sized free-list. Or, even if
  we do free it to the right-sized bucket, we still end-up marking the
  free-fragment's size as smaller that what it really is. Over time, this
  may add up to a small memory leak, but hasn't been found to be crippling
  in current runs. (There is definitely no issue here with over-writing memory
  due to incorrect sizes.)

- Copious debug and platform asserts have been added in shmem alloc/free
  methods to cross-check to some extent illegal calls.

**Fingerprint Object Management:**

  Managing memory for fingerprint arrays was particularly problematic.
  This was the case even in a previous commit, before the introduction
  of the memfrag{} approach. Managing fingerprint memory was found to
  be especially cantankerous due to the way filter-building and compaction
  tasks are queued and asynchronously processed by some other
  thread / process.

  The requirements from the new interfaces are handled as follows:

   - Added a new fingerprint{} object, struct fp_hdr{}, which embeds
     at its head a platform_memfrag{}. And few other short fields are
     added for tracking fingerprint memory mgmt gyrations.

   - Various accessor methods are added to manage memory for fingerprint
     arrays through this object. E.g.,

     - fingerprint_init() - Will allocate required fingerprint for 'ntuples'.
     - fingerprint_deinit() - Will dismantle object and free the memory
     - fingerprint_start() - Returns start of fingerprint array's memory
     - fingerprint_nth() - Returns n'th element of fingerprint

  Packaging the handling of fingerprint array through this object and
  its interfaces helped greatly to stabilize the memory histrionics.

- When SplinterDB is closed, shared memory dismantling routine will tag
  any large-fragments that are still found "in-use". This is percolated
  all the way back to splinterdb_close(), unmount() and to
  platform_heap_destory() as a failure $rc. Test will fail if they have
  left some un-freed large fragments. (Similar approach was considered to
  book-keep all small fragments used/freed, but due to some rounding
  errors, it cannot be a reliable check at this time. So hasn't been done.)

Test changes:

**Miscellaneous**:

-  Elaborate and illustrative tracing added to track memory mgmt done for fingerprint arrays, especially when they are bounced around queued / re-queued tasks. (Was a very problematic debugging issue.)

-  Extended tests to exercise core memory allocation / free APIs, and to exercise fingerprint object mgmt, and writable_buffer interfaces:

    - platform_apis_test:
    - splinter_shmem_test.c: Adds specific test-cases to verify that
        free-list mgmt is happening correctly.

-  Enhanced various diagnostics, asserts, tracing

-  Improved memory usage stats gathering and reporting
- Added hooks to cross-check multiple-frees of fragments, and testing hooks to verify if a free'd fragment is relocated to the right free-list


include/splinterdb/splinterdb.h
```diff
@@ -179,7 +179,7 @@ splinterdb_open(splinterdb_config *cfg, splinterdb **kvs);
 // Close a splinterdb
 //
 // This will flush all data to disk and release all resources
-void
+int
```
gapisback, line 182: To percolate errors found by shm-destroy, if large-fragments not free are still found hanging around.

src/PackedArray.c
```diff
@@ -380,7 +380,6 @@ void PACKEDARRAY_JOIN(__PackedArray_unpack_, PACKEDARRAY_IMPL_BITS_PER_ITEM)(con
 #include "poison.h"
 
 #define PACKEDARRAY_MALLOC(size) platform_malloc(size)
-#define PACKEDARRAY_FREE(p) platform_free(p)
```
gapisback, line 383: Unused interface.

src/btree.c
```diff
@@ -2908,8 +2914,8 @@ btree_pack_loop(btree_pack_req *req,       // IN/OUT
    log_trace_key(tuple_key, "btree_pack_loop (bottom)");
 
    if (req->hash) {
-      platform_assert(req->num_tuples < req->max_tuples);
-      req->fingerprint_arr[req->num_tuples] =
+      platform_assert(btree_pack_can_fit_tuple(req));
+      fingerprint_start(&req->fingerprint)[req->num_tuples] =
```
gapisback, line 2918: Here's where you will start to see the use of fingerprint object and its accessor / interfaces.

src/btree.h
```diff
@@ -151,9 +151,9 @@ typedef struct btree_pack_req {
    btree_config *cfg;
    iterator     *itor; // the itor which is being packed
    uint64        max_tuples;
-   hash_fn       hash; // hash function used for calculating filter_hash
-   unsigned int  seed; // seed used for calculating filter_hash
-   uint32       *fingerprint_arr; // IN/OUT: hashes of the keys in the tree
+   hash_fn       hash;        // hash function used for calculating filter_hash
+   unsigned int  seed;        // seed used for calculating filter_hash
+   fp_hdr        fingerprint; // IN/OUT: hashes of the keys in the tree
```
gapisback, line 156: The in-place `char * array` is now replaced by the fingerprint object, which carries inside of it `platform_memfrag{}` handle to track allocate memory fragment's size and to free it reliably.

src/btree.h
```diff
@@ -341,26 +346,27 @@ btree_pack_req_init(btree_pack_req  *req,
    req->hash       = hash;
    req->seed       = seed;
    if (hash != NULL && max_tuples > 0) {
-      req->fingerprint_arr =
-         TYPED_ARRAY_ZALLOC(hid, req->fingerprint_arr, max_tuples);
+
+      fingerprint_init(&req->fingerprint, hid, max_tuples); // Allocates memory
```
gapisback, line 350: Inline memory allocation on old L345 is, henceforth, replaced by init()'ing the fingerprint object ... And so on ...

src/btree.h
```diff
@@ -341,26 +346,27 @@ btree_pack_req_init(btree_pack_req  *req,
...
 
       // When we run with shared-memory configured, we expect that it is sized
       // big-enough to not get OOMs from here. Hence, only a debug_assert().
-      debug_assert(req->fingerprint_arr,
+      debug_assert(!fingerprint_is_empty(&req->fingerprint),
                    "Unable to allocate memory for %lu tuples",
                    max_tuples);
-      if (!req->fingerprint_arr) {
+      if (fingerprint_is_empty(&req->fingerprint)) {
```
gapisback, line 357: You can no longer check for NULL array ptr, to detect OOM. You must consult the `is_empty()` method to figure out if there is memory or not.

src/btree.h
```diff
@@ -341,26 +346,27 @@ btree_pack_req_init(btree_pack_req  *req,
...
 }
 
+// Free memory if any was allocated for fingerprint array.
 static inline void
 btree_pack_req_deinit(btree_pack_req *req, platform_heap_id hid)
 {
-   if (req->fingerprint_arr) {
-      platform_free(hid, req->fingerprint_arr);
+   if (!fingerprint_is_empty(&req->fingerprint)) {
+      fingerprint_deinit(hid, &req->fingerprint);
```
gapisback, line 369: `deinit()` will free memory.

src/clockcache.c
```diff
@@ -1822,20 +1822,25 @@ clockcache_init(clockcache        *cc,   // OUT
...
    /* lookup maps addrs to entries, entry contains the entries themselves */
-   cc->lookup =
-      TYPED_ARRAY_MALLOC(cc->heap_id, cc->lookup, allocator_page_capacity);
+   platform_memfrag memfrag_cc_lookup;
+   cc->lookup = TYPED_ARRAY_MALLOC_MF(
+      cc->heap_id, cc->lookup, allocator_page_capacity, &memfrag_cc_lookup);
    if (!cc->lookup) {
       goto alloc_error;
    }
+   cc->lookup_size = memfrag_size(&memfrag_cc_lookup);
```
gapisback, line 1831: Here's the first instance of an init / deinit pair of calls, which now need to communicate the size of memory fragment allocated by init().
Like it's done on this line, few common structures now gain a new `size` field to track the memory fragment's size.

src/clockcache.c
```diff
@@ -1910,11 +1920,17 @@ clockcache_deinit(clockcache *cc) // IN/OUT
 #endif
    }
 
+   platform_memfrag  memfrag = {0};
+   platform_memfrag *mf      = &memfrag;
    if (cc->lookup) {
-      platform_free(cc->heap_id, cc->lookup);
+      memfrag_init_size(mf, cc->lookup, cc->lookup_size);
+      platform_free(cc->heap_id, mf);
```
gapisback, line 1927: `free()` needs to be told the fragment's size correctly. This is obtained from the `size` field stashed away when `init()` was done.

src/clockcache.c
```diff
@@ -1910,11 +1920,17 @@ clockcache_deinit(clockcache *cc) // IN/OUT
...
+      cc->lookup = NULL;
    }
    if (cc->entry) {
-      platform_free(cc->heap_id, cc->entry);
+      memfrag_init_size(mf, cc->entry, cc->entry_size);
+      platform_free(cc->heap_id, mf);
```
gapisback, line 1932: Same pattern of changes continues. This will appear in many more instances ...

src/memtable.c
```diff
@@ -331,7 +333,10 @@ memtable_context_destroy(platform_heap_id hid, memtable_context *ctxt)
    ref = allocator_dec_ref(al, ctxt->insert_lock_addr, PAGE_TYPE_LOCK_NO_DATA);
    platform_assert(ref == AL_FREE);
 
-   platform_free(hid, ctxt);
+   platform_memfrag  memfrag_ctxt;
+   platform_memfrag *mf = &memfrag_ctxt;
+   memfrag_init_size(mf, ctxt, ctxt->mt_ctxt_size);
+   platform_free(hid, mf);
```
gapisback, line 339: NOTE: This does look like verbose multi-line repeat code.

@rtjohnso - I did consider whether to add a packaged macro, say, `memfrag_init()`, to which you supply the addr / init. Inside the body, we can declare a hidden structures and `platform_memfrag *mf`, do the setup, and pass-it as the 2nd arg.

Can be done ... probably .. did not try it too hard. Wanted to get this into review, and then I expect anyway to get comments on this approach.

We can re-discuss the coding impact this approach has ... and re-visit during review.

src/platform_linux/laio.c
```diff
@@ -287,7 +291,11 @@ io_handle_deinit(laio_handle *io)
    }
    platform_assert(status == 0);
 
-   platform_free(io->heap_id, io->req);
+   platform_memfrag  memfrag = {.addr = io->req, .size = io->req_size};
```
gapisback, line 294: NOTE to myself: This should go away. Currently, `memfrag_init_size()` is a #define and this struct is exposed in `platform.h`. Rework this so the fields are hidden, and only `memfrag_init_size()` is exposed to client code.

This will prevent such naked assignments.

src/platform_linux/platform.c
```diff
@@ -84,13 +84,16 @@ platform_heap_create(platform_module_id    UNUSED_PARAM(module_id),
    return STATUS_OK;
 }
 
-void
+platform_status
```
gapisback, line 87: Percolating errors upstream from `platform_shmdestroy`; see L94 below.

src/platform_linux/platform.c
```diff
@@ -395,12 +398,14 @@ platform_histo_create(platform_heap_id       heap_id,
                       const int64 *const     bucket_limits,
                       platform_histo_handle *histo)
 {
+   platform_memfrag      memfrag_hh;
    platform_histo_handle hh;
-   hh = TYPED_MANUAL_MALLOC(
+   hh = TYPED_ARRAY_MALLOC(
```
gapisback, line 403: Equivalent calls.

src/platform_linux/platform.h
```diff
@@ -328,23 +328,26 @@ extern platform_heap_id     Heap_id;
...
  * -----------------------------------------------------------------------------
  */
-#define TYPED_MANUAL_MALLOC(hid, v, n)                                         \
+#define TYPED_MANUAL_MALLOC(hid, v, n, mf)                                     \
    ({                                                                          \
       debug_assert((n) >= sizeof(*(v)));                                       \
       (typeof(v))platform_aligned_malloc(hid,                                  \
                                          PLATFORM_CACHELINE_SIZE,              \
                                          (n),                                  \
+                                         (mf),                                 \
```
gapisback, line 337: Allocation now receives and returns a `platform_memfrag{} *`, So, macro's call changes.

src/platform_linux/platform.h
```diff
@@ -368,13 +371,13 @@ extern platform_heap_id     Heap_id;
    ({                                                                          \
       debug_assert((n) >= sizeof(*(v)));                                       \
       (typeof(v))platform_aligned_malloc(                                      \
-         hid, (a), (n), STRINGIFY(v), __func__, __FILE__, __LINE__);           \
+         hid, (a), (n), NULL, STRINGIFY(v), __func__, __FILE__, __LINE__);     \
```
gapisback, line 374: For most of the consumers, this is good-enough. I could have changed this to require all callers to also declare an on-stack `platform_memfrag{}`, but that would be more code changes.

The one 'minor' issue with this is that we might incorrectly free a smaller-sized fragment . But that's not a huge loss, so I went with current solution.

src/platform_linux/platform.h
```diff
@@ -739,6 +756,18 @@ platform_enable_tracing_shm_frees();
...
+ * Structure to encapsulate a {memory-addr, memory-size} pair. Used to track
+ * allocation and, more importantly, free of memory fragments for opaque
+ * "objects". Used typically to manage memory for arrays of things.
+ * The 'addr' field is intentionally -not- the 1st field, to reduce lazy
+ * programming which might try to bypass provided interfaces.
+ */
+typedef struct platform_memfrag {
+   size_t size;
+   void  *addr;
+} platform_memfrag;
```
gapisback, line 769: NOTE to myself: Move this to some private header and only expose `memfrag_init_size()` method, somehow.

src/platform_linux/platform.h
```diff
@@ -747,51 +776,199 @@ platform_enable_tracing_shm_frees();
...
+ * ----------------------------------------------------------------------------
  */
-#define platform_free(id, p)                                                   \
+#define platform_free(hid, p)                                                  \
+   do {                                                                        \
+      platform_assert(((p) != NULL),                                           \
+                      "Attempt to free a NULL ptr from '%s', line=%d",         \
+                      __func__,                                                \
+                      __LINE__);                                               \
+      if (IS_MEM_FRAG(p)) {                                                    \
```
gapisback, line 866: This change is key ... and needs understanding. Please review carefully.

src/platform_linux/platform.h
```diff
@@ -747,51 +776,199 @@ platform_enable_tracing_shm_frees();
...
+          * Expect that 'p' is pointing to a struct. So get its size.          \
+          * Except that while allocating memory for objects, we had previously \
+          * aligned the required size to PLATFORM_CACHELINE_SIZE. Redo that    \
+          * work here, so we correctly tell shared-memory the size to free.    \
+          */                                                                   \
+         const size_t _size = sizeof(*p);                                      \
+         const size_t _reqd =                                                  \
+            (_size + platform_alignment(PLATFORM_CACHELINE_SIZE, _size));      \
+         platform_free_mem((hid), (p), _reqd, STRINGIFY(p));                   \
+         (p) = NULL;                                                           \
```
gapisback, line 891: @rtjohnso - This line and L907 below is what makes it necessary for clients calling `free()` to do two things:

```
   platform_memfrag memfrag;
   platform_memfrag *mf;
   ... Do the initialization ...
  platform_free(mf);
```

I would have liked to skip the `mf` and simply pass-in `&memfrag` to `free()`, but there is a compiler error.

I think this can be fixed with some rework ... but I ran out of energy. Let's review if this can be improved.

src/platform_linux/platform_inline.h
```diff
@@ -469,43 +484,58 @@ platform_aligned_malloc(const platform_heap_id heap_id,
...
+
 static inline void *
-platform_realloc(const platform_heap_id heap_id,
-                 const size_t           oldsize,
-                 void                  *ptr, // IN
-                 const size_t           newsize)       // IN
+platform_do_realloc(const platform_heap_id heap_id,
+                    const size_t           oldsize,
+                    void                  *ptr,     // IN
+                    size_t                *newsize, // IN/OUT
```
gapisback, line 499: Reallocation now returns the `*newsize`, so clients like writable buffer resize can record the new fragment's size in its `buffer_capacity` field. 

This, then, allows `writable_buffer_deinit()` to correctly supply the newly realloc'ed fragment's size to `free`.

src/platform_linux/platform_inline.h
```diff
@@ -453,8 +459,17 @@ platform_aligned_malloc(const platform_heap_id heap_id,
    const size_t padding  = platform_alignment(alignment, size);
    const size_t required = (size + padding);
 
-   void *retptr = (heap_id ? splinter_shm_alloc(heap_id, required, objname)
-                           : aligned_alloc(alignment, required));
+   void *retptr = NULL;
+   if (heap_id == PROCESS_PRIVATE_HEAP_ID) {
```
gapisback, line 463: Clarified usage of the semantic of `heap_id`; NULL means process-private-heap-ID, so go thru old `malloc()`-style code-flow.

src/platform_linux/shmem.c
```diff
@@ -39,35 +80,26 @@ static bool Trace_large_frags  = FALSE;
...
-   int      shm_frag_freed_by_pid;     // OS-pid that freed this
-   threadid shm_frag_freed_by_tid;     // Splinter thread-ID that freed this
+   void       *frag_addr;             // Start addr of this memory fragment
+   const char *frag_func;             // Calling func which lead to creation
+   size_t      frag_size;             // bytes
+   threadid    frag_allocated_to_tid; // Alloc'ed to this Splinter thread-ID
+   int         frag_allocated_to_pid; // Allocated to this OS-pid
+   int         frag_freed_by_pid;     // OS-pid that freed this fragment
+   threadid    frag_freed_by_tid;     // Splinter thread-ID that freed this
+   int         frag_line;
```
gapisback, line 99: No changed, indented and aligned fields for readability.

src/platform_linux/shmem.c
```diff
@@ -39,35 +80,26 @@ static bool Trace_large_frags  = FALSE;
...
- * memory chunks for fingerprint array, which is more than a MiB. For scans,
- * splinterdb_iterator_init() allocates memory for an iterator which is ~92+KiB.
- * Set this to a lower value so we can re-cycle free fragments for iterators
- * also.
- */
-#if SPLINTER_DEBUG
-#   define SHM_LARGE_FRAG_SIZE (90 * KiB)
-#else
-#   define SHM_LARGE_FRAG_SIZE (38 * KiB)
-#endif // SPLINTER_DEBUG
```
gapisback: Diff in this limit was causing unit-tests to fail in debug builds and pass in release build.

This artifact is leftover from my poc-dev days ... no longer should be needed to separate out limits.

Have stabilized on 32K as lower limit for large fragments in both builds. Moved to `shmem.h`

src/platform_linux/shmem.c
```diff
@@ -76,6 +108,65 @@ typedef struct shm_frag_info {
...
+#define SHM_SMALL_FRAG_MAX_SIZE 512 // Will not track for size > max bytes
+
+/*
+ * Each free fragment that is hooked into the free-list is described by this
+ * tiny tracking structure.
+ */
+typedef struct free_frag_hdr {
+   struct free_frag_hdr *free_frag_next;
+   size_t                free_frag_size;
+} free_frag_hdr;
```
gapisback, line 126: Used to chain free'd-fragments that are returned to the free-list. This tiny struct lives at the head of each free fragment.

Min frag-size is 64 bytes, so we have room.

src/platform_linux/shmem.c
```diff
@@ -76,6 +108,65 @@ typedef struct shm_frag_info {
...
+
+/*
+ * ------------------------------------------------------------------------
+ * Shared-memory usage statistics & metrics:
+ *
+ * Set of usage-stats fields copied from shmem_info{} struct, so that we
+ * can print these after shared segment has been destroyed.
+ * ------------------------------------------------------------------------
+ */
+typedef struct shminfo_usage_stats {
```
gapisback, line 136: Consolidated all metrics, usage-stats into this common struct.

It will be update in-place nested in the shmem control block.

And will also be used to return metrics when shared memory is being dismantled.

Works much better this way!!

src/platform_linux/shmem.c
```diff
@@ -198,13 +309,13 @@ platform_shm_hip(platform_heap_id hid)
 static inline void
 shm_lock_mem_frags(shmem_info *shminfo)
 {
-   platform_spin_lock(&shminfo->shm_mem_frags_lock);
+   platform_mutex_lock(&shminfo->shm_mem_frags_mutex);
```
gapisback, line 312: Better solution.... Otherwise, for some cases of new `large_inserts_stress_test`, we were simply burning up 100% CPU. 

In all workloads, pthread-semaphore is less than 5%, sometimes even 1-2% as seen in `perf top`.

src/platform_linux/shmem.c
```diff
@@ -278,20 +404,62 @@ platform_shm_print_usage_stats(shminfo_usage_stats *usage)
...
+ * Returns: # of large free-fragments found in-use. Usually when this
+ * function is called for diagnostics, this may not matter. When the
+ * shared segment is dismantled, a non-zero count for "in-use" large fragments
+ * is an indication that something is amiss.
+ */
+int
+platform_save_usage_stats(shminfo_usage_stats *usage, shmem_info *shminfo)
+{
+   *usage                          = shminfo->usage;
+   usage->large_frags_found_in_use = platform_trace_large_frags(shminfo);
```
gapisback, line 453: Much simpler rather than the line-by-line copy that old code from L303 onwards was doing.

src/platform_linux/shmem.c
```diff
@@ -525,22 +685,38 @@ platform_shm_alloc(platform_heap_id hid,
...
+                                            func, file, line))
+                    != NULL)
+      {
+         return retptr;
+      }
+      // clang-format on
+   } else {
+      // Try to satisfy small memory fragments based on requested size, from
+      // cached list of free-fragments.
+      retptr = platform_shm_find_frag(shminfo, size, objname, func, file, line);
```
gapisback, line 710: New `find` method to find small free fragments, thru its free-list-by-size.

src/platform_linux/shmem.c
```diff
@@ -742,42 +965,248 @@ platform_shm_track_alloc(shmem_info *shm, void *addr, size_t size)
...
+/*
+ * Connect a free fragment to the chain provided. Record free-fragment's
+ * size, so we can find it when next required by-size.
+ */
+static inline void
+platform_shm_hook_free_frag(free_frag_hdr **here, void *ptr, size_t size)
+{
+   ((free_frag_hdr *)ptr)->free_frag_next = *here;
+   ((free_frag_hdr *)ptr)->free_frag_size = size;
+   *here                                  = ptr;
```
gapisback, line 983: Inserting free'd- fragment at the head of its free-list

src/platform_linux/shmem.c
```diff
@@ -742,42 +965,248 @@ platform_shm_track_alloc(shmem_info *shm, void *addr, size_t size)
...
+ * list. That means, fragment was previously freed.
+ *
+ * Returns: The size of the free-fragment list in which this 'ptr' was found.
+ *  0, otherwise; (i.e. 'ptr' is not an already-freed-fragment.)
+ *  Optionally, the size marked in this freed-fragment is returned via
+ *  'freed_size'. If a client incorrectly specified the memfrag's size at
+ *  the time of free(), that will be reported here, and can be detected.
+ */
+static size_t
+platform_shm_find_frag_in_freed_lists(const shmem_info *shm,
```
gapisback, line 1061: Used for debug-asserts, and also a wrapper to this is provided, so test code can call it and do this deep cross-check.

src/platform_linux/shmem.c
```diff
@@ -742,42 +965,248 @@ platform_shm_track_alloc(shmem_info *shm, void *addr, size_t size)
...
+          * the sizeof(struct) for alignment. But it could well be that we
+          * recycled a small fragment from a free-list, where the fragment's
+          * original size was larger than sizeof(struct). So, come now to
+          * free()-time, caller only knows of sizeof(struct) which will be
+          * smaller than the actual fragment's size. This leads to a small
+          * bit of unused space in the to-be-freed-fragment, but should not
+          * cause any memory corruption.
+          * ------------------------------------------------------------------
+          */
+         platform_shm_hook_free_frag(next_frag, addr, size);
```
gapisback, line 1166: Here is where we insert free'd fragment to its bucket free-list.

src/platform_linux/shmem.c
```diff
@@ -859,52 +1323,59 @@ platform_shm_find_free(shmem_info *shm,
...
-      frag->shm_frag_allocated_to_tid = platform_get_tid();
+      frag->frag_allocated_to_pid = getpid();
+      frag->frag_allocated_to_tid = platform_get_tid();
 
-      shm->shm_num_frags_inuse++;
-      if (shm->shm_num_frags_inuse > shm->shm_num_frags_inuse_HWM) {
-         shm->shm_num_frags_inuse_HWM = shm->shm_num_frags_inuse;
+      shm->usage.nfrags_inuse++;
+      if (shm->usage.nfrags_inuse > shm->usage.nfrags_inuse_HWM) {
+         shm->usage.nfrags_inuse_HWM = shm->usage.nfrags_inuse;
```
gapisback, line 1360: This does a "first-fit" search. It will be better to enhance this to "best-fit " search.
Otherwise, fragments freed from large fingerprint arrays, I've seen, get re-cycled for requests much smaller.

Not a deal breaker ... leave it for later ...

src/platform_linux/shmem.c
```diff
@@ -943,11 +1414,65 @@ platform_shm_find_free(shmem_info *shm,
...
+   // If we are not tracking fragments of this size, nothing further to do.
+   free_frag_hdr **next_frag = platform_shm_free_frag_hdr(shm, size);
+   if (next_frag == NULL) {
+      return NULL;
+   }
+   shm_lock_mem_frags(shm);
+
+   // Find the next free frag which is big enough
+   while (*next_frag && ((*next_frag)->free_frag_size < size)) {
+      next_frag = &(*next_frag)->free_frag_next;
```
gapisback, line 1457: Again, could improve this to do a 'best-fit' search.

However, for buckets of size `<= 64` and `<= 128`, by nature of cache line alignments, all free fragments will per force be of 64 or 128 bytes.

The advantages of best-fit or right-fit could be seen only in higher sized buckets: <=256 or higher ...

For another rainy day ...

src/trunk.c
```diff
@@ -703,7 +703,11 @@ typedef enum trunk_compaction_type {
    NUM_TRUNK_COMPACTION_TYPES,
 } trunk_compaction_type;
 
-// arguments to a compact_bundle job
+/*
+ * Arguments to a compact_bundle job.
+ * Memory for the fingerprint array will be allocated, when needed,
+ * and hung off of breq_fingerprint.
+ */
```
gapisback, line 710: NOTE to reviewers: Fingerprint object handling and gyrations are very complex.

I had to struggle to get this right.

Kindly review this file very carefully.

src/trunk.c
```diff
@@ -719,7 +723,9 @@ struct trunk_compact_bundle_req {
    uint64                output_pivot_kv_byte_count[TRUNK_MAX_PIVOTS];
    uint64                tuples_reclaimed;
    uint64                kv_bytes_reclaimed;
-   uint32               *fp_arr;
+   fp_hdr                breq_fingerprint;
+   uint64                num_tuples;
```
gapisback, line 727: NOTE to me: Revisit field. fp_hdr has ntuples inside it already. This field was probably added in early rounds before `fp_hdr` was formalized.

src/trunk.c
```diff
@@ -7092,23 +7229,57 @@ trunk_range(trunk_handle  *spl,
    return rc;
 }
 
+/*
+ * Allocate memory for stats-related structures off the trunk.
+ */
+static trunk_stats *
+trunk_stats_init(trunk_handle *spl)
```
gapisback, line 7236: Modularized common code from `trunk_create` and `trunk_mount`, so that the memory-related changes could be applied to just one place.

src/trunk.c
```diff
@@ -7092,23 +7229,57 @@ trunk_range(trunk_handle  *spl,
...
  */
-trunk_handle *
-trunk_create(trunk_config     *cfg,
-             allocator        *al,
-             cache            *cc,
-             task_system      *ts,
-             allocator_root_id id,
-             platform_heap_id  hid)
+static trunk_handle *
+trunk_handle_init(trunk_config     *cfg,
```
gapisback, line 7270: Modularized common code from `trunk_create` and `trunk_mount`, so that the memory-related changes could be applied to just one place.

src/util.c
```diff
@@ -20,8 +20,14 @@ writable_buffer_ensure_space(writable_buffer *wb,
...
-   void *oldptr  = wb->can_free ? wb->buffer : NULL;
-   void *newdata = platform_realloc(wb->heap_id, oldspace, oldptr, minspace);
+   void *oldptr = wb->can_free ? wb->buffer : NULL;
+
+   // NOTE: realloc() may adjust-up 'minspace' for alignment
+   // ALSO: We must supply correct size of fragment being freed, which
+   //   will be the buffer capacity, but not just the oldspace in-use.
+   //   (Otherwise, free-fragment management will run into memory leaks.)
+   void *newdata =
+      platform_realloc(wb->heap_id, wb->buffer_capacity, oldptr, &minspace);
```
gapisback, line 30: Important bug-fix that call to `realloc` should pass buffer's capacity (i.e., old memory fragment's size) but not `oldspace` (which could be far less than capacity).

src/util.h
```diff
@@ -278,6 +289,276 @@ writable_buffer_append(writable_buffer *wb, uint64 length, const void *newdata)
    writable_buffer_copy_slice(&dst##wb, src);                                  \
    slice dst = writable_buffer_to_slice(&dst##wb);
 
+/*
+ * ----------------------------------------------------------------------
+ * Fingerprint Header: Fingerprints are managed while building routing filters.
```
gapisback, line 294: This almost wants to be in its own file: `fingerprint.h` and included in the two consumers: `btree.c` and `trunk.c`.

Wanted to ask your opinions before adding a new file.

gapisback:
NOTE: to the reviewers: @rtjohnso @ajhconway @rosenhouse :

This is part-3 of the shared memory support dev work.

I have layered this on top of the [aguraajda/shmem-mp-support-Rev](https://github.com/vmware/splinterdb/tree/aguraajda/shmem-mp-support-Rev), so you will see 3 commits. ONLY review the diffs in the [3rd commit]([Support free-fragment recycling in shared-segment. Add fingerprint ob…](https://github.com/vmware/splinterdb/pull/569/commits/c2c437998b32167277463b2fbcce5b2091625ace)).

The other two previous commits are being reviewed under previous 2 PRs.

Here is the order to review files in, to get a good grip on this change-set:

1. `platform.h`, `platform_inline.h`: For changes in `TYPED_*ALLOC` interfaces. Introduction of `platform_memfrag{}`
2. `util.h` and `util.c` - See the changes to add `struct fp_hdr{}` and fingerprint object mgmt APIs. Some changes related to writable_buffers are also in these two files.
3. `btree.h` and `btree.c` - Initial exposure to finger print object changes
4. `shmem.c` - For the core of free-list management logic changes
5. Rest of the files where the interfaces changes are religiously plumbed through alloc / free actions.
6. Check `test.sh` to see extended test-coverage being brought-in.
7. Specifically enabling `--use-shmem` to functional `filter_test.c` was a BIG Help to stabilize fingerprint object mgmt
8. See also in unit: `platform_apis_test.c`, `splinter_shmem_test.c`, and few others that were enhanced.

gapisback:
HI, @rtjohnso - Status update.

I have re-based this work on top of `/main` and have gone thru multiple commits to stabilize this work. However, it's not quite, yet, ready for review, for these reasons:

1. While  finalizing these changes, I came across some calls to `platform_free()` which looked suspiciously wrong. They could result in unaccounted memory-usage being 'freed' causing, over time, a slow "memory leak". 

2. Rather than fixing instances one-at-a-time, I started to implement some sanity checking to verify that the memory fragment being freed is, indeed, an allocated one and that the allocated fragment's size matches the `size` provided to free. Even early implementations of this sanity-checking tripped up quite easily.

3. I realized that in my previous implementation, I had not roto-tilled interfaces such as `TYPED_MALLOC()`, `TYPED_ZALLOC` and `TYPED_ALIGNED_MALLOC()`, `TYPED_ALIGNED_ZALLOC` to take the `platform_memfrag *` argument. (They were left to work in the current style, somewhat as a convenience.) I now think it's better to tighten all memory allocation and free interfaces to always take a `platform_memfrag *` argument.  This will greatly simplify and tighten sanity and assertion checking during free.

4. CI-jobs are jammed -- reasons are unknown to me. They are not getting unblocked. So, I have no insight into overall stability of this work for the larger set of CI test-jobs. (Pending item ...) 

I'm off rest of this week of Thanksgiving. Will try to see if I can get some time off-line to work on (3) and finish-up (2) above. I will try to complete this work and bring something to review 1st week of Dec.

Makefile
```diff
@@ -55,7 +55,7 @@ TESTSRC := $(COMMON_TESTSRC) $(FUNCTIONAL_TESTSRC) $(UNIT_TESTSRC)
 # Construct a list of fast unit-tests that will be linked into unit_test binary,
 # eliminating a sequence of slow-running unit-test programs.
 ALL_UNIT_TESTSRC := $(call rwildcard, $(UNIT_TESTSDIR), *.c)
-SLOW_UNIT_TESTSRC = splinter_test.c config_parse_test.c large_inserts_bugs_stress_test.c splinterdb_forked_child_test.c
+SLOW_UNIT_TESTSRC = splinter_test.c config_parse_test.c large_inserts_stress_test.c splinterdb_forked_child_test.c
```
gapisback, line 58: Test file is renamed to take out `bugs` from the name.

And this `large_inserts_stress_test` will not be part of basic `unit_test` execution.

`test.sh` invokes this test independently with different test-execution parameters.

Makefile
```diff
@@ -476,6 +476,10 @@ $(BINDIR)/$(UNITDIR)/splinter_shmem_test: $(UTIL_SYS)            \
                                           $(COMMON_UNIT_TESTOBJ) \
                                           $(LIBDIR)/libsplinterdb.so
 
+$(BINDIR)/$(UNITDIR)/splinter_shmem_oom_test: $(UTIL_SYS)            \
```
gapisback, line 479: Split out OOM-related test case(s) out of `unit/splinter_shmem_test`

src/btree.c
```diff
@@ -3167,8 +3173,8 @@ btree_pack_loop(btree_pack_req *req,       // IN/OUT
    log_trace_key(tuple_key, "btree_pack_loop (bottom)");
 
    if (req->hash) {
-      platform_assert(req->num_tuples < req->max_tuples);
-      req->fingerprint_arr[req->num_tuples] =
+      platform_assert(btree_pack_can_fit_tuple(req));
```
gapisback, line 3176: Packaged the condition on deleted L3170 to this simple `bool` checker fn.

src/clockcache.c
```diff
@@ -1816,20 +1816,25 @@ clockcache_init(clockcache        *cc,   // OUT
    cc->heap_id = hid;
 
    /* lookup maps addrs to entries, entry contains the entries themselves */
-   cc->lookup =
-      TYPED_ARRAY_MALLOC(cc->heap_id, cc->lookup, allocator_page_capacity);
+   platform_memfrag memfrag_cc_lookup;
+   cc->lookup = TYPED_ARRAY_MALLOC_MF(
+      cc->heap_id, cc->lookup, allocator_page_capacity, &memfrag_cc_lookup);
```
gapisback, line 1821: This is the 1st example of use of the MF-named array-allocator; where the `memfrag{} *` pointer is passed-in through the allocator macro.

This interface exists as the object for which the allocation is being done, i.e., `cc->lookup`, is not amenable to name translation using `##` preprocessor directives. 

[ See further below for explanations on other examples of new interface usages. ]

src/clockcache.c
```diff
@@ -1816,20 +1816,25 @@ clockcache_init(clockcache        *cc,   // OUT
...
 
-   cc->entry =
-      TYPED_ARRAY_ZALLOC(cc->heap_id, cc->entry, cc->cfg->page_capacity);
+   platform_memfrag memfrag_cc_entry;
+   cc->entry = TYPED_ARRAY_ZALLOC_MF(
+      cc->heap_id, cc->entry, cc->cfg->page_capacity, &memfrag_cc_entry);
    if (!cc->entry) {
       goto alloc_error;
    }
+   cc->entry_size = memfrag_size(&memfrag_cc_entry);
```
gapisback, line 1837: Alternative: One thing I considered was to require that all such structures have a commonly-named `mf_size` field and this caller-macro will (behind-the-scenes) stash away `memfrag->size` into this structure's`mf_size` field.

But that solution hides some of this machinations behind macros, which might make code tracking difficult.

Also, not all structures need, or even have, this `mf_size` field. You only need this tracking `mf_size` if the free is done in some entirely different function. 

For most usages, where allocation and free is done in the same function, we just directly use `memfrag->size` to do the free. In such cases, the allocating structure (or memory ptr) will not need, and will not have, a `mf_size` field.

Hence, this alternative hasn't been implemented.

src/clockcache.h
```diff
@@ -139,15 +139,17 @@ struct clockcache {
 
    // Stats
    cache_stats stats[MAX_THREADS];
+   size_t      lookup_size;
+   size_t      entry_size;
+   size_t      pincount_size;
+   size_t      batch_busy_size;
```
gapisback, line 145: Here is an example of a structure which has 4 member ptr-fields for which we have to allocate separate memory fragments.

Each of these memory fragment's size has to be independently tracked. So, it is difficult to have just one structure-wide `mf_size` field to track size of fragment allocation.

(This point is in continuation of the [alternate solution remark](https://github.com/vmware/splinterdb/pull/569/files#r1421429375) on L1837 above.)

src/memtable.c
```diff
@@ -307,15 +307,19 @@ memtable_context_create(platform_heap_id hid,
                         process_fn       process,
                         void            *process_ctxt)
 {
+   platform_memfrag  memfrag_ctxt = {0};
```
gapisback, line 310: Here is the 1st example of the new memory allocation interface(s), some of which require an on-stack `platform_memfrag` structure to be declared, named as an extension of the `ptr-handle-variable-name` for which allocation is being done.

Here, we are allocating memory for `ctxt` variable, so the macro expects that a `memfrag_ctxt` be declared on-stack. Otherwise, compilation will fail.

You will see many such patterns of this coding style.

I chose this approach so that we do not perturb the call to memory allocation on L312 below, and still bring-in the requirement of `platform_memfrag *` to memory allocation interfaces.

The approach chosen reduces code churn to some extend in this review packet.

src/memtable.c
```diff
@@ -307,15 +307,19 @@ memtable_context_create(platform_heap_id hid,
...
    memtable_context *ctxt =
       TYPED_FLEXIBLE_STRUCT_ZALLOC(hid, ctxt, mt, cfg->max_memtables);
-   ctxt->cc = cc;
+   ctxt->mf_size = memfrag_size(&memfrag_ctxt);
```
gapisback, line 313: And the 2nd pattern you will see is:

After a successful memory allocation, code will (and needs to):

1. Save off the size of the memory fragment allocated using `memfrag_size()`.
2. Consistently, all structures that need to track this fragment-size will have a field named `mf_size`. 
3. Much later, `platform_free()` will be invoked using this `structure->mf_size` field.

I have tried to keep this name the same in all places. If you find something is not named like so, flag it and I will fix that.

src/memtable.c
```diff
@@ -343,9 +347,8 @@ memtable_context_destroy(platform_heap_id hid, memtable_context *ctxt)
    }
 
    platform_mutex_destroy(&ctxt->incorporation_mutex);
-   platform_free(hid, ctxt->rwlock);
-
-   platform_free(hid, ctxt);
+   platform_free_mem(hid, ctxt->rwlock, ctxt->rwlock_mf_size);
+   platform_free_mem(hid, ctxt, ctxt->mf_size);
```
gapisback, line 351: NOTE: Here is an example where we have two memory fragment sizes to be tracked.

1. `rwlock_mf_size` - tracks size of fragment for rw-lock field.
2. `mf_size` - tracks size of fragment for holding `ctxt` structure.

Hence, these two slightly differently named `size`-fields.

src/merge.c
```diff
@@ -545,6 +545,7 @@ merge_iterator_create(platform_heap_id hid,
                      == ARRAY_SIZE(merge_itor->ordered_iterators),
                   "size mismatch");
 
+   platform_memfrag memfrag_merge_itor;
```
gapisback, line 548: Another idiomatic usage example:

To allocate memory for a field named `merge_itor`, declare an on-stack `platform_memfrag memfrag_merge_itor`, which is used by the allocator macro.

src/platform_linux/shmem.c
gapisback: Updated: _Nothing_ changed ... only indentation changes ...

src/rc_allocator.c
```diff
@@ -240,13 +240,16 @@ rc_allocator_init_meta_page(rc_allocator *al)
...
-   al->meta_page = TYPED_ALIGNED_ZALLOC(al->heap_id,
-                                        al->cfg->io_cfg->page_size,
-                                        al->meta_page,
-                                        al->cfg->io_cfg->page_size);
+   platform_memfrag memfrag_meta_page;
+   al->meta_page = TYPED_ALIGNED_ZALLOC_MF(al->heap_id,
+                                           al->cfg->io_cfg->page_size,
+                                           al->meta_page,
+                                           al->cfg->io_cfg->page_size,
+                                           &memfrag_meta_page);
```
gapisback, line 248: Here's an example of an allocation interface where the macro cannot simply synthesize the `platform_memfrag* mf` name using `##v` notation.

The caller has to supply the required `memfrag *mf` arg by name; i.e. `&memfrag_meta_page`.

src/routing_filter.c
```diff
@@ -595,7 +595,15 @@ routing_filter_add(cache          *cc,
...
          // ALEX: for now the indices must fit in a single extent
-         debug_assert(index_no / addrs_per_page < pages_per_extent);
+         platform_assert((index_no / addrs_per_page < pages_per_extent),
+                         "index_no=%u, addrs_per_page=%lu"
+                         ", (index_no / addrs_per_page)=%lu"
+                         ", pages_per_extent=%lu",
+                         index_no,
+                         addrs_per_page,
+                         (index_no / addrs_per_page),
+                         pages_per_extent);
```
gapisback, line 605: Changed to `platform_assert()` and elaborated diag info ... as I was running into some bugs in this area before fingerprint / fp-array handling was eventually stabilized.

test.sh
```diff
@@ -232,23 +232,34 @@ function nightly_functionality_stress_tests() {
 # of execution, especially to shake out AIO / thread registration issues.
 # #############################################################################
 function nightly_unit_stress_tests() {
-    local use_shmem=$1
+    local use_shmem=""
+    if [ $# -eq 1 ]; then
+        use_shmem="$1"
```
gapisback, line 237: This block of diffs will be seen in multiple places in this file.

This change now allows one to call the test-execution function, like so:

 `$ ./test.sh nightly_unit_stress_tests`
 
 
 i.e., we can now invoke this shell script to do dev-test on individual functions without shared memory configuration. (Previously that invocation would return a shell error.)

test.sh
```diff
@@ -857,19 +926,43 @@ function run_btree_tests() {
...
+   fi
     local use_msg=
     if [ "$use_shmem" != "" ]; then
         use_msg=", using shared memory"
    fi
+
+    # shellcheck disable=SC2086
     run_with_timing "Cache test${use_msg}" \
-        "$BINDIR"/driver_test cache_test --seed "$SEED"
+        "$BINDIR"/driver_test cache_test --seed "$SEED" $use_shmem
```
gapisback, line 940: These diffs fix an existing test-execution bug. These functional tests were supposed to have been enabled to be run with `--use-shmem` flag, but the invocation was missing that `$use_shmem` arg.

tests/config.c
```diff
@@ -28,9 +28,6 @@
 #define TEST_CONFIG_DEFAULT_MAX_BRANCHES_PER_NODE 24
 
 // Deal with reasonable key / message sizes for tests
-// There are open issues in some tests for smaller key-sizes.
-// For now, restrict tests to use this minimum key-size.
-#define TEST_CONFIG_MIN_KEY_SIZE         ((int)sizeof(uint64))
```
gapisback, line 33: Define moved to `tests/config.h` as it's now also referenced in `tests/unit/large_inserts_stress_test.c`.

tests/unit/splinterdb_stress_test.c
```diff
@@ -219,60 +219,6 @@ CTEST2(splinterdb_stress, test_iterator_over_many_kvs)
...
 }
 
-/*
- * Test case that inserts large # of KV-pairs, and goes into a code path
- * reported by issue# 458, tripping a debug assert. This test case also
- * triggered the failure(s) reported by issue # 545.
- * FIXME: This test still runs into an assertion "filter->addr != 0"
- * from trunk_inc_filter(), which is being triaged separately.
- */
-CTEST2_SKIP(splinterdb_stress, test_issue_458_mini_destroy_unused_debug_assert)
```
gapisback, line 229: Test case relocated to `unit/large_inserts_stress_test.c`, as it better belongs to that test file.

gapisback:
@rtjohnso - The final part-3 shared memory support change-set is now ready for review.

The suggested order in which to review these diffs is:

1. src/platform_linux/platform.h
2. src/platform_linux/platform_inline.h
3. src/platform_linux/platform_types.h
4. src/util.h, src/util.c
5. src/platform_linux/shmem.h
6. src/platform_linux/shmem.c
7. src/trunk.h, src/trunk.c
8. src/routing_filter.c
9. Then the rest of the files.
10. Good luck!

rtjohnso:
I think the current memfrag interface is leaky and not general.

I think the interface should look like this:

```
platform_status
platform_alloc(memfrag *mf, // OUT
               int size);

platform_status
platform_realloc(memfrag *mf, // IN/OUT
                 int newsize);

platform_status
platform_free(memfrag *mf); // IN

void *
memfrag_get_pointer(memfrag *mf);
```

(Note that details, like the exact names of the functions or the memfrag datatype are not too important in this example.)

The point is that the rest of the code should treat memfrags as opaque objects.  In the current code, the rest of the code goes around pulling out fields and saving them for later use.  It means that internal details of the current allocator implementation are being leaked all over the rest of the code.  This will make it difficult to change the allocator implementation down the road.

As for names, I would advocate renaming memfrag to `memory_allocation`.

gapisback:
HI, @rtjohnso -- 

Thanks for your [initial approach on reworking the interfaces](https://github.com/vmware/splinterdb/pull/569#issuecomment-1913053101).

I'm happy to take this further, but I feel this round-trip discussion will become long and meandering. And this review panel UI exchange is not ideally suited for that kind of interaction.

I want to avoid re-doing the implementation till we've settled on and agreed to the new interfaces. Every bit of code rework requires massively editing the change-set and re-stabilizing - an effort I would like to avoid doing multiple times.

How about I start a new thread under [Discussions tab](https://github.com/vmware/splinterdb/discussions), with your initial proposal? And, will give you my responses, rebuttal. I suspect we will have to go back-and-forth a few times before settling on the final interfaces.

(As a team, we haven't used the Discussions tab feature internally. As I am beginning my transition to fully out-of-VMware, it may be a good opportunity to engage using this GitHub feature, so it continues even when I'm a fully O-Sourced' engineer.)

gapisback:
@rtjohnso - My CI-stabilization jobs have succeeded. I have squashed all changes arising from our proposal [discussion thread](https://github.com/vmware/splinterdb/discussions/615) into this [one single commit](https://github.com/vmware/splinterdb/pull/569/commits/9281c83f8d953edfaa6d8a338f97da32a53001ed) and have refreshed this change-set.

You can restart your review on this amended change-set. (I expect CI-jobs will succeed as they did in the stabilization PR #616 )

gapisback:
@rtjohnso : Fyi -- I want to log this one MSAN-instability the most recent round of CI-jobs ran into, as I am not going to remember all this later. 

Here is the state of affairs and results of my investigations.

1. CI [Job no. 109](https://runway-ci.eng.vmware.com/teams/splinterdb/pipelines/splinterdb/jobs/main-pr-asan/builds/109) (main-pr-asan) job failed with this error:

```
 build/release-asan/bin/driver_test splinter_test --perf --use-shmem --max-async-inflight 0 --num-insert-threads 4 --num-lookup-threads 4 --num-range-lookup-threads 0 --tree-size-gib 2 --cache-capacity-mib 512

build/release-asan/bin/driver_test: splinterdb_build_version 9281c83f

Dispatch test splinter_test

Attempt to create shared segment of size 8589934592 bytes.

Created shared memory of size 8589934592 bytes (8 GiB), shmid=8617984.

Completed setup of shared memory of size 8589934592 bytes (8 GiB), shmaddr=0x7f6924570000, shmid=8617984, available memory = 8589894272 bytes (~7.99 GiB).

Running splinter_test with 1 caches

splinter_perf_inserts() starting num_insert_threads=4, num_threads=4, num_inserts=27185152 (~27 million) ...

Thread 2 inserting  37% complete for table 0 ... =================================================================

==2666==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7f68fba16f80 at pc 0x7f6b276fef50 bp 0x7f69044aa080 sp 0x7f69044a9828

READ of size 589 at 0x7f68fba16f80 thread T1

Thread 2 inserting  42% complete for table 0 ... OS-pid=2666, OS-tid=2669, Thread-ID=3, Assertion failed at src/trunk.c:2213:trunk_get_new_bundle(): "(node->hdr->end_bundle != node->hdr->start_bundle)". No available bundles in trunk node. page disk_addr=1513291776, end_bundle=3, start_bundle=3

./test.sh: line 115:  2666 Aborted                 "$@"

make: *** [Makefile:558: run-tests] Error 134
```
2. Upon re-run this [asan job no. 109.1](https://runway-ci.eng.vmware.com/teams/splinterdb/pipelines/splinterdb/jobs/main-pr-asan/builds/109.1) succeeded.

3. Attempted to manually re-run this specific test multiple times on my Nimbus-VM,but could not reproduce the ASAN heap-buffer-overflow error. Ran the exact test with different combinations (one run, 4 concurrent runs with the exact same params, 4 concurrent executions with increasing thread-count up to `--num-insert-threads 8 --num-lookup-threads 8`, and similar stress load on the VM), but could not repro the problem outside CI.

The last variation of this test in manual repro attempts I tried is 4 concurrent invocations of this test: (Logging this here so I can refer to this later on.)
```
./driver_test splinter_test --perf --use-shmem --max-async-inflight 0 --num-insert-threads 8 --num-lookup-threads 8 --num-range-lookup-threads 0 --tree-size-gib 2 --cache-capacity-mib 512
```

The VM has 16 vCPUs, so I figured by running with 8 insert-threads and 4 concurrent instances, we'd load the CPU high-enough to tickle any bugs out. But the ASAN problem did not recur in these manual repro attempts.

----

**NOTE**: In the original failure in CI, hard to tell exactly, but it seems like the thread ID 2 ran into the ASAN memory over flow and soon after, thread ID=3 ran into this assertion a few lines later:

```
OS-pid=2666, OS-tid=2669, Thread-ID=3,  Assertion failed at src/trunk.c:2213:trunk_get_new_bundle(): "(node->hdr->end_bundle != node->hdr->start_bundle)". No available bundles in trunk node. page disk_addr=1513291776, end_bundle=3, start_bundle=3
```

You may recall that I had reported issue #474 some time ago for this trunk bundle mgmt assertion.

I suspect that there is something lurking there that popped up in the CI-run.

I cannot explain how / whether / if this assertion tripping is caused by the ASAN heap-buffer-overflow error or if they are even related. Unfortunately, I could not repro the ASAN issue outside CI, so have to give up on this investigation now.

The rest of the test runs are stable, and this MSAN-job did succeed on a re-run. I have re-reviewed the code-diffs applied recently and could not find anything obviously broken. For now, I will have to conclude that the changes are fine except there may be some hidden instability popping up, possibly triggered by issue #474 mentioned earlier.

src/platform_linux/platform.c
rtjohnso: Actually, I believe the correct macro for this situation is `TYPED_FLEXIBLE_STRUCT_ZALLOC`.

src/platform_linux/platform.h
```diff
@@ -356,24 +357,32 @@ extern platform_heap_id Heap_id;
...
    ({                                                                          \
       debug_assert((n) >= sizeof(*(v)));                                       \
       (typeof(v))platform_aligned_malloc(                                      \
-         hid, (a), (n), STRINGIFY(v), __func__, __FILE__, __LINE__);           \
+         (mf), hid, (a), (n), STRINGIFY(v), __func__, __FILE__, __LINE__);     \
    })
-#define TYPED_ALIGNED_ZALLOC(hid, a, v, n)                                     \
+
+#define TYPED_ALIGNED_MALLOC(hid, a, v, n)                                     \
+   TYPED_ALIGNED_MALLOC_MF(&memfrag_##v, hid, a, v, n)
```
rtjohnso, line 374: I hate this hard-coding of this naming convention, and I think it doesn't solve an important problem.

Let me propose something that might be more useful.

```
#define TYPED_ALIGNED_MALLOC_AUTO(hid, a, v, n)                                     \
  platform_memfrag v_##memfrag __attribute__(cleanup(platform_free));
  TYPED_ALIGNED_MALLOC_MF(&v_##memfrag, hid, a, v, n)
```

This allocates memory that will automatically get freed when the function exits the scope of the allocation.

Furthermore, it alleviates the user of the responsibility to declare a memfrag at all.

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
  * Non-platform-specific inline implementations
  */
 
+/*
+ * Utility macro to test if an argument to platform_free() is a
+ * platform_memfrag *.
+ */
+#define IS_MEM_FRAG(x)                                                         \
```
rtjohnso, line 795: This will go away, right?

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
...
+   __builtin_choose_expr(                                                      \
+      __builtin_types_compatible_p(typeof((platform_memfrag *)0), typeof(x)),  \
+      1,                                                                       \
+      0)
+
+/* Helper methods to do some common operations */
+#define memfrag_start(mf) ((mf)->addr)
```
rtjohnso, line 802: Any reason not to make these static inline functions?

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
...
+             void             *ptr,
+             size_t            nbytes)
+{
+   mf->hid  = hid;
+   mf->addr = ptr;
+   mf->size = nbytes;
+}
+
+static inline bool
+memfrag_is_empty(const platform_memfrag *mf)
```
rtjohnso, line 818: `empty` is vague.  Let's stick with `null`.

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
...
+ *
+ * Free a memory fragment at address 'p' of size 'size' bytes.
+ *
+ * These exist to facilitate re-cycling of free'd fragments in a shared-memory
+ * usage. That machinery works off of the fragment's 'size', hence we need to
+ * provide 'size' through this interface.
+ * ----------------------------------------------------------------------------
+ */
+// clang-format off
+#define platform_free_mem(hid, p, size)                                        \
```
rtjohnso, line 875: This api will go away, right?

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
...
+        platform_do_free((hid), (p), (size), STRINGIFY(p))
 
-#define platform_free_volatile(id, p)                                          \
+#define platform_do_free(hid, p, size, objname)                              \
```
rtjohnso, line 878: And this will go away, right?

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
...
+                                       __FILE__,                               \
+                                       __LINE__);                              \
+      _mf->addr = NULL;                                                        \
+      _mf->size = 0;                                                           \
    } while (0)
 
 // Convenience function to free something volatile
 static inline void
 platform_free_volatile_from_heap(platform_heap_id heap_id,
                                  volatile void   *ptr,
```
rtjohnso, line 922: This will change to accept a memfrag, right?

src/platform_linux/platform_inline.h
```diff
@@ -40,10 +40,11 @@ platform_checksum_is_equal(checksum128 left, checksum128 right)
 static void
 platform_free_from_heap(platform_heap_id UNUSED_PARAM(heap_id),
                         void            *ptr,
+                        const size_t     size,
```
rtjohnso, line 43: This will be modified to take a memfrag, right?

src/platform_linux/platform.h
```diff
@@ -734,51 +788,161 @@ platform_strtok_r(char *str, const char *delim, platform_strtok_ctx *ctx);
...
+ * There is just one consumer of this interface, so we don't go to the full
+ * distance as its sibling interface, to do error checking of args etc.
+ * ----------------------------------------------------------------------------
+ */
+#define platform_free_volatile(hid, p)                                         \
    do {                                                                        \
-      platform_free_volatile_from_heap(                                        \
-         id, (p), STRINGIFY(p), __func__, __FILE__, __LINE__);                 \
-      (p) = NULL;                                                              \
+      debug_assert(((p) != NULL),                                              \
```
rtjohnso, line 898: Once you change `platform_free_volatile_from_heap` to take a memfrag, you won't need all this type checking.

src/platform_linux/platform_inline.h
```diff
@@ -463,62 +469,99 @@ platform_aligned_malloc(const platform_heap_id heap_id,
...
    const size_t required = (size + padding);
 
-   void *retptr =
-      (heap_id
-          ? platform_shm_alloc(heap_id, required, objname, func, file, lineno)
-          : aligned_alloc(alignment, required));
+   void *retptr = NULL;
+   if (heap_id == PROCESS_PRIVATE_HEAP_ID) {
+      retptr = aligned_alloc(alignment, required);
+      if (memfrag) {
```
rtjohnso, line 475: Get rid of the `if`.  Always require a memfrag parameter.

src/platform_linux/platform_inline.h
```diff
@@ -463,62 +469,99 @@ platform_aligned_malloc(const platform_heap_id heap_id,
...
+/*
+ * platform_free_from_heap() - Free memory from the heap.
+ *
+ * If Splinter is running with shared memory configured, this calls into
+ * shared-memory based free() method. Othewise, run standard free().
+ */
 static inline void
 platform_free_from_heap(platform_heap_id heap_id,
                         void            *ptr,
+                        const size_t     size,
```
rtjohnso, line 555: Take a memfrag instead.

rtjohnso (changes requested):
I've mostly just gone through the headers in the `platform` code, plus the fingerprint array api.

I assume once we get these nailed down, then most of the changes in the rest of the code will be relatively straightforward updates to the new apis.

Or is there anything else major?

Let's get the new apis sorted and then I can review the whole PR.

gapisback:
@rtjohnso - I've gone thru your review comments quickly. Most of those are easily implementable. I will get to it.

> I've mostly just gone through the headers in the `platform` code, plus the fingerprint array api.
>
I am curious about your review of the fingerprint array API rework. Did you not find any issues with that? I was bracing myself to get lots of comments as this area is fragile and the rework is a bit tricky. If you think this array API is acceptable, then that will reduce a bunch of rework rounds on me.
 
> Let's get the new apis sorted and then I can review the whole PR.

Let me apply the changes requested and then re-test. (CI-re-test stabilization will be a nightmare starting tomorrow.)

Once I go over all the changes, I will be better able to answer this question of yours:

> Or is there anything else major?
>
... for which the answer now is, I don't think so, off-hand.

rtjohnso:
I left a few comments on the fingerprint array code already.

I haven't done a full evaluation.  It seemed more complex than I expected, but I see that it is trying to make explicit some of the complex sharing that goes on with the fingerprint arrays, which is a goal I like.  I will want to do a more thorough review of how it is used to understand how it all fits together.

rtjohnso:
I spoke with Alex today about the overall design, and he really doesn't like how the whole concept of memfrags puts a burden on the rest of the code.

So let's do the following.  Whenever the shm code allocates memory, it allocates one extra cache line in front, and stores the memfrag on that cacheline.  Later, during a free, you use pointer arithmetic to find the memfrag for that pointer.

PR #575, opened 2023-04-25
fix next_req node-split bug in trunk

rtjohnso:


rtjohnso:
Confirmed w/ Alex over slack.

Issue #581, opened 2023-06-03
A segmentation fault during closing database


deukyeon:
I got a segmentation fault error in `trunk_unmount()`.
https://github.com/vmware/splinterdb/blob/6a2348c0eb9887cdafee3dad674a96e41edddb28/src/trunk.c#L7725-L7746
This function calls `trunk_prepare_for_shutdown(spl);`, followed by `trunk_set_super_block(spl, FALSE, TRUE, FALSE);`.

In the `trunk_prepare_for_shutdown()`, `spl->log` is freed and then becomes NULL by `platform_free()`. https://github.com/vmware/splinterdb/blob/6a2348c0eb9887cdafee3dad674a96e41edddb28/src/trunk.c#L7646
However, the `spl->log` is referenced by `trunk_set_super_block()`, which causes a segmentation fault. https://github.com/vmware/splinterdb/blob/6a2348c0eb9887cdafee3dad674a96e41edddb28/src/trunk.c#L934

It seems that the `trunk_set_super_block()` function is based on the assumption, "init means destroying the current super block".  So, it is used in both creating and destroying the database. One way to resolve this is setting the addresses of log and log metadata to 0 if `spl->log` is NULL, which can indicate that the database is being unmounted.

PR #582, opened 2023-06-03
(#581) Resolve a segmentation fault on trunk_unmount()

deukyeon:
`trunk_set_super_block()` is used when both creating and closing the splinterdb. However, `spl->log` is already freed and becomes NULL when it gets to this function. This causes a segmentation fault because `spl->log` is referenced when it stores the log addresses into the super block.  As a fix, it sets `super->log_addr` and `super->log_meta_addr` if `spl->log` is valid. Otherwise, they are set to 0.

### DIFF
--- a/src/trunk.c
+++ b/src/trunk.c
@@ -931,8 +931,13 @@ trunk_set_super_block(trunk_handle *spl,
    super->root_addr = spl->root_addr;
    super->meta_tail = mini_meta_tail(&spl->mini);
    if (spl->cfg.use_log) {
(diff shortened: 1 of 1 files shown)


ajhconway (approved):
Ship it!

rtjohnso (approved):
LGTM

deukyeon:
I can't merge this PR. Maybe due to "some checks haven't completed yet"? How can I resolve this?

rtjohnso:
Let me try to poke the CI to run the checks.

Best,
Rob

PR #583, opened 2023-06-06
Rust wrapper for SplinterDB

etwest:
A rust wrapper for SplinterDB. Implements simple key/value operations by default and allows for more complicated update behavior with user defined rust functions.

The wrapper is an iterative improvement to a previous version created by @rosenhouse.

Changes from the previous version of the wrapper:
- Automatic splinter binding (`splinterdb-sys`) generation when building the rust library.
- More tests of the rust wrapper (`splinterdb-rs`).
- Support for new data_config layout and other changes to SplinterDB.
- Ability to callback to rust functions via the SplinterDB data_config. These callbacks are defined by the `SdbRustDataFuncs` trait.

### DIFF
--- a/rust/splinterdb-rs/src/lib.rs
+++ b/rust/splinterdb-rs/src/lib.rs
@@ -1,0 +1,299 @@
+use std::io::{Error, Result};
+use std::path::Path;
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug)]
+pub struct DBConfig {
+    pub cache_size_bytes: usize,
+    pub disk_size_bytes: usize,
+    pub max_key_size: u8,
+    pub max_value_size: u8,
+}
+
+#[derive(Debug)]
+pub struct SplinterDB {
+    _inner: *mut splinterdb_sys::splinterdb,
+    sdb_cfg: splinterdb_sys::splinterdb_config,
+    data_cfg: splinterdb_sys::data_config,
+}
+
+unsafe impl Sync for SplinterDB {}
+unsafe impl Send for SplinterDB {}
+
+impl Drop for SplinterDB {
+    fn drop(&mut self) {
+        unsafe { splinterdb_sys::splinterdb_close(&mut self._inner) };
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub enum LookupResult {
+    Found(Vec<u8>),
+    FoundTruncated(Vec<u8>),
+    NotFound,
+}
+
+fn as_result(rc: ::std::os::raw::c_int) -> Result<()> {
+    if rc != 0 {
+        Err(Error::from_raw_os_error(rc))
+    } else {
+        Ok(())
+    }
+}
+
+fn create_splinter_slice(ref v: &[u8]) -> splinterdb_sys::slice {
+    unsafe {
+        splinterdb_sys::slice {
+            length: v.len() as u64,
+            data: ::std::mem::transmute(v.as_ptr()),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct IteratorResult<'a> {
+    pub key: &'a [u8],
+    pub value: &'a [u8],
+}
+
+#[derive(Debug)]
+pub struct RangeIterator<'a> {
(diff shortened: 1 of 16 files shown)


etwest:
This pull request does not yet incorporate the rust wrapper tests into the SplinterDB tests. @rtjohnso suggested making the tests check if the user has installed rust, and if so, to perform the rust wrapper tests.

rust/splinterdb-rs/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+        cfg: &DBConfig,
+        open_existing: bool,
+    ) -> Result<()> {
+        let path = path_as_cstring(path); // don't drop until init is done
+
+        // set up the splinterdb config
+        self.sdb_cfg.filename = path.as_ptr();
+        self.sdb_cfg.cache_size = cfg.cache_size_bytes as u64;
+        self.sdb_cfg.disk_size = cfg.disk_size_bytes as u64;
+        self.sdb_cfg.data_cfg = &mut self.data_cfg;
```
XiangpengHao, line 211: This line creates a self referential struct, which is not move safe. We should either make the struct [pinned](https://doc.rust-lang.org/std/pin/struct.Pin.html), or move `self.data_cfg` to heap, e.g., Box<data_cfg>.

The bug can be reproduced by first creating a SpinterDB instance, then assign (move) it to a different variable.
The `self.sdb_cfg.data_cfg` would still point to the old address which is no longer valid.

Sorry for jumping into this pull request. I'm looking for ways to use SplinterDB in Rust and can't wait to try this wrapper :-)

rtjohnso: Hi Xiangpeng,

Thank you for catching this!  Would you be willing to add a commit to this PR fixing the bug?  I think for now pinning it would be most in the spirit of how this struct should be used.  But I'm open to your suggestions, since I'm not a rust expert.

XiangpengHao: Sure I'm happy to fix the bug. It seems I don't have push permission on the pull request. Below is my patch, feel free to append it to the pull request! 

I go with the second option as pinning on the struct requires quite a bit change on how we use the wrapper, while boxing the variable is more straightforward to me (impose no runtime overhead as well). 
 

```diff
From 4857a9898cded6a8704b711e08455ea0daa292d8 Mon Sep 17 00:00:00 2001
From: Xiangpeng <me@haoxp.xyz>
Date: Mon, 15 Jan 2024 15:00:29 -0700
Subject: [PATCH] box the data_cfg to prevent self referential struct

---
 rust/splinterdb-rs/src/lib.rs | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/rust/splinterdb-rs/src/lib.rs b/rust/splinterdb-rs/src/lib.rs
index 3eab876..16fe991 100644
--- a/rust/splinterdb-rs/src/lib.rs
+++ b/rust/splinterdb-rs/src/lib.rs
@@ -18,7 +18,7 @@ pub struct DBConfig {
 pub struct SplinterDB {
     _inner: *mut splinterdb_sys::splinterdb,
     sdb_cfg: splinterdb_sys::splinterdb_config,
-    data_cfg: splinterdb_sys::data_config,
+    data_cfg: Box<splinterdb_sys::data_config>,
 }
 
 unsafe impl Sync for SplinterDB {}
@@ -192,7 +192,7 @@ impl SplinterDB
         SplinterDB {
             _inner: std::ptr::null_mut(),
             sdb_cfg: unsafe { std::mem::zeroed() },
-            data_cfg: new_sdb_data_config::<T>(0),
+            data_cfg: Box::new(new_sdb_data_config::<T>(0)),
         }
     }
 
@@ -208,7 +208,7 @@ impl SplinterDB
         self.sdb_cfg.filename = path.as_ptr();
         self.sdb_cfg.cache_size = cfg.cache_size_bytes as u64;
         self.sdb_cfg.disk_size = cfg.disk_size_bytes as u64;
-        self.sdb_cfg.data_cfg = &mut self.data_cfg;
+        self.sdb_cfg.data_cfg = self.data_cfg.as_mut();
 
         // set key bytes
         self.data_cfg.max_key_size = cfg.max_key_size as u64;
-- 
2.34.1
```

PR #588, opened 2023-07-02
Bidirectional Iterators

rtjohnso:
This PR allows splinter iterators to be bidirectional. To do this

- It renames some of the iterator functions to make them more direction-neutral.  For example, "advance" is renamed "next", so that we can add a "prev".  "at_end" becomes 3 functions "can_prev", "can_next", and "can_curr" which allow for more granular bounds checking and inform us of what state the iterator is in.

- It changes the possible states of an iterator to: 
   - `before start`: Allowed to call "next", not allowed to call "get_current" or "prev"
   - `in range`: Allowed to call all operations
   - `after end`: Allowed to call "prev", not allowed to call "get_current" or "next"
   - `empty`: Not allowed to perform any operations.

- It changes which iterator operations can fail:  Now, only the "next" or "prev" method can fail.  Getting the current key/value pair or checking bounds cannot fail.  The rationale for this decision is: the fewer operations that can fail, the easier it is to use the iterator; in practice, all of the bounds implementations never fail; "next" and "prev" _can_ fail; if there is an implementation that can fail on bounds checks, it is easy to move that failure to "next"/"prev" by performing the bounds check at the end of the "next"/"prev"; same for getting the current key/value pair.

- It adds prev pointers in the btree which must be maintained during operations such as splitting leaves. 

- It adds "start_key" and "start_type" parameters to most iterator init functions. These determine the value of the key we are positioning the iterator relative to and where to actually begin relative to the start key, respectively. "start_type" may be one of the following: less_than, less_than_or_equal, greater_than, or greater_than_or_equal.

---------

Co-authored-by: Evan West <evan.ts.west@gmail.com>

### DIFF
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -340,6 +340,11 @@ splinterdb_iterator_valid(splinterdb_iterator *iter);
 void
 splinterdb_iterator_next(splinterdb_iterator *iter);
 
+// Attempts to move the iterator to the previous item.
+// Any error will cause valid() == false and be visible with status()
+void
+splinterdb_iterator_prev(splinterdb_iterator *iter);
+
 // Sets *key and *value to the locations of the current item
 // Callers must not modify that memory pointed to by the slice
 //
--- a/include/splinterdb/splinterdb.h
+++ b/include/splinterdb/splinterdb.h
@@ -330,20 +330,39 @@ splinterdb_iterator_deinit(splinterdb_iterator *iter);
 
 // Checks that the iterator status is OK (no errors) and that get_current()
 // will succeed. If false, there are two possibilities:
-// 1. Iterator has passed the final item.  In this case, status() == 0
+// 1. Iterator is out of bounds.  In this case, status() == 0
 // 2. Iterator has encountered an error.  In this case, status() != 0
 _Bool
 splinterdb_iterator_valid(splinterdb_iterator *iter);
 
-// Attempts to advance the iterator to the next item.
-// Any error will cause valid() == false and be visible with status()
-void
-splinterdb_iterator_next(splinterdb_iterator *iter);
+/*
+ * splinterdb_iterator_can_next --
+ * splinterdb_iterator_can_prev --
+ *
+ * Knowing the iterator is invalid does not provide enough information to
+ * determine if next and prev are safe operations.
+ *
+ * These functions provide granular information on which operations are safe.
+ * splinterdb_iterator_can_next == TRUE <-> splinterdb_iterator_next is safe
+ * splinterdb_iterator_can_prev == TRUE <-> splinterdb_iterator_prev is safe
+ */
+_Bool
+splinterdb_iterator_can_prev(splinterdb_iterator *iter);
 
-// Attempts to move the iterator to the previous item.
+_Bool
+splinterdb_iterator_can_next(splinterdb_iterator *iter);
+
+// Moves the iterator to the previous item.
+// Precondition for calling: splinterdb_iterator_can_prev() returns TRUE
 // Any error will cause valid() == false and be visible with status()
 void
 splinterdb_iterator_prev(splinterdb_iterator *iter);
+
+// Moves the iterator to the next item.
+// Precondition for calling: splinterdb_iterator_can_next() returns TRUE
+// Any error will cause valid() == false and be visible with status()
+void
+splinterdb_iterator_next(splinterdb_iterator *iter);
 
 // Sets *key and *value to the locations of the current item
 // Callers must not modify that memory pointed to by the slice
--- a/src/btree.c
+++ b/src/btree.c
@@ -1304,7 +1304,6 @@ btree_split_child_leaf(cache                 *cc,
 
    leaf_splitting_plan plan =
       btree_build_leaf_splitting_plan(cfg, child->hdr, spec);
-
    /* p: claim, c: claim, rc: -, on: - */
 
    btree_alloc(cc,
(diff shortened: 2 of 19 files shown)


rtjohnso (commented):
Just checkpointing this.  All files reviewed except for trunk.c and all tests.

include/splinterdb/splinterdb.h
```diff
@@ -340,6 +340,11 @@ splinterdb_iterator_valid(splinterdb_iterator *iter);
 void
 splinterdb_iterator_next(splinterdb_iterator *iter);
 
+// Attempts to move the iterator to the previous item.
```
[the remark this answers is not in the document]
etwest: What exactly are the prereqs?

etwest:
From Alex: Add tests for empty iterators

src/btree.c
```diff
@@ -1298,24 +1300,41 @@ btree_split_child_leaf(cache                 *cc,
 {
    btree_node right_child;
 
-   /* p: claim, c: claim, rc: - */
+   /* p: claim, c: claim, rc: -, on: -  */
```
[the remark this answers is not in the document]
etwest: Added but the summary may not be very good :)

src/trunk.c
```diff
@@ -6012,43 +6017,54 @@ trunk_split_root(trunk_handle *spl, trunk_node *root)
...
+   .prev     = trunk_range_iterator_prev,
 };
 
 platform_status
 trunk_range_iterator_init(trunk_handle         *spl,
                           trunk_range_iterator *range_itor,
                           key                   min_key,
                           key                   max_key,
+                          key                   start_key,
+                          bool                  forwards,
```
rtjohnso, line 6043: Can we use comparison instead of bool forwards?

src/trunk.c
```diff
@@ -6197,77 +6221,79 @@ trunk_range_iterator_init(trunk_handle         *spl,
...
    /*
     * if the merge itor is already exhausted, and there are more keys in the
     * db/range, move to next leaf
     */
-   if (at_end) {
-      KEY_CREATE_LOCAL_COPY(rc,
-                            local_max_key,
-                            spl->heap_id,
-                            key_buffer_key(&range_itor->local_max_key));
+   if (!in_range && trunk_key_compare(spl, local_max, max_key) < 0) {
```
rtjohnso, line 6239: Need to handle empty leaves, which could require going backwards

src/iterator.h
```diff
@@ -8,18 +8,29 @@
...
+                                            comparison from_above);
 typedef void (*iterator_print_fn)(iterator *itor);
 
 typedef struct iterator_ops {
    /* Callers should not modify data pointed to by *key or *data */
-   iterator_get_curr_fn get_curr;
-   iterator_at_end_fn   at_end;
-   iterator_advance_fn  advance;
+   iterator_curr_fn     curr;
+   iterator_in_range_fn in_range;
```
rtjohnso, line 30: Break into `can_next`, `can_prev`, and implement an `iterator_can_curr` method in this file.

etwest:
Splinterdb stress test that ensures iterator next/prev work across multiple trunk leaves.

rtjohnso (commented):
Alas, more work.

src/btree.c
```diff
@@ -2458,28 +2460,31 @@ btree_lookup_and_merge_async(cache             *cc,          // IN
...
 {
-   return itor->idx < itor->curr_min_idx;
+   btree_iterator *itor = (btree_iterator *)base_itor;
+   return itor->curr.addr != itor->end_addr
+          || (itor->idx < itor->end_idx && itor->curr_min_idx != itor->end_idx);
 }
 
-static bool
+// function internal to the btree iterator that checks if the iterator
+// is in a valid state
```
ajhconway, line 2481: Would rephrase "valid state"

ajhconway: Maybe should be called `can_curr`?

ajhconway: Actually kill for the generic

src/btree.c
```diff
@@ -2720,9 +2725,7 @@ btree_iterator_next(iterator *base_itor)
    btree_iterator *itor = (btree_iterator *)base_itor;
 
    // We should not be calling advance on an empty iterator
-   debug_assert(!btree_iterator_after_end(itor));
-   debug_assert(itor->idx >= 0
-                || (btree_iterator_before_start(itor) && itor->idx == -1));
+   debug_assert(btree_iterator_can_next(base_itor));
```
ajhconway, line 2728: Should this be a `platform_assert`?

src/btree.c
```diff
@@ -2832,12 +2833,16 @@ find_btree_node_and_get_idx_bounds(btree_iterator *itor,
...
+   // check if we already need to move to the prev/next leaf
+   if (itor->curr.addr != itor->end_addr
        && itor->idx == btree_num_entries(itor->curr.hdr))
    {
       btree_iterator_next_leaf(itor);
+      itor->curr_min_idx = 0; // we came from an irrelevant leaf
+   }
+   if (itor->curr_min_idx == -1 && itor->idx == -1) {
+      btree_iterator_prev_leaf(itor);
    }
```
ajhconway: Are these adjustments safe if we are currently in the last or first valid leaf?

src/iterator.h
```diff
@@ -47,10 +48,22 @@ iterator_curr(iterator *itor, key *curr_key, message *msg)
...
+static inline bool32
+iterator_can_next(iterator *itor)
+{
+   return itor->ops->can_next(itor);
+}
+
+static inline bool32
+iterator_can_curr(iterator *itor)
+{
+   return itor->ops->can_next(itor) && itor->ops->can_prev(itor);
```
ajhconway, line 66: I guess I would prefer to call the above wrapper functions, but nbd

src/trunk.c
```diff
@@ -6058,15 +6070,20 @@ trunk_range_iterator_init(trunk_handle         *spl,
...
-      return STATUS_OK;
+   if (trunk_key_compare(spl, min_key, start_key) > 0) {
+      // in bounds, start at min
+      start_key = min_key;
    }
+   if (trunk_key_compare(spl, max_key, start_key) <= 0) {
+      // out of bounds, start at max
+      start_key            = max_key;
+      range_itor->can_next = FALSE;
+   }
```
ajhconway: How should we handle `start_key` >= `max_key` and `start_key` < `min_key`?

ajhconway (approved):
Ship it

PR #590, opened 2023-07-07
Trunk Assertion Fix and Simplification

etwest:
This pull request does:

1. Fixes an assertion bug in `trunk_replace_routing_filter`. Previously we asserted that if we compacted down to an empty bundle then the filter must also be NULL. However, this is not necessarily the case. Changed the assertion to assert that if the compacted bundle and the filter are empty then the filter is NULL.
2. Simplifies the `trunk_range_iterator` struct by the removal of `local_max_key`. local_max_key is now a temporary variable in `trunk_range_iterator_init`.


rtjohnso (approved):
LGTM, but let's see if we can get Alex's input.

ajhconway (approved):
Good catch

PR #593, opened 2023-07-24
CI fix: use new version of registry-image resource

rosenhouse:
I've applied this configuration already.  It appears to work.


rtjohnso (approved):
Ship it!

PR #601, opened 2023-11-18
Robj/fallocate

rtjohnso:
Catch fallocate failure on creation.


rosenhouse:
CI should be fixed now.  If you squash/rebase on `main`, things should pass now, I hope?

Issue #599, opened 2023-11-10
Remove #include <unistd.h> from various files that use getpid(). Instead provide platform'ized version of getpid()


gapisback:
This cleanup comment comes out of the review cycle for PR #568 . See [this comment](https://github.com/vmware/splinterdb/pull/568#discussion_r1388416947) from @rtjohnso 

---

I think all the other instances of #include <unistd.h> were introduced by you in the previous PR for multi-process support. We need to clean them all out by platformizing getpid.

---

Solution that will be attempted:

- Add `platform_getpid()` in `platform_linux/platform_inline.h`, defined as OS `getpid()`
- Use `platform_getpid()` everywhere `getpid()` is being used now.
- Will need to `#include <unistd.h>` in above `.h`

This seems like it will work ... try it out.

gapisback:
@rtjohnso -- This is the clean-up issue you raised. Let me know if the proposed solution is acceptable to you. I'll get to this after part-3 lands.

PR #613, opened 2024-01-27
(#599) Replace calls to getpid() with  platform_getpid()

gapisback:
This commit cleans-up the `#include <unistd.h>` from several .c files that needed access to get pid(). Instead, change the code to invoke platform_getpid(), now defined in platform_inline.h


rtjohnso (approved):
Thank you!

PR #616, opened 2024-01-31
Stabilization of part-3 of shared-memory support, based on review comments.

gapisback:
This PR is to test out CI-jobs while applying review comments received for change-set under PR #569.

This PR will not be checked-in, but will be used to refresh that other PR's change-set.


rtjohnso:
Obviated by #655

PR #628, opened 2024-05-23
fix compact_bundle to enqueue filter compaction for entire range

rtjohnso:


src/trunk.c
```diff
@@ -5303,6 +5304,9 @@ trunk_compact_bundle(void *arg, void *scratch_buf)
 
    deinit_saved_pivots_in_scratch(scratch);
 
+   rc = key_buffer_init_from_key(&scratch->req_original_start_key, spl->heap_id, key_buffer_key(&req->start_key));
```
rtjohnso, line 5307: Since we may overwrite the start key of the compaction request, save it here.

src/trunk.c
```diff
@@ -5445,8 +5450,10 @@ trunk_compact_bundle(void *arg, void *scratch_buf)
          key_string(trunk_data_config(spl), end_key),
          req->height,
          req->bundle_no);
+      key_buffer_copy_key(&req->start_key, key_buffer_key(&scratch->req_original_start_key));
```
rtjohnso, line 5453: Restore it here so that the filter compaction will cover the entire range of the original bundle compaction request,

ajhconway (changes requested):
Looks good, but I think it would be better to change the code not to modify `req->start_key` and instead use a live variable for the start key during compaction cleanup.

Issue #306, opened 2022-03-02
driver_test cache_test --async --cache-capacity-gib 20 --db-capacity-gib 100 fails silently. Debugger reports SIGKILL.


gapisback:
On /main, this fails silently: Hard to know where it's failing:

```
Fusion-LocalVM:[63] $ ./bin/driver_test cache_test --async --cache-capacity-gib 20 --db-capacity-gib 100
./bin/driver_test: splinterdb_build_version a2768f5d-dirty
Dispatch test cache_test

Started cache_test!!
fingerprint_size: 27
cache_test: async test started with 1+0 threads (ws=10%)
cache_test: 10485760 pages allocated
test  36% completeKilled
```

Same silent failure in DEBUG build.
Under the debugger, it reports that program terminated with SIGKILL ... Not sure where this is happening. Could not get a backtrace.

```
(gdb) run cache_test --async --cache-capacity-gib 20 --db-capacity-gib 100
Starting program: /home/agurajada/Code/splinterdb/bin/driver_test cache_test --async --cache-capacity-gib 20 --db-capacity-gib 100
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
/home/agurajada/Code/splinterdb/bin/driver_test: splinterdb_build_version a2768f5d-dirty
Dispatch test cache_test

Started cache_test!!
fingerprint_size: 27
cache_test: async test started with 1+0 threads (ws=10%)
cache_test: 10485760 pages allocated
[New Thread 0x7ffade44a700 (LWP 13303)]
test  36% complete[Thread 0x7ffff7b94740 (LWP 13223) exited]

Program terminated with signal SIGKILL, Killed.
The program no longer exists.
```

PR #642, opened 2025-06-24
Fix several bugs in cache_test

rtjohnso:
`test_wait_inflight` did not match how it was used from `test_do_read_batch`.

Also, `cache_assert_ungot` was attempting to implement an impossible check and the test was attempting to use that check.

`clockcache_assert_ungot` was trying to check only that the current thread did not hold a read lock on the specificied page.  But, since CC_RC_WIDTH is only 4, multiple threads may share the same read counter and hence, even if one thread holds no locks, the read counter for a page may be non-zero.  So I changed `clockcache_assert_ungot` to check that no thread has a read lock on the page, which matches its description.

I then modified `cache_test` to use the corrected `cache_assert_ungot`.

### DIFF
--- a/src/clockcache.c
+++ b/src/clockcache.c
@@ -2499,12 +2499,19 @@ clockcache_validate_page(clockcache *cc, page_handle *page, uint64 addr)
 void
 clockcache_assert_ungot(clockcache *cc, uint64 addr)
 {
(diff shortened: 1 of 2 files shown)


rtjohnso:
Fixes #306 .

PR #646, opened 2025-08-18
rc_allocator: fix refcount I/O size (use bytes not elements)

SarahParkSehyun:
**Summary**
io_size for the refcount array was computed from the element count, so only part of the array was persisted/loaded. This left high-index refcounts as zero under long workloads.

**Root cause**
`io_size = ROUNDUP(extent_capacity, page_size);`
used the number of elements. Each refcount is 4 bytes, so the code read/wrote extent_capacity bytes instead of extent_capacity * sizeof(refcount) bytes.

**Fix**
Compute in bytes in both mount and unmount:
```
uint32 io_size =
    ROUNDUP(cfg->extent_capacity * sizeof(refcount), cfg->io_cfg->page_size);
```
Buffer allocation already used sizeof(refcount), so this makes the logic consistent end-to-end.

### DIFF
--- a/src/rc_allocator.c
+++ b/src/rc_allocator.c
@@ -459,8 +459,8 @@ rc_allocator_mount(rc_allocator      *al,
    }
 
    // load the ref counts from disk.
-   uint32 io_size =
-      ROUNDUP(al->cfg->extent_capacity, al->cfg->io_cfg->page_size);
+   uint64 bytes   = (uint64)al->cfg->extent_capacity * sizeof(refcount);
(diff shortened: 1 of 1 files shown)


rtjohnso:
This PR didn't pass tests for some reason, so I reimplemented it in PR #647 , which is now merged into main.

Can you confirm that this bug is now fixed on main?  If so, then we can close this PR.

Thank you!

rtjohnso:
Closing since this is now fixed by PR #647.

